PR #6, opened 2021-08-03
household settings

captaincrazybro:


src/components/Breadcrumb.svelte
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function capitalize(string) {
+    return string.replace(/^\w/, (c) => c.toUpperCase());
+}
+</script>
+
+<!--TODO: change the '>' with the actual icon-->
+{#each urls as val, i}
+    <a href="https://{val.url}" style="margin: 0 3px;">{val.name}</a><!--
```
forevermatt, line 19: You could shorten `https://{val.url}` to `//{val.url}` so that it will work for non-https contexts as well (such as local development).

forevermatt (approved):

src/components/Breadcrumb.svelte
```diff
@@ -0,0 +1,25 @@
+<script>
+let urls = []
+let href = window.location.href.replace("https://", "").replace("http://", "").split("/")
```
hobbitronics, line 3: you might be able to use url() from routify here instead.

src/components/Breadcrumb.svelte
```diff
@@ -0,0 +1,25 @@
...
+
+while (href.length > 1) {
+  let name = href[href.length - 1].toLowerCase()
+  urls.unshift({url: href.join("/"), name: capitalize(name) })
+
+  href.pop()
+}
+
+function capitalize(string) {
```
hobbitronics, line 12: you can use `.capitalize {
	text-transform: capitalize;
}` instead, which will soon be a global class anyways

src/components/Breadcrumb.svelte
```diff
@@ -0,0 +1,25 @@
...
+  return string.replace(/^\w/, (c) => c.toUpperCase());
+}
+</script>
+
+<!--TODO: change the '>' with the actual icon-->
+{#each urls as val, i}
+  <a href="//{val.url}" style="margin: 0 3px;">{val.name}</a><!--
```
hobbitronics, line 19: does this work like a string literal? <a href="//{val.url}"
Also instead of using inline styles we usually prefer using a class

src/components/Breadcrumb.svelte
```diff
@@ -0,0 +1,25 @@
...
+  -->{#if (urls.length - 1) != i}
```
hobbitronics, line 20: indentation looks off here

src/pages/household/settings/index.svelte
```diff
@@ -0,0 +1,69 @@
...
+
+<style>
+#members-page {
+    margin-top: 30px;
+}
+
+#members-list {
+    counter-reset: item;
+    list-style-type: none;
+    padding-left: 0 !important;
```
hobbitronics, line 36: I'm not sure if using !important is a best practice in most cases. @forevermatt ?

src/pages/household/settings/index.svelte
```diff
@@ -0,0 +1,69 @@
...
+    <ul id="members-list">
+      {#each members as m, i}
+        <li class="members-list-item">
+          {m.name}
+          {m.isYou ? "(you)" : ""}
+          <br />
+          <small>{m.isDependent ? "Dependent" : m.email}</small>
+        </li>
+        {#if i == members.length - 1}
+          <li class="members-list-item" style="padding: 0 !important;" />
```
hobbitronics, line 63: same as above re: style and important. If you need important though that's fine for now of course.

src/components/Breadcrumb.svelte
captaincrazybro: If I understand what you are saying correctly, I put the `<!---->` there to make it so there isn't a whitespace and I can make the `#if` statement's contents on a new line from the `a` element.

hobbitronics (approved):
Nice work, just some minor things to possibly change.

hobbitronics:
Don't forget to remove Feature from the PR name

forevermatt (approved):

PR #7, opened 2021-08-03
Simple "add dependent" form

forevermatt:
### Added
- Add a simple "add-dependent" form

### DIFF
--- a/src/components/DependentForm.svelte
+++ b/src/components/DependentForm.svelte
@@ -4,14 +4,14 @@ import { createEventDispatcher } from 'svelte'
 
 const dispatch = createEventDispatcher()
 
+let name = ''
+
 const onCancel = () => {
   dispatch('cancel')
 }
 const onSubmit = () => {
   dispatch('submit', { name })
 }
-
-let name = ''
 </script>
 
 <div class="w-50">
--- a/src/components/DependentForm.svelte
+++ b/src/components/DependentForm.svelte
@@ -1,8 +1,21 @@
 <script>
+import { Button, Form, TextField } from '@silintl/ui-components'
 
+const onSubmit = () => {
+  console.log('Form submitted:', name)
+}
+
+let name = ''
 </script>
 
-<style>
-  
-</style>
-
+<div class="w-50">
+  <Form on:submit={onSubmit}>
+    <TextField label="Dependent Name" bind:value={name} autofocus />
+    <p>
+      Dependents include non-member spouses and children under 26 who haven't
+      married or finished college. Coverage for dependents is limited to $3,000
+      per person.
+    </p>
+    <Button raised>Save</Button>
+  </Form>
+</div>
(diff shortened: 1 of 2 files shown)


src/components/DependentForm.svelte
```diff
@@ -0,0 +1,28 @@
...
+const dispatch = createEventDispatcher()
+
+const onCancel = () => {
+  dispatch('cancel')
+}
+const onSubmit = () => {
+  dispatch('submit', { name })
+}
+
+let name = ''
```
hobbitronics, line 14: Just a note on where we want things declared. Are you going with const first, whether functions or not?

src/components/DependentForm.svelte
```diff
@@ -0,0 +1,28 @@
...
+<div class="w-50">
+  <Form on:submit={onSubmit}>
+    <TextField label="Dependent Name" bind:value={name} autofocus />
+    <p>
+      Dependents include non-member spouses and children under 26 who haven't
+      married or finished college. Coverage for dependents is limited to $3,000
+      per person.
+    </p>
+    <Button on:click={onCancel}>Cancel</Button>
+    <Button raised>Save</Button>
```
hobbitronics, line 26: I added a capitalize class to ui-components, so we will probably want to do that here and add 
`.capitalize {
	text-transform: capitalize;
}` to this file for now.

hobbitronics: until we publish the ui-components repo that is.

src/pages/household/settings/dependent.svelte
```diff
@@ -0,0 +1,17 @@
+<script>
+import { goto } from '@roxi/routify'
+import { Page } from '@silintl/ui-components'
+import DependentForm from '../../../components/DependentForm.svelte'
```
captaincrazybro, line 4: Have a consistent, alphabetic order of organizing the imports. For WeCarry we put the local imports (ones where you have to navigate to the file in your projects) first and then the package imports.

src/pages/household/settings/dependent.svelte
```diff
@@ -0,0 +1,17 @@
...
+
+const onCancel = () => {
+  $goto('../settings')
```
hobbitronics, line 7: is /settings what you want here? maybe this works, I assume you tested.

hobbitronics (approved):

src/pages/household/settings/dependent.svelte
forevermatt: It does work, thanks. It seemed unintuitive, but this is apparently the syntax for going up that level.

src/components/DependentForm.svelte
forevermatt: I was yes. Thoughts?

src/pages/household/settings/dependent.svelte
forevermatt: Okay, great. I'll do that, thanks.

PR #8, opened 2021-08-03
Add Drawer to _layout.svelte

hobbitronics:
- added drawer and fixed layout
- moved css rule to body
- arrange imports and remove extra {}
- made Drawer menu narrower - will work with ui-components update as no class prop now
- capitalize menu items

### DIFF
--- a/src/pages/_layout.svelte
+++ b/src/pages/_layout.svelte
@@ -1,4 +1,6 @@
 <script>
+  import AppFooter from '../components/AppFooter.svelte'
+  import AppHeader from '../components/AppHeader.svelte'
   import { goto } from '@roxi/routify'
   import { Drawer } from '@silintl/ui-components'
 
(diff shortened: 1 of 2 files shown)


src/pages/_layout.svelte
```diff
@@ -1,27 +1,52 @@
 <script>
-import AppFooter from '../components/AppFooter.svelte'
-import AppHeader from '../components/AppHeader.svelte'
-
+  import AppFooter from '../components/AppFooter.svelte'
```
forevermatt, line 2: Completely optional, but it looks like something added a 2-space initial indent for everything in the `<script>` tag, which you could remove.

src/pages/_layout.svelte
```diff
@@ -1,27 +1,52 @@
...
+      icon: 'label',
+      label: 'Claims',
+    },
+    {
+      url: '/faq',
+      icon: 'article',
+      label: 'FAQ',
+    },
+    {
+      url: '/chat',
```
forevermatt, line 25: Do we want to leave this out until the chat feature is within scope? Either is fine, just asking.

forevermatt (approved):

src/pages/_layout.svelte
hobbitronics: I will comment it out for now, if it will be a long time we can delete.

src/pages/_layout.svelte
hobbitronics: yeah, I will comment it out at least for now.

PR #19, opened 2021-08-04
Tables and Filtering

captaincrazybro:
These tables are mainly for the admin view of the policies and items. They do work completely.
The filtering still needs a bit of work, but also does work.


src/components/filteredList/filter.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * 
+     * @description function to filter items, applying this filter
+     * @param {Array} items
+     * @returns {Array} 
+     */
+    filter(items) {
+        if (!this.func) return items
```
forevermatt, line 24: Please wrap if blocks in curly braces 🙂

forevermatt (approved):

PR #21, opened 2021-08-04
Removed TopAppBar Drawer border color and changed background color

hobbitronics:
Removed TopAppBar
changed background color in Drawer content
added theme color - may not be accurate yet!

### DIFF
--- a/src/pages/_layout.svelte
+++ b/src/pages/_layout.svelte
@@ -4,6 +4,9 @@ import AppHeader from '../components/AppHeader.svelte'
 import { goto } from '@roxi/routify'
 import { Drawer } from '@silintl/ui-components'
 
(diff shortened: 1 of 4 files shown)


src/pages/_layout.svelte
```diff
@@ -4,6 +4,9 @@ import AppHeader from '../components/AppHeader.svelte'
 import { goto } from '@roxi/routify'
 import { Drawer } from '@silintl/ui-components'
 
+let isFullHeightMenu = false
+let hasTopAppBar = false
```
forevermatt, line 8: Do you have to define these, or if you simply leave them off the Drawer do they default to false?

forevermatt (approved):
Approved, with question

PR #24, opened 2021-08-04
Update our "Dependent" form

forevermatt:
### Added
- Add new fields to "dependent" form
- Enable editing a dependent

### Fixed
- Stop using `<p>` tags inside RadioOptions component

### DIFF
--- a/src/pages/household/settings/dependent.svelte
+++ b/src/pages/household/settings/dependent.svelte
@@ -6,10 +6,6 @@ import { Page } from '@silintl/ui-components'
 const onCancel = () => {
   $goto('../settings')
 }
-const onRemove = event => {
-  const dependentUuid = event.detail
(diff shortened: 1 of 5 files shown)


src/pages/household/settings/dependent.svelte
```diff
@@ -6,12 +6,16 @@ import { Page } from '@silintl/ui-components'
 const onCancel = () => {
   $goto('../settings')
 }
+const onRemove = event => {
```
captaincrazybro, line 9: If I'm not mistaken, isn't this the form to create a dependent? In that case, I don't think you need to address the on:remove event, since it will never be fired off from this page.

captaincrazybro (approved):

hobbitronics (approved):

PR #28, opened 2021-08-04
Add menu

hobbitronics:
- Created menu
- Added to Header
- updated index.js

### DIFF
--- a/src/components/AppHeader.svelte
+++ b/src/components/AppHeader.svelte
@@ -55,13 +55,13 @@ header {
   </div>
 
   <div id="toolbar" class="flex justify-end toolbar mdc-menu-surface--anchor">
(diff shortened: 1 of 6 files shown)


src/components/AppHeader.svelte
```diff
@@ -1,16 +1,27 @@
...
+import { Badge, IconButton, isAboveTablet } from '@silintl/ui-components'
+import { createEventDispatcher } from 'svelte'
+import { Menu } from './index';
 
 let showImage = true
 let alt = 'avatar'
 let showDrawerButton
+let menuToggler = false
+
+const menuItems = [
```
captaincrazybro, line 13: I think Matt said for the constants to be before `let` declarations.

captaincrazybro: Besides `const` functions

PR #36, opened 2021-08-05
Added policies endpoints wrapper

captaincrazybro:


src/data/policies.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ *
+ * @description updates a current policy with the given 'policyData'
+ * @export
+ * @param {Number} id
+ * @param {Object} policyData
+ * @return {Object} 
+ */
+export async function updatePolicy(id, policyData) {
+    let parsedPolicyData = {
```
hobbitronics, line 21: could this be a const?

src/data/policies.js
```diff
@@ -0,0 +1,94 @@
...
+        household_id: policyData.household_id,
+        cost_center: policyData.cost_center,
+        account: policyData.account,
+        entity_code: policyData.entity_code
+    }
+
+    let updatedPolicy = await PUT(`/policies/${id}`, parsedPolicyData)
```
hobbitronics, line 28: same as above

hobbitronics (approved):
Looks good!

src/data/policies.js
```diff
@@ -0,0 +1,94 @@
...
+        household_id: policyData.household_id,
+        cost_center: policyData.cost_center,
+        account: policyData.account,
+        entity_code: policyData.entity_code
+    }
+
+    const updatedPolicy = await PUT(`/policies/${id}`, parsedPolicyData)
+
+    policies.update(currPolicies => {
+        let i = currPolicies.findIndex(pol => pol.id == id)
```
forevermatt, line 31:
```suggestion
        let i = currPolicies.findIndex(pol => pol.id === id)
```

src/data/policies.js
```diff
@@ -0,0 +1,94 @@
...
+        name: itemData.name,
+        category_id: itemData.category_id,
+        country: itemData.country,
+        description: itemData.description,
+        make: itemData.make,
+        model: itemData.model,
+        serial_number: itemData.serial_number,
+    }
+
+    const item = await POST(`/policies/${id}/items`, parsedItemData)
```
forevermatt, line 74: After adding an item, you may want to reload the policies:
```suggestion
    const item = await POST(`/policies/${id}/items`, parsedItemData)
    loadPolicies()
```

captaincrazybro: The GET `/policies` endpoint does not output any items property in each policy so I don't think that would really help any.

forevermatt: Oh, sorry. You're right. I mixed up endpoints.

forevermatt (approved):

PR #38, opened 2021-08-05
Added item details page

captaincrazybro:


src/pages/items/[itemId]/index.svelte
```diff
@@ -0,0 +1,78 @@
...
+<!--TODO: get this from backend when available-->
+<p>Annual premium: ${16.20}</p>
+<p>Description: {itemDetails.description}</p>
+<!--TODO: get this from backend when available-->
+<p>Accountable person: {"Jeff Smith"}</p>
+<p>Unique identifier: {itemDetails.serial_number}</p>
+<p>Coverage added: {new Date(itemDetails.coverage_start_date).toDateString()}</p>
+<p>Coverage ends: {"13 December 2029"}</p>
+<div>
+  <Button on:click={handlePrimaryClicked} raised>Edit Details</Button>
```
Comment (forevermatt) [inline src/pages/items/[itemId]/index.svelte:71]: If you don't mind, I think this will be more readable if you name the functions for what they do, so something like "goToEditItem()" instead of "handlePrimaryClicked". Similarly for handleDeleteClicked.

forevermatt (approved):

src/components/Banner.svelte
```diff
@@ -0,0 +1,21 @@
+<script>
+
+export let color = 'white'
+export let border = 'none'
+export let background = 'var(--mdc-theme-primary)'
+
+</script>
+
+<style>
+.badge {
```
hobbitronics, line 10: we have a badge component, you don't need to use it here, just want to make you aware if you are not.

hobbitronics (approved):

PR #39, opened 2021-08-05
Fake authentication

forevermatt:
### Added
- Add authentication, but fake it for the moment

### DIFF
--- a/src/i18n/index.js
+++ b/src/i18n/index.js
@@ -3,6 +3,7 @@ import phrases from './phrases.json'
 export default function t(key) {
   const phrase = phrases[key]
   if (phrase === undefined) {
+    console.log(key)
     return '⤂ translation key not found! ⤃'
   }
 
--- a/src/pages/index.svelte
+++ b/src/pages/index.svelte
@@ -1,15 +1,5 @@
 <script>
-import { loadUser } from '../data/user'
 import { goto } from '@roxi/routify'
-import { onMount } from 'svelte'
 
-onMount(async () => {
-  loadUser().then(() => {
-    $goto('/home')
-  }).catch(error => {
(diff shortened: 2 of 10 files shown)


src/i18n/index.js
```diff
@@ -3,6 +3,7 @@ import phrases from './phrases.json'
 export default function t(key) {
   const phrase = phrases[key]
   if (phrase === undefined) {
+    console.log(key)
```
captaincrazybro, line 6: Is this a debug `console.log()`?

src/pages/index.svelte
```diff
@@ -1,22 +1,5 @@
...
-import { logout } from '../authn'
-import { GET } from '../data'
-import { onMount } from 'svelte'
-import user from '../authn/user'
+import { goto } from '@roxi/routify'
 
-onMount(async () => {
-  await GET('bearer')  // this helps simulate a page accessed that may require the user be authenticated
-})
+$goto('/home')
```
captaincrazybro, line 4: I thought `index.svelte` was for logging in?

captaincrazybro (approved):

hobbitronics (approved):

src/i18n/index.js
forevermatt: Only sort of. If a key is provided that isn't in our phrases file, we don't have any clues as to what it was. This lets us at least see in the console what it was, so we can know what to add.

src/pages/index.svelte
forevermatt: It was going to be... but I realized that when I refresh the page on any other route, we don't do the login for unauthenticated users. So my file move was mostly unnecessary after all. (Sorry again.)

PR #40, opened 2021-08-05
Added items data wrapper

captaincrazybro:


src/data/items.js
```diff
@@ -0,0 +1,41 @@
...
+ * @export
+ * @param {Object} item
+ * @return {Object} 
+ */
+export async function updateItem(item) {
+    if (!item.id) throwError("item id not set")
+    
+    loading.set(true)
+
+    delete item.id
```
hobbitronics, line 19: how can the UPDATE function work if there is no id?

hobbitronics (approved):
Approved with a question

src/data/items.js
captaincrazybro: Oh, woops

PR #54, opened 2021-08-06
Finished (for now) the remaining endpoint wrapper functions in `data`

captaincrazybro:


src/data/dependents.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // TODO: uncomment when endpoint is finished
+    // await DELETE(`/dependents/${depId}`)
+
+    dependents.update(currDeps => {
+      return currDeps.filter(dep => dep.id !== depId)
+    })
+
+    loading.set(false)
+    return null
```
forevermatt, line 63: You don't need to return anything from functions where you have no use for a return value, so you could remove this `return null`

forevermatt (approved):

hobbitronics (approved):

PR #60, opened 2021-08-10
Menu bugfixes

hobbitronics:
## Fixed:

- bug requiring a second click to open menu
- bug causing an extra focusable mdc-list-item

### Added support for space and Enter keys to select menu items

Do we need the menu items to remain an anchor element? Seems to work fine with `<li>` but thought you and Levi mentioned something about needing it for the breadcrumbs.

### DIFF
--- a/src/components/AppHeader.svelte
+++ b/src/components/AppHeader.svelte
@@ -61,7 +61,6 @@ header {
       {/if}
     </button>
 
-    <!-- TODO set menuOpen to false when menu closes -->
     <Menu autofocus bind:menuOpen {menuItems} on:syncToggler={() => menuOpen = false}/>
   </div>
 </header>
--- a/src/components/AppHeader.svelte
+++ b/src/components/AppHeader.svelte
@@ -18,7 +18,7 @@ const menuItems = [
 let showImage = true
 let alt = 'avatar'
 let showDrawerButton
-let menuToggler = false
+let menuOpen = false
 
 const dispatch = createEventDispatcher()
 
@@ -28,7 +28,7 @@ $: ownerInitial = $user.first_name?.charAt(0) || ''
 onMount(() => showOrHideDrawerButton())
 
 const avatarError = () => showImage = false
-const toggleMenu = () => menuToggler = !menuToggler
+const toggleMenu = () => menuOpen = !menuOpen
 const showOrHideDrawerButton = () => isAboveTablet() ? (showDrawerButton = false) : (showDrawerButton = true)
 const toggleDrawerHandler = () => dispatch('toggleDrawer') //TODO toggle drawer
 </script>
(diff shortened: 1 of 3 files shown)


forevermatt:
> Do we need the menu items to remain an anchor element? Seems to work fine with `<li>` but thought you and Levi mentioned something about needing it for the breadcrumbs.

I seem to remember it keyboard navigation being affected by whether something was a button vs. link vs. other DOM element, but feel free to do whatever seems to work as desired.

src/components/mdc/Menu/Menu.svelte
```diff
@@ -33,26 +34,25 @@ const handleItemClick = url => {
     $goto(url)
   }
 }
+const handleItemKeydown = (e, url) => (e.code == 'Space' || e.code == 'Enter') && handleItemClick(url)
+const onMenuClose = () => !menu.open && (menuToggler = false)
```
forevermatt, line 38: I'm a little confused about the purpose of the `menuToggler` variable. It's name sound like it's an active thing, but it seems to be a boolean for... indicating whether the menu should be open?

Okay, I think it's making more sense now. What if we changed this line as follows?
```suggestion
const closeMenu = () => {
  if (menu.open) {
    menuToggler = false
  }
}
```

We might not even need the `if` condition, but I don't know what the Menu component does if `open` is false and we set it to false again.

forevermatt (approved):
Approved, with one question/suggestion

src/components/mdc/Menu/Menu.svelte
hobbitronics: menuToggler is a boolean used to set menu.open which makes the menu open or close. I will change menuToggler to menuOpen. If we could bind directly to menu.open we wouldn't need it. The problem was that clicking off the menu set menu.open to false but I couldn't communicate that back to menuToggler without circular statements.  After trying without the if I realized we need it for clicking on the menu itself.
`const closeMenu = () => {
  if (!menu.open) {
    menuToggler = false
  }
}`

hobbitronics: Just to clarify, the menu closes itself we the user clicks, focuses elsewhere. We are just trying to let the parent know to synchronize with the prop so that menuOpen is reset to false when we try to open it the second time.

hobbitronics:
> > Do we need the menu items to remain an anchor element? Seems to work fine with `<li>` but thought you and Levi mentioned something about needing it for the breadcrumbs.
> 
> I seem to remember it keyboard navigation being affected by whether something was a button vs. link vs. other DOM element, but feel free to do whatever seems to work as desired.

Keyboard nav seems to work. We'll use li and keep an eye open for problems, but I don't like using the anchor without using href since we use $goto.

src/components/AppHeader.svelte
```diff
@@ -61,8 +61,8 @@ header {
       {/if}
     </button>
 
-    <!-- TODO set menuToggler to false when menu closes -->
-    <Menu autofocus bind:menuToggler {menuItems} on:syncToggler={() => menuToggler = false}/>
+    <!-- TODO set menuOpen to false when menu closes -->
```
hobbitronics, line 64:
```suggestion
```

PR #62, opened 2021-08-11
Centralize loading

hobbitronics:
used loading store from /components/progress/index.js and its functions start and stop.

here's an example of how it works https://codepen.io/midotron/pen/gOWqQWK?editors=1111

### DIFF
--- a/src/pages/home.svelte
+++ b/src/pages/home.svelte
@@ -5,6 +5,7 @@ import { claims } from '../data/claims.js'
 import { Datatable, Menu, ClaimCards, Row } from '../components/'
 import { Checkbox, Page } from '@silintl/ui-components'
 import { goto } from '@roxi/routify'
+import { loading } from '../components/progress/index'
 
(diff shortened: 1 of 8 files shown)


src/pages/home.svelte
```diff
@@ -152,8 +152,7 @@ const handleMoreVertClick = id => {
   </Row>
 
   <Row cols={'12'}>
-    <!--TODO: add an '$' before the 'loading' when it because a store-->
-    {#if loading }
+    {#if $loading }
```
forevermatt, line 155: This looks like it would hide all of the items any time anything was being loaded, even if the list of items has already been loaded. Is that an acceptable UX? If not, do we want to make our "this thing is currently loading" logic more fine grained?

hobbitronics: I added another export function to check pending for the id used in start(). That way it will only show "loading..." for that api call.

PR #65, opened 2021-08-13
claimsView endpoint call

hobbitronics:
call the api for getClaim() instead of use store

### DIFF
--- a/src/data/claims.js
+++ b/src/data/claims.js
@@ -1,6 +1,6 @@
 import { writable } from "svelte/store"
 import { start, stop } from "../components/progress"
-import { GET } from "."
+import { CREATE, GET } from "."
 
 export const claims = writable([])
 export const initialized = writable(true)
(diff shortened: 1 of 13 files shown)


src/data/claims.js
```diff
@@ -123,31 +74,29 @@ export async function createClaim(item, claimData) {
...
 }
 
-export function getClaim(claims, itemId) {
-  return claims.find(clm => clm.itemId === itemId)
+export async function getClaim(itemId) {
+  start(itemId)
+
+  let claim = await GET(`claims/${id}`)
+
+  return claim
```
Schparky, line 97: Looks like this bypasses `stop(itemId)`

src/data/claims.js
```diff
@@ -123,31 +74,29 @@ export async function createClaim(item, claimData) {
...
 
   stop(item.id)
 }
 
-export function getClaim(claims, itemId) {
-  return claims.find(clm => clm.itemId === itemId)
+export async function getClaim(itemId) {
+  start(itemId)
+
+  let claim = await GET(`claims/${id}`)
```
Schparky, line 95:
```suggestion
  const claim = await GET(`claims/${id}`)
```

Schparky (changes requested):

src/data/claims.js
hobbitronics: I had one little job to do and codium even shaded stop() out. Nice catch.

Schparky (approved):

PR #66, opened 2021-08-13
Editor config

forevermatt:
### Added
- Add an `.editorconfig` file to reduce cross-IDE style differences

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# https://EditorConfig.org
+
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


Schparky (approved):

hobbitronics (approved):
One thing I should have asked earlier. Do we wan't semi colons to be in the config?

forevermatt:
> One thing I should have asked earlier. Do we wan't semi colons to be in the config?

EditorConfig doesn't seem to support that. We could certainly pick out a more in-depth style config as well, if desired.

PR #70, opened 2021-08-16
Avoid API calls with a policy id of `null`

forevermatt:
### Fixed
- Only try to load the items if we have a policy ID
- Don't load the list of dependents until we have a policy ID
- Fix parameter documentation for `getItems()`

### DIFF
--- a/src/pages/home.svelte
+++ b/src/pages/home.svelte
@@ -26,12 +26,14 @@ let goToItemDetails = true
 let shownMenus = {}
 let items = []
 
+$: if ($user.policy_id) {
+  getItems($user.policy_id).then(loadedItems => items = loadedItems)
+}
+
 onMount( () => {
   loadClaims()
 
   loadPolicies()
-  
-  getItems($user.policy_id).then(loadedItems => items = loadedItems)
 })
 
 const redirect = url => {
--- a/src/data/items.js
+++ b/src/data/items.js
@@ -7,7 +7,7 @@ import { start, stop } from "../components/progress/index.js"
  *
  * @description a function to fetch the items of a policy
  * @export
- * @param {Number} id
+ * @param {string} policyId -- The UUID for the desired policy
  * @return {Object} 
  */
 export async function getItems(policyId) {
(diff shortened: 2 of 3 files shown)


src/pages/home.svelte
```diff
@@ -26,12 +26,14 @@ let goToItemDetails = true
 let shownMenus = {}
 let items = []
 
+$: if ($user.policy_id) {
+  getItems($user.policy_id).then(loadedItems => items = loadedItems)
```
Schparky, line 30: I've become accustomed to this form:
```javascript
 items = await getItems($user.policy_id)
```
Is the "then" style becoming the norm?

Schparky (approved):

src/pages/home.svelte
hobbitronics: then is the older syntax replaced by async await. They do then same thing but await looks more like synchronous code

hobbitronics (approved):
If I understand correctly this is in case the user changes from their home policy to a work policy?

forevermatt:
> If I understand correctly this is in case the user changes from their home policy to a work policy?

Actually, this is so that we don't call `/policies/null/items` if we don't yet have the policy ID (or if the user lacks a policy ID). As soon as `$user.policy_id` is truthy, the reactive statement will run and will get the items.

src/pages/home.svelte
forevermatt: @Schparky - I prefer the `await` style, too, but it can only be used in an `async` function. This is the root-level `<script>` tag in this component, so `await` isn't available.

I could change it to call an async function, but it would either look weird (because we're calling a function defined later, which normally doesn't work with `const`-defined functions) or I'd need to put the function definition above these reactive statements, contrary to our general organization convention within Svelte components:
```js
$: $user.policy_id && loadItemsFor($user.policy_id)

const loadItemsFor = async policyId => {
  items = await getItems(policyId)
}
```

hobbitronics:
> > If I understand correctly this is in case the user changes from their home policy to a work policy?
> 
> Actually, this is so that we don't call `/policies/null/items` if we don't yet have the policy ID (or if the user lacks a policy ID). As soon as `$user.policy_id` is truthy, the reactive statement will run and will get the items.

If only I could remember your bullet points until I get to the end of the PR.

PR #72, opened 2021-08-16
Edit dependents

forevermatt:
### Fixed
- Enabled editing a household policy's dependents
- Preselect the matching option in RadioOptions component

### Changed
- On the household settings page, only load the dependents if they haven't been loaded yet

---

I'm avoiding reloading the dependents on subsequent visits to "/household/settings", primarily to avoid overwriting our changes when we delete one (since the delete-dependent API endpoint isn't ready yet). It seems reasonable to keep that change even after the delete endpoint is ready, though that could certainly be changed if we choose to.


src/pages/household/settings/dependent/[uuid].svelte
```diff
@@ -1,24 +1,32 @@
...
+  deleteDependent,
+  dependents,
+  initialized,
+  loadDependents,
+  updateDependent
+} from '../../../../data/dependents'
 import DependentForm from '../../../../components/DependentForm.svelte'
 import { goto } from '@roxi/routify'
 import { Page } from '@silintl/ui-components'
+import user from "../../../../authn/user";
```
Comment (Schparky) [inline src/pages/household/settings/dependent/[uuid].svelte:12]:
```suggestion
import user from "../../../../authn/user"
```
Remove the semicolon? Also, are imports sorted in this project?

Schparky (approved):

src/pages/household/settings/dependent/[uuid].svelte
Comment (forevermatt) [reply src/pages/household/settings/dependent/[uuid].svelte]: Yes, and yes. I'll update it, thanks.

Comment (forevermatt) [reply src/pages/household/settings/dependent/[uuid].svelte]: I probably pasted in code that used `$user`, so PhpStorm automatically added the import

PR #75, opened 2021-08-17
Household policy members

forevermatt:
### Added
- Load household members from API
- Cache policy members, but refresh on each visit to household settings

---

I'm curious what you think about this "caching" strategy, @Schparky and @hobbitronics. This avoids an empty list of household members after we've loaded them the first time, but ensures the user sees current data quickly even on subsequent visits to that page. (Note that this is only caching them in memory, not across full page reloads.)

### DIFF
--- a/src/data/policy-members.js
+++ b/src/data/policy-members.js
@@ -12,13 +12,14 @@ export const membersByPolicyId = writable({})
  * @return {Object[]} 
  */
 export async function loadMembersOfPolicy(policyId) {
-  start(policyId)
+  const urlPath = `policies/${policyId}/members`
+  start(urlPath)
 
-  const policyMembers = await GET(`policies/${policyId}/members`)
+  const policyMembers = await GET(urlPath)
   membersByPolicyId.update(data => {
     data[policyId] = policyMembers
     return data
   })
 
-  stop(policyId)
+  stop(urlPath)
 }
--- a/src/data/policy-members.js
+++ b/src/data/policy-members.js
@@ -1,5 +1,8 @@
 import { GET } from './index.js'
 import { start, stop } from '../components/progress'
+import { writable } from 'svelte/store';
+
+export const membersByPolicyId = writable({})
 
 /**
  * A function to fetch the items of a policy
@@ -8,11 +11,15 @@ import { start, stop } from '../components/progress'
  * @param {string} policyId -- The UUID for the desired policy
  * @return {Object[]} 
  */
-export async function getPolicyMembers(policyId) {
+export async function loadMembersOfPolicy(policyId) {
+  console.log('loadMembersOfPolicy(', policyId, ')')
   start(policyId)
 
   const policyMembers = await GET(`policies/${policyId}/members`)
+  membersByPolicyId.update(data => {
+    data[policyId] = policyMembers
+    return data
+  })
 
   stop(policyId)
-  return policyMembers
 }
--- a/src/data/policy-members.js
+++ b/src/data/policy-members.js
@@ -1,5 +1,4 @@
 import { GET } from './index.js'
-import { throwError } from '../error'
 import { start, stop } from '../components/progress'
 
 /**
--- a/src/data/policy-members.js
+++ b/src/data/policy-members.js
@@ -1,6 +1,6 @@
 import { GET } from './index.js'
 import { start, stop } from '../components/progress'
-import { writable } from 'svelte/store';
+import { writable } from 'svelte/store'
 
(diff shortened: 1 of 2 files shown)


src/data/policy-members.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * A function to fetch the items of a policy
+ * 
+ * @export
+ * @param {string} policyId -- The UUID for the desired policy
+ * @return {Object[]} 
+ */
+export async function loadMembersOfPolicy(policyId) {
+  start(policyId)
```
forevermatt, line 15: We've overloaded calls to start with this value, as we provide the same argument when loading items, adding a dependent, loading dependents, and now loading members.

I suppose we could use the full path of the API endpoint being called as the start/stop key... thoughts?

hobbitronics: that's a good idea, the start and stop functions don't care what we use.

src/data/policy-members.js
```diff
@@ -0,0 +1,24 @@
+import { GET } from './index.js'
+import { start, stop } from '../components/progress'
+import { writable } from 'svelte/store';
```
hobbitronics, line 3: the autoimport added a ;

src/data/policy-members.js
```diff
@@ -0,0 +1,24 @@
...
+ * @export
+ * @param {string} policyId -- The UUID for the desired policy
+ * @return {Object[]} 
+ */
+export async function loadMembersOfPolicy(policyId) {
+  start(policyId)
+
+  const policyMembers = await GET(`policies/${policyId}/members`)
+  membersByPolicyId.update(data => {
+    data[policyId] = policyMembers
```
hobbitronics, line 19: I don't see data declared. Do you mean to add a const here?

forevermatt: `data` is the callback parameter for the data already in the `membersByPolicyId` store.

src/pages/household/settings.svelte
```diff
@@ -1,31 +1,28 @@
...
-$: if ($user.policy_id && !$initialized) {
+$: if ($user.policy_id && !$haveLoadedDependents) {
   loadDependents($user.policy_id)
 }
 
+$: if ($user.policy_id) {
+  loadMembersOfPolicy($user.policy_id)
+}
+
+$: haveLoadedPolicyMembers = $membersByPolicyId[$user.policy_id] !== undefined
```
hobbitronics, line 19: in some places we are using initialized in the stores, others locally by checking length, etc. Is there a preferable way doing this?

hobbitronics (approved):
Approved with question

src/data/policy-members.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {string} policyId -- The UUID for the desired policy
+ * @return {Object[]} 
+ */
+export async function loadMembersOfPolicy(policyId) {
+  const urlPath = `policies/${policyId}/members`
+  start(urlPath)
+
+  const policyMembers = await GET(urlPath)
+  membersByPolicyId.update(data => {
+    data[policyId] = policyMembers
```
Schparky, line 20: Clever use of the URL Path as the map key. I like it.

src/pages/household/settings.svelte
```diff
@@ -1,31 +1,28 @@
...
+$: if ($user.policy_id && haveLoadedPolicyMembers) {
+  householdMembers = $membersByPolicyId[$user.policy_id]
+}
+
 const edit = id => $goto(`/household/settings/dependent/${id}`)
+const isYou = householdMember => householdMember.id === $user.id
```
Schparky, line 25: "you" seems like an unusual pronoun. Why not "me"?

Schparky (approved):

src/pages/household/settings.svelte
forevermatt: Hmm... interesting question. I think this one ends up simpler this way, because of wanting it to be re-evaluated any time `$membersByPolicyId` or `$user.policy_id` changes.

For stores where the entire store has either been loaded (initialized) or not, I think using that `initialized` store is still a good idea. For stores that hold multiple sets of data (such as a collection of data about various policies, like this one), I think checking it like this will probably be simpler. Worth considering (and possibly reevaluation later), though.

src/pages/household/settings.svelte
forevermatt: I hadn't even thought of that. I think it's because the mockups used "you" instead of "me", but you could certainly ping the design team about that. Either seems fine to me.

PR #76, opened 2021-08-17
PolicyCreateItem api call

hobbitronics:
- Parsed formdata to make the PolicyItemCreate api call. Example from console below
- brought in Dependants
- added dates/index.js for formatting dates
- refactored and added some mock data back in for non-hydrated api fields

```
category_id: "63bcf980-e1f0-42d3-b2b0-2e4704159f4f"
country: "USA"
coverage_amount: 123
coverage_start_date: "2021-08-17"
coverage_status: "Draft"
description: "iphone se"
in_storage: false
make: "se"
model: "iphone"
name: "se"
purchase_date: "2021-08-17"
serial_number: "123"
```

Note: still getting an error from the server:` debug_msg: "failed to close prepared statement: ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02): ERROR: insert or update on table \"items\" violates foreign key constraint \"items_category_id_fkey\" (SQLSTATE 23503)"`

### DIFF
--- a/src/data/itemCategories.js
+++ b/src/data/itemCategories.js
@@ -1,5 +1,6 @@
 import { GET } from "."
-import { writable } from "svelte/store";
+import { writable } from "svelte/store"
+import { start, stop } from "../components/progress"
 
 export const categories = writable([])
 export const initialized = writable(false)
@@ -14,8 +15,12 @@ export async function init() {
  * @export
  */
 export async function loadCategories() {
+  start()
+
   let catz = await GET('item-categories')
 
+  stop()
+  
   categories.set(catz)
   initialized.set(true)
 }
--- a/src/data/itemCategories.js
+++ b/src/data/itemCategories.js
@@ -15,11 +15,11 @@ export async function init() {
  * @export
  */
 export async function loadCategories() {
-  start()
+  start('itemCategories')
 
   let catz = await GET('item-categories')
 
-  stop()
+  stop('itemCategories')
   
   categories.set(catz)
   initialized.set(true)
--- a/src/data/itemCategories.js
+++ b/src/data/itemCategories.js
@@ -1,6 +1,6 @@
 import { GET } from "."
-import { writable } from "svelte/store"
 import { start, stop } from "../components/progress"
+import { writable } from "svelte/store"
 
 export const categories = writable([])
 export const initialized = writable(false)
--- a/src/pages/items/new.svelte
+++ b/src/pages/items/new.svelte
@@ -2,7 +2,7 @@
 import user from '../../authn/user.js'
 import { addItem } from '../../data/items.js'
 import { dependents, loadDependents, initialized as depsInitialized } from '../../data/dependents.js'
-import { categories as categoryOptions, init, initialized as CatItemsInitialized } from '../../data/itemCategories'
+import { categories as categoryOptions, init, initialized as catItemsInitialized } from '../../data/itemCategories'
 import { formatDate } from '../../dates.js'
 import { Breadcrumb, Description, MoneyInput } from '../../components'
 import { goto } from '@roxi/routify'
@@ -42,7 +42,7 @@ onMount(async () => {
   if (! $depsInitialized && $user.policy_id) {
     loadDependents($user.policy_id)
   }
-  if (! $CatItemsInitialized) {
+  if (! $catItemsInitialized) {
     await init()
   }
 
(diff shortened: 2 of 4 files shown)


src/pages/items/new.svelte
```diff
@@ -1,55 +1,67 @@
 <script>
 import user from '../../authn/user.js'
 import { addItem } from '../../data/items.js'
+import { dependents, loadDependents, initialized as depsInitialized } from '../../data/dependents.js'
+import { categories as categoryOptions, init, initialized as CatItemsInitialized } from '../../data/itemCategories'
```
forevermatt, line 5: You might want to lowercase the "C" at the beginning of `CatItemsInitialized`, to keep our naming convention consistent (to match `depsInitialized` on the line above).

src/pages/items/new.svelte
```diff
@@ -1,55 +1,67 @@
...
+  uniqueIdentifier: '',
+}
+
+const dateFormat = {
+  year: 'numeric',
+  month: 'numeric',
+  day: 'numeric'
 }
+  
+const [month, day, year] = formatDate(Date(), 'en-US', dateFormat).split('/')  //en-US to give consistent response
```
forevermatt, line 33: Another option, if you're interested, is to call `.toISOString()` on the given Date object and take the first 10 characters of it. That will give you a YYYY-MM-DD formatted date string. Just an option.

src/pages/items/new.svelte
```diff
@@ -1,55 +1,67 @@
...
-  {
-    name: 'Sarah Smith',
-    id: '22222222-2222-4222-2222-222222222222',
-  },
-]
+$: formData.coverage_start_date = `${year}-${formatMonthOrDay(month)}-${formatMonthOrDay(day)}` //api requires yyyy-mm-dd
+$: formData.purchase_date = formData.coverage_start_date
+$: accountablePersons = $dependents //TODO add current User to this: = [$dependents..., currentUser]
+
+onMount(async () => {
```
forevermatt, line 41: Is there a benefit to calling these in the `onMount()` function?

src/pages/items/new.svelte
```diff
@@ -1,55 +1,67 @@
...
-const onSubmit = event => {
-  // TEMP
-  formData.category = categoryOptions.find(cat => cat.id === formData.category)
+
+const onSelectCategory = event => {
+  formData.category = event.detail?.id
+}
+
+const onSubmit = () => {
   addItem($user.policy_id, formData)
```
forevermatt, line 63: It may be worth adding an `await` here, so that if addItem() gets an error response back from the API (such as if some piece of data was missing), the user will still be here on the form instead of having already been sent to "/home" and potentially lost the information they'd entered into the form. I don't know for certain whether that would happen, though.

src/pages/items/new.svelte
hobbitronics: i wanted last statement to run after await init() but I suppose I can write that like: `$: $catItemsInitialized && categories = ....`

forevermatt: Yeah, it seems like reactive statements let us write more declarative code (rather than imperative), which can really maximize some of the strengths of Svelte.

It avoids the need to care about specific timing issues (e.g. onMount), because we can initialize the variable to a valid "empty" value, and whenever the data is loaded, it will automatically fill it with the real value.

forevermatt (approved):

src/data/itemCategories.js
```diff
@@ -1,5 +1,6 @@
 import { GET } from "."
-import { writable } from "svelte/store";
+import { writable } from "svelte/store"
+import { start, stop } from "../components/progress"
```
Schparky, line 3: sort?

src/data/itemCategories.js
```diff
@@ -14,8 +15,12 @@ export async function init() {
  * @export
  */
 export async function loadCategories() {
+  start()
```
Schparky, line 18: Does this need a parameter now?

Schparky (approved):

src/pages/items/new.svelte
hobbitronics: that's a lot less code. I even used that method a couple days ago, just forgot about it already.

src/data/itemCategories.js
hobbitronics: this just starts generic loading "mode". We could give a static id instead of allowing too many generic loadings.

PR #77, opened 2021-08-17
Group dependents data by policy ID

forevermatt:
### Changed
- Refactor dependents store to be grouped by policy ID
- Use API URL path as unique start/stop key for "dependent" data functions

### Fixed
- Remove unused import
- Update household settings page to use `$dependentsByPolicyId`
- Update edit-dependent page to use `$dependentsByPolicyId`

### DIFF
--- a/src/data/dependents.js
+++ b/src/data/dependents.js
@@ -13,7 +13,8 @@ export const dependentsByPolicyId = writable({})
  * @return {Object}
  */
 export async function addDependent(policyId, depData) {
-  start(policyId)
+  const urlPath = `policies/${policyId}/dependents`
+  start(urlPath)
   
   let parsedDep = {
     name: depData.name,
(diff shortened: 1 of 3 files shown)


src/data/dependents.js
```diff
@@ -1,104 +1,115 @@
...
-    })
-
-    stop(policyId)
-
-    return parsedDep
-} 
+  const urlPath = `policies/${policyId}/dependents`
+  start(urlPath)
+  
+  let parsedDep = {
```
Schparky, line 19:
```suggestion
  const parsedDep = {
```

src/data/dependents.js
```diff
@@ -1,104 +1,115 @@
...
+  
   // TODO: uncomment when endpoint is finished
-  // let dpndt = await UPDATE(`dependents/${depId}`)
-
-  dependents.update(currDeps => {
-    let i = currDeps.findIndex(dep => dep.id === depId)
-    currDeps[i] = parsedDep
-    return currDeps
+  // const updatedDependent = await UPDATE(urlPath)
+  const updatedDependent = parsedDep // TEMP - until we can use API return value.
```
Schparky, line 85: What is incorrect in the return value?

Schparky (approved):

src/data/dependents.js
forevermatt: It doesn't look like that endpoint exists yet, so... there is no return value yet 🙂

PR #79, opened 2021-08-17
Use a static id instead of generic loading

hobbitronics:
Since we should only be making one loadClaims or loadPolices call at a time, a static id should do.

### DIFF
--- a/src/pages/items/new.svelte
+++ b/src/pages/items/new.svelte
@@ -1,7 +1,7 @@
 <script>
 import user from '../../authn/user.js'
 import { addItem } from '../../data/items.js'
-import { dependents, loadDependents, initialized as depsInitialized } from '../../data/dependents.js'
+import { dependentsByPolicyId, loadDependents } from '../../data/dependents.js'
 import { categories as categoryOptions, init, initialized as catItemsInitialized } from '../../data/itemCategories'
 import { Breadcrumb, Description, MoneyInput } from '../../components'
 import { goto } from '@roxi/routify'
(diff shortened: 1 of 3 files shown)


Schparky (approved):
I wonder why your IDE isn't adding a newline consistently. I expected that the .editorconfig would make that happen.

hobbitronics:
going to fix some breaking changes that we missed on the refactor of dependents.js, nothing huge, just importing things that aren't there

src/pages/items/new.svelte
```diff
@@ -27,9 +27,9 @@ let today = new Date()
 
 $: formData.coverage_start_date = today.toISOString().slice(0, 10) //api requires yyyy-mm-dd
 $: formData.purchase_date = formData.coverage_start_date
-$: accountablePersons = $dependents //TODO add current User to this: = [$dependents..., currentUser]
+$: accountablePersons = $dependentsByPolicyId //TODO add current User to this: = [$dependents..., currentUser]
```
forevermatt, line 30: `$dependentsByPolicyId` isn't actually an array. It's an object with policy-id keys which each have an array as their value.

You may need to do something like this...
```suggestion
$: accountablePersons = $dependentsByPolicyId[$user.policy_id] || [] //TODO add current User to this: = [$dependents..., currentUser]
```

src/pages/items/new.svelte
```diff
@@ -27,9 +27,9 @@ let today = new Date()
...
 $: if ($catItemsInitialized) categories = $categoryOptions.length ? $categoryOptions : [{name: 'Electronics', id: '63bcf980-e1f0-42d3-b2b0-2e4704159f4f'}] //TODO categoriesOptions isn't hydrating yet, remove mock data
-$: ! $depsInitialized && $user.policy_id && loadDependents($user.policy_id)
+$: $dependentsByPolicyId.length && $user.policy_id && loadDependents($user.policy_id)
```
forevermatt, line 32: I'd suggest taking out the `$dependentsByPolicyId.length &&` condition from the beginning of this, and simply load the dependents once we have the `$user.policy_id`. Loading them again if we later visit this view doesn't seem like it will cause any problems, and simplifies this code noticeably.

src/pages/items/new.svelte
hobbitronics: nice, everything is back to the level of functioning it was before.

forevermatt (approved):

hobbitronics:
> I wonder why your IDE isn't adding a newline consistently. I expected that the .editorconfig would make that happen.

Maybe its something to do with my settings or prettier interfering. I tried turning format on save on, but it did some things I didn't expect from the config file.

PR #82, opened 2021-08-18
Group items data by policy ID

forevermatt:
### Changed
- Refactor items store to be grouped by policy ID

### Fixed
- Update home page to use items-by-policy-id
- Update item-details page to use items-by-policy-id
- Update new-claim page to use items-by-policy-id

### DIFF
--- a/src/data/items.js
+++ b/src/data/items.js
@@ -1,36 +1,40 @@
-import { CREATE, DELETE, GET, UPDATE } from "./index.js"
+import { CREATE, DELETE, GET, UPDATE } from "."
+import { start, stop } from "../components/progress"
 import { throwError } from "../error"
-import { start, stop } from "../components/progress/index.js"
 import { writable } from "svelte/store"
 
-export const items = writable([])
+export const itemsByPolicyId = writable({})
 
 /**
+ * Load the items for the specified policy.
  *
- * @description a function to fetch the items of a policy
  * @export
  * @param {string} policyId -- The UUID for the desired policy
- * @return {Object} 
  */
-export async function getItems(policyId) {
-  start(policyId)
+export async function loadItems(policyId) {
+  const urlPath = `policies/${policyId}/items`
+  start(urlPath)
 
-  const items = await GET(`policies/${policyId}/items`)
+  const items = await GET(urlPath)
+  itemsByPolicyId.update(data => {
+    data[policyId] = items
+    return data
+  })
 
-  stop(policyId)
-  return items
+  stop(urlPath)
 }
 
 /**
+ * Add an item.
  *
- * @description a function to create an item
  * @export
- * @param {Number} policyId
+ * @param {string} policyId -- The UUID for the applicable policy
  * @param {Object} itemData
  * @return {Object} 
  */
 export async function addItem(policyId, itemData) {
-  start(policyId)
+  const urlPath = `policies/${policyId}/items`
+  start(urlPath)
 
   const parsedItemData = {
     category_id: itemData.category,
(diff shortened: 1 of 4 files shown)


src/data/items.js
```diff
@@ -47,50 +51,70 @@ export async function addItem(policyId, itemData) {
...
 
-  const item = await CREATE(`policies/${policyId}/items`, parsedItemData)
-
-  stop(policyId)
+  const addedItem = await CREATE(urlPath, parsedItemData)
 
-  items.update(data => {
-    data.push(item)
+  itemsByPolicyId.update(data => {
+    const items = data[policyId] || []
```
hobbitronics, line 57: just curious if `data[policyId] = [data[policyId]..., addedItem]` would work since it would be pass by reference.

forevermatt: I'm not sure. I think the existing code is enough simpler / easy-to-understand that I'd prefer to leave it as-is.

src/pages/items/[itemId].svelte
```diff
@@ -74,21 +41,26 @@ p {
 }
 </style>
 
-{#if items && !item } 
-  Item does not exist!
-{:else if items }
+{#if !item.id } 
+  {#if $loading}
```
Comment (hobbitronics) [inline src/pages/items/[itemId].svelte:45]: would it be better to use $isLoadingById() here? I can't recall where the parent is getting the item id.

src/pages/items/[itemId]/new-claim.svelte
```diff
@@ -74,18 +76,18 @@ let formData = {
...
+$: item = items.find(itm => itm.id === itemId) || {}
 
 // TODO: get accountable person from item 
 // TODO: add reimbursed value
 $: isNotRepairableOrMoneyInputsAreSet = (formData.isRepairable !== "repairable" || (formData.repairCost && formData.fairMarketValue))
 $: seventyPercentCheck = (!formData.repairCost || !formData.fairMarketValue || (formData.repairCost/formData.fairMarketValue) >= .7)
 $: payoutOptionCheck = formData.lossReason && isNotRepairableOrMoneyInputsAreSet && seventyPercentCheck
 $: canRepair = formData.lossReason === "impact" || formData.lossReason === "lightning" || formData.lossReason === "water_damage" || formData.lossReason === "other"
-$: claimExists = $claims.find(clm => clm.itemId === $params.itemId)
+$: claimExists = $claims.find(clm => clm.itemId === itemId)
```
Comment (hobbitronics) [inline src/pages/items/[itemId]/new-claim.svelte:90]: this name implies it is a boolean, yet find will return the first matching item.

src/data/items.js
hobbitronics: ok, I personally find the one liner easier to understand at least when my attention span is reduced do to the distractions, but I am fine with this too.

hobbitronics (approved):
Looks good, just a few questions/suggestions

src/pages/items/[itemId].svelte
Comment (forevermatt) [reply src/pages/items/[itemId].svelte:45]: I considered that, but it hits a downside with my idea of using the URL path as the start/stop key: to use `isLoadingById()`, the calling code needs to know an implementation detail that's otherwise hidden in our data files (namely, what the URL is to the loading call we care about).

I think simply using `$loading` here is acceptable, both because it keeps this code simpler and because there is little else likely to be loading at this point. We also only check `$loading` if we don't yet have the item, so any loading of other things once we have the item won't affect us here.

src/pages/items/[itemId]/new-claim.svelte
Comment (forevermatt) [reply src/pages/items/[itemId]/new-claim.svelte:90]: True. All I did on that line was change `$params.itemId` to `itemId`, so I didn't do any fixing or cleanup of the logic there, but your right about that name being misleading.

Comment (hobbitronics) [reply src/pages/items/[itemId]/new-claim.svelte:90]: maybe just add a todo?

Comment (forevermatt) [reply src/pages/items/[itemId]/new-claim.svelte]: Okay, fixed.

PR #87, opened 2021-08-18
Adjust claims page

forevermatt:
### Added
- Extract ClaimCardBanner component (for banner at top of ClaimCard)

### Fixed
- Update getState() to use a claim's status (untested)
- Fix retrieval of items on new-claim page
- Update new-claim page's check for whether an item was found

### Changed
- Pass the list of claims (not items) into the ClaimCards component
- Show a ClaimCard for each item on each claim, not just for each claim
- Change ClaimCard(s) to dispatch events rather than calling `$goto`, since they're just components
- Update "claims" page to listen for events and handle URL changes, since it's a page

---

We're waiting on some changes on the API (returning the list of items with the claim's data), so I haven't fully tested this, but it seems to build successfully and work as far as I can manually test.

### DIFF
--- a/src/components/ClaimCardBanner.svelte
+++ b/src/components/ClaimCardBanner.svelte
@@ -1,0 +1,19 @@
+<script>
+export let item = {}
+export let state = {}
+
+$: bgColor = state.bgColor || ''
+$: color = state.color || ''
+$: icon = state.icon || ''
+$: message = item.message || ''
+$: title = state.title || ''
+</script>
+
+<div class="flex justify-start align-items-center black mb-2 p-1 pl-50px" style="background: var({bgColor});">
+  <span class="material-icons" style="color: var({color});">{icon}</span>
+
+  <div class="mdc-theme--primary pl-10px">
+    <div class="mdc-typography--headline6 multi-line-truncate content" style="color: var({color});">{title}</div>
+    <div class="multi-line-truncate fs-14" style="color: var({color});">{message}</div>
+  </div>
+</div>
--- a/src/components/ClaimCardBanner.svelte
+++ b/src/components/ClaimCardBanner.svelte
@@ -1,6 +1,10 @@
 <script>
+import { getState } from '../data/claims'
+
+export let claim = {}
 export let item = {}
-export let state = {}
+
+$: state = getState(claim)
 
 $: bgColor = state.bgColor || ''
 $: color = state.color || ''
--- a/src/components/ClaimCardBanner.svelte
+++ b/src/components/ClaimCardBanner.svelte
@@ -13,16 +13,6 @@ $: message = item.message || ''
 $: title = state.title || ''
 </script>
 
-<style>
-.pl-10px {
-  padding-left: 10px;
-}
-
-.pl-50px {
-  padding-left: 50px;
-}
-</style>
-
 <div class="flex justify-start align-items-center black mb-2 p-1 pl-50px" style="background: var({bgColor});">
   <span class="material-icons" style="color: var({color});">{icon}</span>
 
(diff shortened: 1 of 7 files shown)


src/components/ClaimCards.svelte
```diff
@@ -1,22 +1,22 @@
...
   flex-basis: 330px;
 }
 </style>
 
 <div class="flex justify-start flex-wrap {$$props.class}">
-  {#each items as item}
-    <div class="cards">
-      <ClaimCard state={ getState(item) } {item} buttons={[ { label: "Edit coverage", url: "/items/edit-coverage" } ]} />
-    </div>
+  {#each claims as claim}
```
Schparky, line 15: Do either of these need to be a [keyed](https://svelte.dev/tutorial/keyed-each-blocks) `each`? I can't quickly discern whether it matters here.

Schparky (approved):

src/components/ClaimCardBanner.svelte
```diff
@@ -0,0 +1,33 @@
...
+
+$: bgColor = state.bgColor || ''
+$: color = state.color || ''
+$: icon = state.icon || ''
+$: message = item.message || ''
+$: title = state.title || ''
+</script>
+
+<style>
+.pl-10px {
```
hobbitronics, line 17: this is now a global class

src/pages/claims.svelte
```diff
@@ -2,8 +2,18 @@
...
 $: $initialized || loadClaims()
+
+const onEditClaim = event => {
+  const claimId = event.detail
+  $goto(`/claims/${claimId}/edit`)
+}
+const onGotoClaim = event => {
+  const claimId = event.detail
+  $goto(`/claims/${claimId}`)
+}
```
hobbitronics, line 16: So will going to the claim show the applicable actions shown [here](https://www.figma.com/file/FPjZNKp9JomrGF4CnG5Uvj/Cover?node-id=129%3A30847)?

src/pages/items/[itemId]/new-claim.svelte
```diff
@@ -78,7 +78,7 @@ let items
 let item
 
 $: $user.policy_id && loadItems($user.policy_id)
-$: items = itemsByPolicyId[$user.policy_id] || []
+$: items = $itemsByPolicyId[$user.policy_id] || []
```
Comment (hobbitronics) [inline src/pages/items/[itemId]/new-claim.svelte:81]: this fixed the problem I was running into yesterday.

src/components/ClaimCards.svelte
```diff
@@ -1,22 +1,22 @@
...
+    {#each (claim.items || []) as item}
+      <div class="card">
+        <ClaimCard {claim} {item} on:edit-claim on:goto-claim />
+      </div>
+    {/each}
   {/each}
```
hobbitronics, line 21: it looks like you have 2 {/each}'s but I only see one being opened.

hobbitronics (approved):

src/components/ClaimCardBanner.svelte
forevermatt: Okay, I'll remove those. Just for my edification, where are those defined globally? I vaguely remember you adding some dynamic code to generate those, but I don't see them anywhere.

src/components/ClaimCards.svelte
forevermatt: Good idea. I don't know whether it will matter here (it seems mostly to matter when the list will dynamically change while the user is viewing that page), but as a general practice it seems like a good idea to always key our uses of `#each`.

src/pages/claims.svelte
forevermatt: I'm not sure what you're referring to, sorry. (That link seems to go to the entire "Customer Claims" area in Figma, not a particular screen mockup.)

Going to the claim will show whatever we decide to show when the user is viewing one of their claims. This is just to let them get there from this claim card.

PR #90, opened 2021-08-19
add ClaimBanner

hobbitronics:
- created ClaimBanner based on SuccessBanner and ClaimCardBanner so we wont have to create a different banner component for each state
- deleted SuccessBanner and WarningBanner


src/pages/claims/[claimId].svelte
```diff
@@ -1,6 +1,17 @@
 <script>
-import { Banner, MoneyInput, Row, SuccessBanner, WarningBanner } from '../../components'
+import user from '../../authn/user.js'
+import { Banner, MoneyInput, Row, ClaimBanner } from '../../components'
+import { loadClaims, claims } from '../../data/claims'
+import { loadItems, itemsByPolicyId } from '../../data/items'
 import { Button, Page } from '@silintl/ui-components'
+import { params } from '@roxi/routify'
+
+loadClaims()
```
Comment (hobbitronics) [inline src/pages/claims/[claimId].svelte:9]: this is all temporary, stashed it initially then realized I had a broken build from deleting SuccessBanner which was being used here.

src/components/banners/ClaimBanner.svelte
```diff
@@ -0,0 +1,33 @@
...
+export let claim = {}
+export let item = {}
+
+$: state = getState(claim) || {}
+
+$: bgColor = state.bgColor || ''
+$: color = state.color || ''
+$: icon = state.icon || ''
+$: title = state.title || ''
+$: message = item.message || ''
```
hobbitronics, line 14: I think `message = claim.help_text` might be correct in which case we won't need item

hobbitronics, line 14: ok, I i'm just going to get rid of message and put a slot in its place for now.

src/pages/claims/[claimId].svelte
```diff
@@ -1,32 +1,37 @@
 <script>
-import { Banner, MoneyInput, Row, SuccessBanner, WarningBanner } from '../../components'
+import { Banner, MoneyInput, Row, ClaimBanner } from '../../components'
+import { loadClaims, claims, initialized } from '../../data/claims'
 import { Button, Page } from '@silintl/ui-components'
+import { params } from '@roxi/routify'
```
Comment (hobbitronics) [inline src/pages/claims/[claimId].svelte:5]: sort imports

src/pages/items/[itemId].svelte
```diff
@@ -52,7 +56,7 @@ p {
   <Breadcrumb links={breadcrumbLinks} />
   <h1>{item.name}</h1>
   <h3>{item.make} {item.model}</h3>
-  <Banner background="var(--mdc-theme-neutral">{item.category?.name}</Banner>
+  <Banner background="var(--mdc-theme-neutral" class="width">{item.category?.name}</Banner>
```
Comment (forevermatt) [inline src/pages/items/[itemId].svelte:59]: I don't think this `class="width"` will work, since the definition of that CSS class is in this component (and will be scoped to only affect this component, not child components like Banner).

Comment (hobbitronics) [reply src/pages/items/[itemId].svelte:59]: oh yeah, I will move it to global.css

forevermatt (approved):

PR #91, opened 2021-08-19
Show a claim card for each item on each claim

forevermatt:
### Changed
- Pass `claim` and `claimItem` (not `item`) data to ClaimCard

### Fixed
- Update claim states to use TitleCase (since that's what we get back from the API)
- Ensure state always contains an object (in ClaimCard)

---

@hobbitronics, we'll probably have conflicts between our two feature branches. If you want to go ahead and merge yours first, I'll deal with them in mine.


src/data/claims.js
```diff
@@ -167,4 +167,9 @@ export async function loadClaims() {
   initialized.set(true)
 }
 
-export const getState = claim => states[claim.status || 'message']
+export const getState = claim => {
+  if (states[claim.status] === undefined) {
+    console.error('No such state (for claim status):', claim.status, Object.keys(states))
+  }
+  return states[claim.status || 'Message']
```
hobbitronics, line 174: I don't think [claim.status || 'Message'] was working so I had changed it to `states[claim.status] || {}` unless we want there to be a default. I will undo my change to this to avoid merge conflicts though.

hobbitronics (approved):

src/data/claims.js
forevermatt: Okay, that works. Thanks. Did you see my comment to you about that, in the description of this PR? (I'm beginning to suspect adding comments in a PR's description isn't a reliable way to communicate...)

hobbitronics: yes I did, but since I still needed to change a few things I thought I would just undo my small change to make it easier for you.

PR #93, opened 2021-08-20
Real data in claim view

hobbitronics:
Replaced static content with real data
added dates.js
changed getState parameter from claim to claim.status
renamed some states to match api

image

### DIFF
--- a/src/components/ClaimCard.svelte
+++ b/src/components/ClaimCard.svelte
@@ -19,7 +19,7 @@ $: items = $itemsByPolicyId[policyId] || []
 $: item = items.find(i => i.id = claimItem.item_id) || {}
 $: msAgo = now - Date.parse(claimItem.updated_at)
 $: daysAgo = msAgo > 0 ? Math.floor(msAgo/day) : '-'
-$: state = getState(claim) || {}
+$: state = claim.status && getState(claim.status) || {}
 
 const editClaim = () => dispatch('edit-claim', claim.id)
 const gotoClaim = () => dispatch('goto-claim', claim.id)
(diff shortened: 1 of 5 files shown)


src/components/ClaimCard.svelte
```diff
@@ -19,7 +19,7 @@ $: items = $itemsByPolicyId[policyId] || []
 $: item = items.find(i => i.id = claimItem.item_id) || {}
 $: msAgo = now - Date.parse(claimItem.updated_at)
 $: daysAgo = msAgo > 0 ? Math.floor(msAgo/day) : '-'
-$: state = getState(claim) || {}
+$: state = claim.status && getState(claim.status) || {}
```
forevermatt, line 22: It's not obvious whether this will return `undefined` or `{}` if the claim has no status. Why the `claim.status` check before the call to `getState()`? I think it will handle undefined just fine.

src/components/dates.js
```diff
@@ -0,0 +1,8 @@
+export const formatDate = (dateString, ...others) => {
```
forevermatt, line 1: What would you think of simplifying this so that all we give it is the date string, and internally we control what format we put it into (so that our dates are consistent throughout the app)?

src/components/ClaimCard.svelte
hobbitronics: before I added it, getState was being called with undefined causing an error. An alternative to this would be passing it a default state if claim.status was undefined.

forevermatt: I think with your corresponding change to `getState()` it will no longer throw an error. However, if you don't mind wrapping this check in parentheses to make it clear we'll always get an object back (not undefined) for `state`, that's also fine:
```suggestion
$: state = (claim.status && getState(claim.status)) || {}
```

src/components/dates.js
hobbitronics: That would be fine, as usually we will stick with a consistent format. The caveat is that when 'Default' was passed to it, it will go with the browser language to determine the order of date, month and year. In the mockup I think it was '01 Jan 2001'. So if we want all users to get that format we would have to use a specific language instead of default.

src/components/ClaimCard.svelte
hobbitronics: You are right about it no longer throwing an error since it only fires off when claim.status changes.

forevermatt (approved):
Looks good 👍

PR #94, opened 2021-08-23
added Needs_initial_changes Theft claim view

hobbitronics:
Added a form to upload receipt and replacement cost for status of Needs_initial_changes and event_type of theft.
This will also likely be used for other replacement views, but we need information from the api to know if it is a replacement.
Since there isn't working endpoint to this yet it will need to be finished later

### DIFF
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -1,11 +1,17 @@
 <script>
 import user from '../../authn/user.js'
 import { Banner, MoneyInput, Row, ClaimBanner } from '../../components'
+import { formatDate } from '../../components/dates.js'
 import { loadClaims, claims, initialized } from '../../data/claims'
 import { loadItems, itemsByPolicyId } from '../../data/items'
 import { goto, params } from '@roxi/routify'
-import { Button, Page } from '@silintl/ui-components'
-import { formatDate } from '../../components/dates.js'
+import { Button, Form, Page } from '@silintl/ui-components'
+
+const formData = new FormData()
+
+let file = {}
+let replacementCost
+let uploading = false
 
 $: ! $initialized && loadClaims()
 
@@ -17,11 +23,34 @@ $: item = items.find(itm => itm.id === claimItem.item_id) || {}
 $: eventDate = formatDate(claim.event_date)
 
 const onClick = () => $goto(`claims/${$params.claimId}/edit)`)
+
+const onSubmit = () => {
+  formData.append('replacement_cost', replacementCost)
+  
+  console.log(formData) //TODO update claim with replacementCost and file to api
+}
+
+async function chosen(event) {
+  formData.append('file', event.target.files[0])
+
+  try {
+    uploading = true
+
+    // file = await upload(formData) //Todo verify the backend will upload files with this method
+
+  } finally {
+    uploading = false
+  }
+}
 </script>
 
 <style>
 .left-detail {
   margin: 0.5rem 0;
+}
+
+.label {
+  color: hsla(219, 53%, 7%, .6);
 }
 
 </style>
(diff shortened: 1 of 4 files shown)


src/pages/claims/[claimId].svelte
```diff
@@ -17,13 +23,36 @@ $: item = items.find(itm => itm.id === claimItem.item_id) || {}
...
+  console.log(formData) //TODO update claim with replacementCost and file to api
+}
+
+async function chosen(event) {
+  formData.append('file', event.target.files[0])
+
+  try {
+    uploading = true
+
+    // file = await upload(formData) //Todo verify the backend will upload files with this method
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:39]: I believe this is now functional. Let me know if you have any trouble with it.

src/pages/claims/[claimId].svelte
```diff
@@ -47,15 +76,32 @@ const onClick = () => $goto(`claims/${$params.claimId}/edit)`)
...
       <!-- TODO get the actual maximum amount -->
       {(item.coverage_amount * .7)  || ''}
     </p>
     <p>
       <Button on:click={onClick} outlined>Edit claim</Button>
     </p>
-    <p>
-      <MoneyInput label="Actual cost of replacement"></MoneyInput>
-    </p>
+    {#if claim.status === 'Needs_initial_changes' && claim.event_type === 'Theft'}
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:86]: Is there a way to make these strings `const` values, or is there no benefit of doing that in Javascript? Sorry, I'm in Go mode so not shifting gears well.

src/pages/claims/[claimId].svelte
```diff
@@ -17,13 +23,36 @@ $: item = items.find(itm => itm.id === claimItem.item_id) || {}
 $: eventDate = formatDate(claim.event_date)
 
 const onClick = () => $goto(`claims/${$params.claimId}/edit)`)
+
+const onSubmit = () => {
+  formData.append('replacement_cost', replacementCost)
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:28]: Not sure where is the best place, but you'll want to convert from fractional dollars (1.00) to integer cents (100) at some point.

Schparky (approved):

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:86]: There may be at some point. We have a states object, which needs to be updated, but I don't see getting the key from that to use here as very efficient. I Think I will just move this condition to a reactive statement to clean up the markup for now and as we expand the claim views it might make more sense to do something different.

src/pages/claims/[claimId].svelte
Comment (forevermatt) [reply src/pages/claims/[claimId].svelte:29]: I'd be tempted for us to use cents everywhere, only converting to decimal dollars when rendering it out to show the user. For a separate (open source) project I've worked on, I've done the conversion between textual decimal money input into cents (in Svelte), so I can help with that if we choose to go with that design.

src/pages/claims/[claimId].svelte
```diff
@@ -15,15 +21,39 @@ $: items = $itemsByPolicyId[$user.policy_id] || []
 $: ! items.length && loadItems($user.policy_id)
 $: item = items.find(itm => itm.id === claimItem.item_id) || {}
 $: eventDate = formatDate(claim.event_date)
+$: needsInitialChangesAndReplacement = claim.status === 'Needs_initial_changes' && claim.event_type === 'Theft'
 
 const onClick = () => $goto(`claims/${$params.claimId}/edit)`)
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:26]: It might be helpful to rename this from `onClick` to something like `gotoEdit` or `editClaim` to make it more indicative of what it does, since in theory there could be multiple things on a page that respond to clicks.

forevermatt (approved):

forevermatt (approved):

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte]: Working now, thanks for the help.

PR #95, opened 2021-08-24
Extract ClaimForm, use on new-claim and edit-claim pages

forevermatt:
### Changed
- Move `DependentForm` to new "forms" subfolder in "components"
- Extract `ClaimForm` component
- Update new-claim page to use ClaimForm
- Update edit-claim page to use ClaimForm

### Added
- Add Makefile for easier / more-expected way of starting local dev

### Fixed
- Load claim-event-types dynamically
- Fix "Edit claim" button on home page's ClaimCards
- Round "Cash now" display to nearest cent


src/components/forms/ClaimForm.svelte
```diff
@@ -0,0 +1,186 @@
...
+$: !payoutOptionCheck && unSetPayoutOption()
+$: !(repairableSelection === "repairable" || payoutOption === "cash_now") && unSetFairMarketValue()
+$: repairableSelection !== "repairable" && unSetRepairCost()
+$: !isPotentiallyRepairable && unSetIsRepairable()
+$: payoutOptionCheck && payoutOption == "evacuation" && unSetPayoutOption()
+
+$: moneyPayoutOptions = [
+  {
+    // TODO: make this the covered amount
+    label: `Replace and get reimbursed later `,
```
Schparky, line 61:
```suggestion
    label: `Replace and get reimbursed later`,
```

Schparky (approved):

src/components/forms/ClaimForm.svelte
```diff
@@ -0,0 +1,186 @@
...
+let repairCost = ''
+let payoutOption = ''
+
+// Set initial values based on the provided claim and claim-item data.
+$: setInitialValues(claim, claimItem)
+
+// TODO: get accountable person from item 
+// TODO: add reimbursed value
+$: isNotRepairableOrMoneyInputsAreSet = (repairableSelection !== "repairable" || (repairCost && fairMarketValue))
+$: seventyPercentCheck = (!repairCost || !fairMarketValue || (repairCost/fairMarketValue) >= .7)
```
hobbitronics, line 48: I think something like isOverSeventyPercent or similar would be a more meaningful.

src/components/forms/ClaimForm.svelte
```diff
@@ -0,0 +1,186 @@
...
+}
+const unSetIsRepairable = () => {
+  repairableSelection = null
+}
+const unSetRepairCost = () => {
+  repairCost = null
+}
+</script>
+
+<style>
```
hobbitronics, line 120: should we get rid of the <style> tags?

src/components/forms/ClaimForm.svelte
```diff
@@ -0,0 +1,186 @@
...
+      </div>
+    {/if}
+    {#if repairableSelection === "repairable"}
+      <p>
+        <MoneyInput label="Cost of repair" bind:value={repairCost} />
+        <Description>
+          How much will it cost to be repaired?
+          <br />
+          To convert to USD, use
+          <a href="https://www.google.com/search?q=currency+converter" target="_blank">this converter</a>.
```
hobbitronics, line 151: since we are using this url in more than one place we should probably set it as a const somewhere.

src/components/forms/ClaimForm.svelte
forevermatt: I think captaincrazybro name it this because it's more than just a seventy-percent calculation, though that's the core of it (if the other conditions are met). The current name is a bit ambiguous, agreed, I just don't want to name it something that implies it's simpler than it really is.

src/components/forms/ClaimForm.svelte
```diff
@@ -0,0 +1,186 @@
...
+        </div>
+      {:else}
+        <div>
+          <p>We are sorry you are experiencing this situation and will keep you in our prayers.</p>
+          <p>We will reach out to SIL HR to get more context about this situation.</p>
+          <p>If approved, you are eligible for 2/3 payout of covered lost assets.</p>
+        </div>
+      {/if}
+    {/if}
+    {#if repairableSelection === "repairable" && (isNotRepairableOrMoneyInputsAreSet && !seventyPercentCheck)}
```
hobbitronics, line 178: this line might benefit from moving the conditional to a function or reactive statement with a meaningful name.

src/components/forms/ClaimForm.svelte
hobbitronics: how about overSeventyPercentCheck? I just think the current name is ambiguous in the wrong way. It could be over or under.

src/data/claim-event-types.js
```diff
@@ -0,0 +1,16 @@
...
+
+export async function loadClaimEventTypes() {
+  const urlPath = 'config/claim-event-types'
+  start(urlPath)
+
+  const results = await GET(urlPath)
+
+  stop(urlPath)
+  
+  claimEventTypes.set(results)
```
hobbitronics, line 15: would it ever be useful to return the results here or should we always be using the store?

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -0,0 +1,61 @@
+<script>
+import user from '../../../authn/user'
+import { Breadcrumb, ClaimForm } from '../../../components'
+import { loading } from '../../../components/progress'
+import { claims, initialized, loadClaims, updateClaim } from '../../../data/claims.js'
+import { itemsByPolicyId, loadItems } from '../../../data/items.js'
+import { goto } from '@roxi/routify'
+import { Page } from '@silintl/ui-components'
+
+export let claimId
```
Comment (hobbitronics) [inline src/pages/claims/[claimId]/edit.svelte:10]: I am wondering why we need an export on a page view. Does this file get used as a component or is there some other use exports have in a svelte file I am not aware of?

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -0,0 +1,61 @@
...
+    name: "This Claim",
+    url: `/claims/${claimId}`
+  },
+  {
+    name: "Edit",
+    url: `/items/${claimId}/edit`
+  }
+]
+
+$: $initialized || loadClaims()
```
Comment (hobbitronics) [inline src/pages/claims/[claimId]/edit.svelte:28]:
```suggestion
$: ! $initialized || loadClaims()
```

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -0,0 +1,61 @@
...
+$: claim = $claims.find(c => c.id === claimId) || {}
+$: claimItems = claim.claim_items || []
+
+/** @todo Update this when claims can have multiple items. */
+$: claimItem = claimItems[0] || {}
+$: itemId = claimItem.item_id
+
+$: $user.policy_id && loadItems($user.policy_id)
+$: items = $itemsByPolicyId[$user.policy_id] || []
+$: item = items.find(anItem => anItem.id === itemId) || {}
```
Comment (hobbitronics) [inline src/pages/claims/[claimId]/edit.svelte:38]: lines 28 through 30 seem to be used in at least in one other place. Would it be possible to make this modular at some point? `<ClaimAndItemsData {claimId} {policyId}> page view slot <ClaimAndItemsData>` then pass the needed data back in with slot props and the let directive?  We could also use a plain old js file as an intermediate data layer to return this set of data. Maybe more work than its worth for now.

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+const onSubmit = async event => {
+  await updateClaim(itemId, event.detail)
+  $goto(`/claims/${claimId}`)
+}
+</script>
+
+{#if $loading }
+  Loading...
+{:else if claims && !claim.id}
```
Comment (hobbitronics) [inline src/pages/claims/[claimId]/edit.svelte:48]: do we need to check the status to see if the claim.status is beyond the point where the user can edit the claim i.e. 'Approved', 'Pending_final_review', etc. I don't remember where that line is or if we have decided fully, but maybe just add a TODO to check and let the user know they can't edit at this point.

src/pages/claims/[claimId]/edit.svelte
Comment (forevermatt) [reply src/pages/claims/[claimId]/edit.svelte:10]: The export is a way to get the URL parameter, and it seems cleaner than using `$params` to me, but that's more of a preference.

hobbitronics (approved):
Looks good. Approved with a few questions, suggestions.

src/data/claim-event-types.js
forevermatt: Always using the store seems better. It helps keep us writing declarative code ("However many of these there are, show them in this way") rather than imperative ("Update that, get the results, go change things that had the old list but are now out of date...").

src/pages/claims/[claimId]/edit.svelte
Comment (forevermatt) [reply src/pages/claims/[claimId]/edit.svelte:38]: It feels like that would add more complexity than it would be worth for just getting rid of these few duplicated lines at this point, in my opinion.

PR #109, opened 2021-08-27
added householdId field to settings

hobbitronics:
- Added to settings view:
  - TextField
  - Snackbar
  - call to updatePolicy endpoint

image

### DIFF
--- a/src/pages/household/settings.svelte
+++ b/src/pages/household/settings.svelte
@@ -2,9 +2,14 @@
 import user from '../../authn/user'
 import { Breadcrumb } from "../../components"
 import { dependentsByPolicyId, loadDependents } from '../../data/dependents'
+import { updatePolicy } from '../../data/policies'
 import { loadMembersOfPolicy, membersByPolicyId } from '../../data/policy-members'
 import { goto } from "@roxi/routify"
-import { Button, IconButton, Page } from "@silintl/ui-components"
+import { Button, TextField, IconButton, Page, Snackbar, setNotice } from "@silintl/ui-components"
+
+const policyData = {}
+
+let householdId = ''
 
 $: policyId = $user.policy_id
 $: if (policyId) {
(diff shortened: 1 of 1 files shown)


src/pages/household/settings.svelte
```diff
@@ -15,6 +20,16 @@ $: if (policyId) {
 $: dependents = $dependentsByPolicyId[policyId] || []
 $: householdMembers = $membersByPolicyId[policyId] || []
 
+const udpateHouseholdId = () => {
+  if(householdId.length === 7) {
```
hobbitronics, line 24: I'm going to replace this with regex `/[0-9]+/` and get rid of the length

src/pages/household/settings.svelte
```diff
@@ -15,6 +20,20 @@ $: if (policyId) {
 $: dependents = $dependentsByPolicyId[policyId] || []
 $: householdMembers = $membersByPolicyId[policyId] || []
 
+const udpateHouseholdId = () => {
+  householdId = householdId.replaceAll(' ', '')
+
+  if(validateId(householdId)) {
+    policyData.household_id = householdId
+    
+    updatePolicy(policyId, policyData)
```
forevermatt, line 29: You will probably want an `await` here, so that if the `updatePolicy()` call fails we don't tell the user it succeeded.

src/pages/household/settings.svelte
```diff
@@ -15,6 +20,20 @@ $: if (policyId) {
 $: dependents = $dependentsByPolicyId[policyId] || []
 $: householdMembers = $membersByPolicyId[policyId] || []
 
+const udpateHouseholdId = () => {
```
forevermatt, line 23: Looks like you did the same typo I regularly do: `udpate` vs `update`

src/pages/household/settings.svelte
hobbitronics: ahh, good point

src/pages/household/settings.svelte
hobbitronics: nice catch

src/pages/household/settings.svelte
```diff
@@ -2,9 +2,14 @@
 import user from '../../authn/user'
 import { Breadcrumb } from "../../components"
 import { dependentsByPolicyId, loadDependents } from '../../data/dependents'
+import { policies, updatePolicy } from '../../data/policies'
```
forevermatt, line 5: Was this `policies` accidentally added? I don't see it being used anywhere

hobbitronics: oops, that's part of the next PR.

PR #111, opened 2021-08-31
update affiliate organization

hobbitronics:
Added:
- `<datalist>` for Affiliate organizaiton (like select but searchable)
- styling for the input
- function for updating policies.entity_code
- control block for the affiliatiton field

I will be putting the datalist element into a component next, but wanted to see if it served our purposes first.

image

### DIFF
--- a/src/pages/household/settings.svelte
+++ b/src/pages/household/settings.svelte
@@ -151,18 +151,20 @@ const isYou = householdMember => householdMember.id === $user.id
   <p>
     <TextField placeholder={'1234567'} autofocus bind:value={householdId} on:blur={updateHouseholdId} />
   </p>
-  
-  <h3 class="ml-1 mt-3" >Affiliation<span class="required">*</span></h3>
-  <label class="custom-field">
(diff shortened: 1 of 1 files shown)


src/pages/household/settings.svelte
```diff
@@ -40,7 +47,26 @@ const updateHouseholdId = async () => {
   }
 }
 
-const validateId = sanitizedId => sanitizedId.length && sanitizedId.split('').every(digit => /[0-9]/.test(digit))
+const updateAffiliation = async () => {
+  for (const [key, value] of Object.entries(affiliations)){
+    if(affiliationChoice === value && affiliationChoice !== policyData.entity_code) {
```
Schparky, line 52: I'm struggling to understand the second condition: `affiliationChoice !== policyData.entity_code`. Why would `affiliationChoice` ever be equal to the entity code?

hobbitronics: that should be key instead of affiliation code, thanks.

Schparky (approved):

PR #112, opened 2021-08-31
SearchableSelect component

hobbitronics:
Changed:
moved datalist to SearchableSelect component

### DIFF
--- a/src/components/SearchableSelect.svelte
+++ b/src/components/SearchableSelect.svelte
@@ -17,7 +17,6 @@ const chosen = () => {
 </script>
 
 <style>
-  /* datalist styling */
 *, *::before, *::after {
   box-sizing: border-box;
 }
--- a/src/components/SearchableSelect.svelte
+++ b/src/components/SearchableSelect.svelte
@@ -3,6 +3,7 @@ import { createEventDispatcher } from "svelte"
 
 export let options = {}
 export let choice = ''
+export let placeholder = ''
 
 const dispatch = createEventDispatcher()
 
@@ -62,7 +63,7 @@ const chosen = () => {
 
 <label class="custom-field">
   <input class="fs-14" list="options" placeholder="&nbsp;" bind:value={choice} on:change={chosen}/>
-  <span class="placeholder">Your entity of affiliation</span>
+  <span class="placeholder">{placeholder}</span>
 </label>
 
 <datalist id="options">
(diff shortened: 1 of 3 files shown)


src/pages/household/settings.svelte
```diff
@@ -47,16 +47,15 @@ const updateHouseholdId = async () => {
...
 
-      policyData.entity_code = key
+  if(choice !== policyData.entity_code) {
+    policyData.entity_code = choice
 
-      await callUpdatePolicy()
-      
-      setNotice('Your affiliation has been saved')
-    }
+    await callUpdatePolicy()
```
forevermatt, line 56: Would you mind passing in the new/updated data to this function call? That can help us trace the logic better, rather than calling parameter-less functions that reach out and get the necessary data themselves.

It's kind of like the benefits of avoiding global variables: it decouples the various pieces of your code so that each individually can be simpler and easier to understand/debug.

src/components/SearchableSelect.svelte
```diff
@@ -0,0 +1,72 @@
...
+  color: #222;
+}
+</style>
+
+<label class="custom-field" style="--field-padding: {padding};">
+  <input class="fs-14 {$$props.class}" style="width: {width}" list="options" placeholder="&nbsp;" bind:value={choice} on:change={chosen}/>
+  <span class="placeholder">{placeholder}</span>
+</label>
+
+<datalist id="options">
```
forevermatt, line 68: Now that this is extracted as a reusable component, and we could therefore have two of these on a single page, it would probably be good to make this a unique `id`. Our `generateRandomID` function (from '@silintl/ui-components/random') is specifically for that purpose.

src/components/SearchableSelect.svelte
```diff
@@ -0,0 +1,72 @@
...
+
+export let options = {}
+export let choice = ''
+export let placeholder = ''
+export let padding = '12px'
+export let width = '232px'
+
+const dispatch = createEventDispatcher()
+
+const chosen = () => {
```
forevermatt, line 12: Since this function is for responding to an event, would you mind prefixing it with "on" (i.e. "onChosen" or "onChange"). That convention seems to help maintain clarity about the purpose of a function.

Schparky (approved):

src/pages/household/settings.svelte
hobbitronics: yup, It adds a little more complexity, but probably worth it.

forevermatt (approved):

PR #113, opened 2021-09-01
Enable editing an item

forevermatt:
### Added
- Enable editing an item

### Fixed
- Key the each loop (of items) on the home page
- Detect if we ever try to update an item not in our local store
- Update dependencies
- Use a separate variable, loaded later, for setting the initial category in `ItemForm`

### Changed
- Rename several variables to conform to lowerCamelCasing convention
- Move `itemId` to separate parameter to `updateItem()` method
- Change to match the new way the API returns the category ID

### DIFF
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -1,7 +1,7 @@
 <script>
 import { ConvertCurrencyLink, Description, MoneyInput } from '../../components'
 import { dependentsByPolicyId, loadDependents } from '../../data/dependents.js'
-import { categories as categoryOptions, init, initialized as catItemsInitialized } from '../../data/itemCategories'
+import { categories, init, initialized as catItemsInitialized } from '../../data/itemCategories'
 import { loadMembersOfPolicy, membersByPolicyId } from '../../data/policy-members'
 import { Button, Form, Select, TextArea, TextField } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
@@ -28,7 +28,6 @@ let uniqueIdentifier = ''
 // Set initial values based on the provided item data.
 $: setInitialValues(item)
 
-let categories = []
 let today = new Date()
 
 $: dependents = $dependentsByPolicyId[policyId] || []
@@ -109,7 +108,7 @@ const setInitialValues = (item) => {
 
 <Form on:submit={onSubmit}>
   <p>
-    <Select label="Category" on:change={onSelectCategory} options={categories} />
+    <Select label="Category" on:change={onSelectCategory} options={$categories} />
   </p>
   <p>
     <TextField label="Short name" bind:value={shortName}></TextField>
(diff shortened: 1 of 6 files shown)


src/components/forms/ItemForm.svelte
```diff
@@ -1,35 +1,35 @@
...
-  uniqueIdentifier: '',
-}
+// Set default values.
+let categoryId = ''
+let country = ''
+let marketValueUSD = ''
+let coverageStartDate = ''
+let coverageStatus = 'Draft'
+let itemDescription = ''
+let inStorage = false  //TODO get data from somewhere
```
Schparky, line 21: Not sure why this has a TODO. It is a field available from the API. Same on `purchaseDate`

src/components/forms/ItemForm.svelte
```diff
@@ -45,55 +45,100 @@ $: policyMemberOptions = policyMembers.map(policyMember => ({
...
+  country = event.detail?.location //TODO handle when Dependents is empty, redirect to settings?
 }
 
 const onSelectCategory = event => {
-  formData.category = event.detail?.id
+  categoryId = event.detail?.id
 }
 
+const getFormData = () => {
+  return {
```
Schparky, line 61: Why not use the API property names here rather than have the `addItem` function in `data/items.js` do it? Is it just to keep the variable case standardized as camelCase?

src/pages/items/[itemId]/edit.svelte
```diff
@@ -0,0 +1,57 @@
...
+  {
+    name: "Edit",
+    url: `/items/${itemId}/edit`
+  }
+]
+
+$: policyId = $user.policy_id
+$: policyId && loadItems(policyId)
+$: items = $itemsByPolicyId[policyId] || []
+$: item = items.find(anItem => anItem.id === itemId) || {}
```
Comment (Schparky) [inline src/pages/items/[itemId]/edit.svelte:30]:
```suggestion
$: item = items.find(theItem => theItem.id === itemId) || {}
```
It's not just "any" item. :-)

Schparky (approved):

src/components/forms/ItemForm.svelte
hobbitronics: This TODO is referring to from the UI, as there isn't a form field for it yet. Maybe it won't have one, not sure.

src/components/forms/ItemForm.svelte
forevermatt: Yeah, I'm trying to keep all use of database field names as close to the "edge" of the UI codebase as possible, so that the UI code is less deeply coupled to API field names. Ideally, it will only be case-changes, though it's not exclusively that currently (e.g. serial number vs. unique identifier).

src/components/forms/ItemForm.svelte
forevermatt: I'll take a swing at going ahead and populating those using the data returned by the API. Thanks.

forevermatt: Looks like it's already taking those values from the API. I just hadn't realized that enough to remove the "TODO" comments.

hobbitronics (approved):

PR #116, opened 2021-09-02
update a claim with receipts

hobbitronics:
# Added
- claimsFileAttach
# Changed
- call claimsFileAttach from [ClaimsID]
- reset files after uploading
- remove preview from the FileDropArea after uploading
- show receipts on a claim
uploadReceipt

### DIFF
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -76,7 +76,8 @@ const onSubmit = async () => {
   }
 
   for (let i = 0; i < files.length; i++) {
-    await claimsFileAttach(claimId, files[i].id)
+    const file = await upload(files[i])
+    await claimsFileAttach(claimId, file.id)
   }
 
   files = []
@@ -93,10 +94,10 @@ async function onUpload(event) {
 
     showPreview = true
 
-    const file = await upload(event.detail)
+    const file = event.detail
 
     files = [...files, file]
-
+    
   } finally {
     uploading = false
   }
@@ -161,7 +162,7 @@ async function onUpload(event) {
     <p>
       <Button on:click={editClaim} outlined>Edit claim</Button>
     </p>
-    {#if needsReceipt}
+    {#if needsReceipt || true}
       <Form on:submit={onSubmit}>
         <MoneyInput bind:value={repairOrReplacementCost} label={moneyFormLabel} />
 
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -36,7 +36,7 @@ $: needsReceipt = (needsRepairReceipt || needsReplaceReceipt)
 $: moneyFormLabel = needsRepairReceipt ? "Actual cost of repair" : "Actual cost of replacement"
 $: receiptType = needsRepairReceipt ? 'repair' : 'replacement'
 $: claimFiles = claim.claim_files || []
-$: setShowImgTrue(claimFiles.length)
+$: showImages(claimFiles.length)
 $: if(payoutOption === 'repair') {
     maximumPayout = computeRepairMaxPayout()
   } else if(payoutOption === 'replacement') {
@@ -57,7 +57,7 @@ const computeCashMaxPayout = () => computePayout(claimItem.coverage_amount, clai
 
 const editClaim = () => $goto(`claims/${claimId}/edit)`)
 
-const setShowImgTrue = length => {
+const showImages = length => {
   for (let i = 0; i < length; i++) {
     showImg[i] = true
   }
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -161,7 +161,7 @@ async function onUpload(event) {
     <p>
       <Button on:click={editClaim} outlined>Edit claim</Button>
     </p>
-    {#if needsReceipt || true}
+    {#if needsReceipt}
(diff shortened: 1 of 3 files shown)


src/pages/claims/[claimId].svelte
```diff
@@ -3,21 +3,26 @@ import user from '../../authn/user.js'
...
 let uploading = false
 let deductible = .05
 let maximumPayout = ''
+let files = []
+let showPreview = true
 
 $: $initialized || loadClaims()
 
-$: claim = $claims.find(clm => clm.id === $params.claimId) || {}
+$: claim = $claims.find(clm => clm.id === claimId) || {}
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:25]: At one point, I _think_ it was decided to use `$params` instead of `export`. I don't have a strong opinion on that, except that it might be good to be consistent.

src/pages/claims/[claimId].svelte
```diff
@@ -30,6 +35,8 @@ $: needsReplaceReceipt = (status === 'Needs_replace_receipt')
 $: needsReceipt = (needsRepairReceipt || needsReplaceReceipt)
 $: moneyFormLabel = needsRepairReceipt ? "Actual cost of repair" : "Actual cost of replacement"
 $: receiptType = needsRepairReceipt ? 'repair' : 'replacement'
+$: claimFiles = claim.claim_files || []
+$: setShowImgTrue(claimFiles.length)
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:39]: can this simply be `showImages` instead of `setShowImgTrue`?

src/pages/claims/[claimId].svelte
```diff
@@ -59,15 +74,29 @@ const onSubmit = () => {
...
-  
+
+  for (let i = 0; i < files.length; i++) {
+    await claimsFileAttach(claimId, files[i].id)
+  }
+
+  files = []
+  showPreview = false
+
+  await loadClaims() //TODO delte this once the updateClaim call is finished
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:85]: Is there something you're waiting on the backend code for the updateClaim call to be finished?

src/pages/claims/[claimId].svelte
```diff
@@ -59,15 +74,29 @@ const onSubmit = () => {
   } else if (needsReplaceReceipt) {
     updatedClaimData.replace_actual = cents
   }
-  
+
+  for (let i = 0; i < files.length; i++) {
```
Comment (Schparky) [inline src/pages/claims/[claimId].svelte:78]: If I'm reading this correctly, files aren't attached at the time they are uploaded, but only after clicking the submit button. If I understand correctly, some fields are saved on blur. This seems a tad inconsistent, but I could be misunderstanding.

Schparky (approved):

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:25]: @forevermatt switched most of the code to export. There might be one more place that needs to be changed.

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte]: Yes

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:78]: Correct. We should be giving a cancel option too(shown by an x in the preview). I was just going off of the mock-up which showed a submit button. TextFields are little easier to quickly edit the string. The claim form has a submit button but the user won't realize we are saving the fields, unlike in user settings. So I don't know, maybe we should be holding off on uploading till submitting?

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:25]: it looks like that was the last one.

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:78]: that is unless we want to upload in order to save their work.

src/pages/claims/[claimId].svelte
Comment (Schparky) [reply src/pages/claims/[claimId].svelte:85]: That's a new endpoint on ClaimItem. Is that what this code is waiting on?

src/pages/claims/[claimId].svelte
Comment (Schparky) [reply src/pages/claims/[claimId].svelte:78]: No problem. As long as the mockups show it this way, I have no problem with it.

Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:78]: Do you think it is better to upload and submit at the same time? I have the code in place to do it that way. Currently if the user uploads, then leaves the page we have no way of getting the uploads anyways.

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:85]: That sounds right. That function will be renamed updatedClaimData

src/pages/claims/[claimId].svelte
Comment (Schparky) [reply src/pages/claims/[claimId].svelte:78]: Maybe, but since we have UX experts on the team, I'm not wanting to go against their recommendation.

src/pages/claims/[claimId].svelte
```diff
@@ -161,7 +162,7 @@ async function onUpload(event) {
     <p>
       <Button on:click={editClaim} outlined>Edit claim</Button>
     </p>
-    {#if needsReceipt}
+    {#if needsReceipt || true}
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:165]: Did you mean to leave this `|| true` here?

Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:165]: woops, nope. I keep forgetting that

PR #119, opened 2021-09-08
animate FileDropArea previews

hobbitronics:
animatedUploader


Schparky (approved):
Can't say I understand this magic, but nothing looks out of place.

PR #121, opened 2021-09-09
Add accountable_person_id to addItem and updateItem

hobbitronics:
## accountable_person_id is now a required field on:
- addItem
- updateItem

### DIFF
--- a/src/data/items.js
+++ b/src/data/items.js
@@ -37,6 +37,7 @@ export async function addItem(policyId, itemData) {
   start(urlPath)
 
   const parsedItemData = {
+    accountable_person_id: itemData.accountablePersonId,
     category_id: itemData.categoryId,
     country: itemData.country,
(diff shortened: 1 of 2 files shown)


src/data/items.js
```diff
@@ -37,6 +37,7 @@ export async function addItem(policyId, itemData) {
   start(urlPath)
 
   const parsedItemData = {
+    accountable_person_id: itemData.accountablePersonId,
```
Schparky, line 40: Do you have a check somewhere that ensures that the ID is not `''`?

Schparky (approved):

src/data/items.js
hobbitronics: It looks like we don't check that the required form fields are full/valid yet. I think that will a separate PR. Thanks for pointing that out.

forevermatt (approved):

PR #122, opened 2021-09-09
Real data in home item table

hobbitronics:
## Fixed
- Replaced the non-exisitng properties in the home item table

Screen Shot 2021-09-09 at 3 39 18 PM

### DIFF
--- a/src/pages/home.svelte
+++ b/src/pages/home.svelte
@@ -35,13 +35,6 @@ onMount( () => {
   loadPolicies()
 })
 
-const getItemName = item => item.name || ''
-const getRecentActivity = item => item.coverage_status || ''
-const getAccountablePerson = item => item.accountable_person || ''
-const getCoverageAmount = item => formatMoney(item.coverage_amount)
-const getPremium = item => formatMoney(item.annual_premium)
-const getRiskCategoryName = item => item.risk_category?.name || ''
-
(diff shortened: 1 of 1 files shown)


src/pages/home.svelte
```diff
@@ -35,6 +35,13 @@ onMount( () => {
   loadPolicies()
 })
 
+const getItemName = item => item.name || ''
```
Schparky, line 38: If you're going to make these little getter functions, should they live somewhere where they can be exported (e.g. data/items.js)?

Schparky (approved):

src/pages/home.svelte
hobbitronics: The issue Ihad is the item in each row is in an #each loop from items, so I couldn't do a reactive statement. We would need a findItemById function as well. Do you think its worth it or should I just put it back in the markup?

Schparky: I guess I was thinking to just move these as-is but no big deal either way.

hobbitronics:
Oh, yeah that would work. I had my brain off for a second.

src/pages/home.svelte
hobbitronics: oh yeah, that works. I just had my brain switched off.

PR #123, opened 2021-09-10
Item details restyle and more data

hobbitronics:
Added real data that was either missing or faked.
Changed the layout and styling to match the mockup as shown below.
before:
Screenshot from 2021-09-10 15-29-34
after:
Screenshot from 2021-09-10 15-13-53

 @forevermatt I changed the Remove button as seen below.
Screenshot from 2021-09-10 17-32-49

### DIFF
--- a/src/pages/items/[itemId].svelte
+++ b/src/pages/items/[itemId].svelte
@@ -1,23 +1,42 @@
 <script>
 import user from '../../authn/user.js'
-import { Banner, Breadcrumb } from '../../components'
+import { Banner, Breadcrumb, ClaimBanner, Row } from '../../components'
+import { day } from '../../components/const.js'
+import { formatDate } from '../../components/dates.js'
 import { loading } from '../../components/progress'
 import { itemsByPolicyId, loadItems } from '../../data/items.js'
+import { init, policies } from '../../data/policies'
 import { formatMoney } from '../../helpers/money'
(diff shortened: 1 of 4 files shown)


Schparky (approved):

src/pages/items/[itemId].svelte
```diff
@@ -34,43 +53,58 @@ const goToDelete = () => {
...
+    {/if}
   {:else}
-    We could not find that item. Please <a href="/items">go back</a> and select
-    an item from the list.
+    <Row cols="12">
+      <div class="flex justify-between align-items-center" >
+        <Breadcrumb links={breadcrumbLinks} />
+        <div>
+          <!-- svelte-ignore a11y-invalid-attribute -->
+          <a on:click={goToDelete} class="delete-button mx-5px" href="">Remove</a>
```
Comment (forevermatt) [inline src/pages/items/[itemId].svelte:73]: If you want, one way to fix the accessibility warning is to just put the destination URL in the `href` itself, rather than listening for the `on:click` event. Just an option.

forevermatt (approved):

src/pages/items/[itemId].svelte
Comment (hobbitronics) [reply src/pages/items/[itemId].svelte]: See my added screenshot above

Comment (hobbitronics) [reply src/pages/items/[itemId].svelte]: Alex says we will rely on a dialogue instead of red so disregard the red button.

PR #126, opened 2021-09-10
Start a claim, plus misc. bug fixes

forevermatt:
### Changed
- **Refactor nearly the entire `ClaimForm` component, for correctness and readability**
- For API calls that are bad requests (400), show the error message
- Remove unneeded `getClaim` function, since we always get the full list

### Added
- Add new claim-payout-amounts file for those business rules (as they affect ClaimForm logic)
- Add helper method for converting dollars to cents, handling `null`/`undefined` appropriately

### Fixed
- Avoid showing "Item does not exist" on new-claim page while loading data
- Keep the `ClaimForm` present across API calls, to avoid losing values
- Fix the call to scroll the error message into view
- Fix placement/spacing of various text labels to match mockup
- Fix display of covered value on claim-details page


forevermatt:
There are a lot of changes in this, so please feel free to discuss, question, challenge, and/or suggest improvements.

I tried to get away from variable names like "payoutOptionCheck" and "seventyPercentCheck", preferring names that better reflect what the actual question is that those variables hold the answer to (e.g. "repairCostIsTooHigh").

In addition, I strategically used `true`, `false`, and `undefined`/`null`. It can make it a little unclear, but I didn't find a better way to handle the various situations like "We know this is too expensive to repair" vs. "We this isn't too expensive to repair" vs. "We don't yet know whether it's too expensive to repair." Feel free to comment / suggest improvements.

I also pulled some logic out into a javascript file in a new "business-rules" folder. What do you think of this approach?

src/components/Error.svelte
```diff
@@ -29,4 +36,4 @@ span small {
   <span bind:this={errContainer}>
```
hobbitronics, line 36: I've been meaning to make a card or TODO to replace this span with Snackbar. Just putting it out there while I'm thinking of it.

hobbitronics (approved):
Kind of big chunk to digest on a Monday morning, so sorry that I don't have more feedback, but looks good.

RussellChamp:
Yes, this looks like code to me as well.

PR #127, opened 2021-09-13
typescript implementation

RussellChamp:
* Rename all the js files to ts files
* Update all the svelte components to add `lang="ts"` to the script line
* Added necessary rollup build options
* Tweaked a few components to make them happy


tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
hobbitronics, line 1: I guess I will have to read the docs to understand this part, but just was curioiu why "__sapper__/*" is here.

hobbitronics (approved):
Looks pretty straightforward so far.

rollup.config.js
```diff
@@ -26,6 +28,9 @@ export default {
 			emitCss: true, // give component style to postcss() for processing
 			preprocess: autoPreprocess(),
 		}),
+		
+		typescript({ sourceMap: !production }),
+		nodePolyfills(),
```
forevermatt, line 33: What were the node polyfills needed for?

src/analytics/index.ts
```diff
@@ -7,10 +7,12 @@ init()
 // https://developers.google.com/analytics/devguides/collection/gtagjs
 // https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs
 function init() {
-  loadLib()
+  loadLib();
+
+  // TODO: get the dataLayer references to play nicer with ts
```
forevermatt, line 12: If it simplifies things, feel free to leave selected files (like this one) as `.js` instead of `.ts`. It's okay if we leave some plain JavaScript files in place, especially for files we won't be frequently changing or importing into other files.

src/data/index.ts
```diff
@@ -3,16 +3,18 @@ import { start, stop } from '../components/progress'
...
 
-export async function CREATE(uri, body) { return await customFetch('post'  , uri, body) }
-export async function GET   (uri      ) { return await customFetch('get'   , uri      ) }
-export async function UPDATE(uri, body) { return await customFetch('put'   , uri, body) }
-export async function DELETE(uri      ) { return await customFetch('delete', uri      ) }
+type FetchMethod = 'post' | 'get' | 'put' | 'delete';
+
+export async function CREATE(uri: string, body: any = undefined) { return await customFetch('post'  , uri, body) }
+export async function GET   (uri: string      ) { return await customFetch('get'   , uri      ) }
+export async function UPDATE(uri: string, body) { return await customFetch('put'   , uri, body) }
```
forevermatt, line 10: Did you mean to also change this `body` to `body: any`? I wasn't sure if this `UPDATE()` or the above `CREATE()` was more in line with the desired outcome here.

src/data/index.ts
```diff
@@ -3,16 +3,18 @@ import { start, stop } from '../components/progress'
 import { throwError } from '../error'
 import t from '../i18n'
 
-export async function CREATE(uri, body) { return await customFetch('post'  , uri, body) }
-export async function GET   (uri      ) { return await customFetch('get'   , uri      ) }
-export async function UPDATE(uri, body) { return await customFetch('put'   , uri, body) }
-export async function DELETE(uri      ) { return await customFetch('delete', uri      ) }
+type FetchMethod = 'post' | 'get' | 'put' | 'delete';
```
forevermatt, line 6: Ahhh.... so _that's_ how you do custom types for that sort of thing. TIL!

ts.errors
```diff
@@ -0,0 +1,1592 @@
+
```
forevermatt, line 1: Looks like you'll want to delete this... and possibly add it to `.gitignore`? I'm not sure whether it was autogenerated or was just concat'd to a file for debugging purposes.

forevermatt (changes requested):
Mostly looks good, just a few questions and one file to delete.

tsconfig.json
RussellChamp: ah, i think that was just a copy pasta that would be more applicable to other svelte apps

rollup.config.js
RussellChamp: i think that was due to the calls to `process` in analytics

src/analytics/index.ts
RussellChamp: either way is probably fine. i was on a roll here, though.

src/data/index.ts
RussellChamp: elsewhere in the code, `CREATE` was being called with only 1 parameter and missing the `body` parameter and caused ts to go "hey that doesn't seem right". I just put a bandaid on things for now. we can come back and do this "right" later.

src/data/index.ts
RussellChamp: yup! that's right!

ts.errors
RussellChamp: doh! yeah, i just wrote out some of the error lines to a temp file for reference. i'll squash that

PR #128, opened 2021-09-14
update components and methods to use strong types

RussellChamp:
fix a few bugs along the way
leave a few TODOs


src/data/claims.ts
```diff
@@ -2,24 +2,135 @@ import { CREATE, GET, UPDATE } from "."
...
 import { convertToCents } from "../helpers/money"
 import { writable } from "svelte/store"
+import type { PolicyItem } from "./items";
 
-export const claims = writable([])
-export const initialized = writable(false)
-export const warning = {
+export type PayoutOption = 'Repair' | 'Replacement' | 'FMV';
+export type ClaimItemStatus = 'Pending' | 'Approved' | 'Denied';
+export type ClaimEventType = 'Theft' | 'Impact' | 'Electrical Surge' | 'Water Damage' | 'Evacuation' | 'Other';
```
Schparky, line 9: These types are defined on the back end and exposed by a REST call: `GET /config/claim-event-types`.

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -2,12 +2,12 @@
 import user from '../../../authn/user'
 import { Breadcrumb, ClaimForm } from '../../../components'
 import { loading } from '../../../components/progress'
-import { claims, initialized, loadClaims, updateClaim } from '../../../data/claims.js'
-import { itemsByPolicyId, loadItems } from '../../../data/items.js'
+import { Claim, ClaimItem, claims, initialized, loadClaims, updateClaim } from '../../../data/claims.js'
```
Comment (forevermatt) [inline src/pages/claims/[claimId]/edit.svelte:5]: Should this be `claims.ts` (or `claims` with no extension)?

Comment (forevermatt) [reply src/pages/claims/[claimId]/edit.svelte:5]: Looks like there may be a couple other imports that still specify `.js` extensions on files that are now `.ts`, too.

Schparky (approved):

src/components/ClaimCardBanner.svelte
```diff
@@ -1,6 +1,9 @@
 <script lang="ts">
-export let item = {}
-export let state = {}
+import type { State } from "../data/claims";
+
+
+export let item = {} as any
```
forevermatt, line 5: It may not matter, but I think this is a `PolicyItem` now

src/data/claims.ts
```diff
@@ -2,24 +2,135 @@ import { CREATE, GET, UPDATE } from "."
 import { start, stop } from "../components/progress"
 import { convertToCents } from "../helpers/money"
 import { writable } from "svelte/store"
+import type { PolicyItem } from "./items";
 
-export const claims = writable([])
-export const initialized = writable(false)
-export const warning = {
+export type PayoutOption = 'Repair' | 'Replacement' | 'FMV';
```
forevermatt, line 7:
```suggestion
export type PayoutOption = 'Repair' | 'Replacement' | 'FMV' | 'FixedFraction';
```

forevermatt: This change is coming, and is already reflected in code in this repo. The API is almost there, but probably wouldn't have had that in the docs yet.

src/data/claims.ts
forevermatt: The type defined as `ClaimEventType` in src/data/claim-event-types.ts is actually correct. This type here is (as @Schparky alluded to) probably one it's better to remove, since those are dynamic values obtained from the API.

src/components/ClaimCardBanner.svelte
```diff
@@ -1,6 +1,9 @@
 <script lang="ts">
-export let item = {}
-export let state = {}
+import type { State } from "../data/claims";
```
hobbitronics, line 2: we are leaving ; out.

src/data/claims.ts
```diff
@@ -105,25 +216,29 @@ export async function createClaim(item, claimData) {
...
   start(urlPath)
   try {
-    const parsedClaimItem = {
+    const parsedClaimItem: CreateClaimItemRequestBody = {
       fmv: convertToCents(claimItemData.fairMarketValueUSD),
       is_repairable: claimItemData.isRepairable,
       item_id: claimItemData.itemId,
       payout_option: claimItemData.payoutOption,
       repair_estimate: convertToCents(claimItemData.repairEstimateUSD),
+      repair_actual: undefined, //TODO: get this value
```
forevermatt, line 229: FWIW, we actually won't be sending the `..._actual` values during a create, though we're still in-process on that change.

src/authn/user.ts
```diff
@@ -1,15 +1,37 @@
...
+    created_by_id: string;
+    id: string;
+    name: string;
+    size: number;
+    url: string;
+    url_expiration: string /*Date*/;
+  },
+  photo_file_id: string;
+  policy_id: string;
+};
```
hobbitronics, line 24: no ; for consistency

src/components/filteredList/FilteredList.svelte
```diff
@@ -1,19 +1,20 @@
 <script lang="ts">
 import { throwError } from '../../error/index.js'
 import { writable } from 'svelte/store';
+import type Filter from './filter';
```
hobbitronics, line 4: ;'s

src/components/forms/ClaimForm.svelte
```diff
@@ -20,9 +20,11 @@ import {
 import { claimEventTypes, loadClaimEventTypes } from '../../data/claim-event-types'
 import { Button, Form, TextArea } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
+import type { Claim, ClaimItem } from '../../data/claims';
+import type { PolicyItem } from '../../data/items';
```
hobbitronics, line 24: :'s

src/data/claims.ts
```diff
@@ -105,25 +216,29 @@ export async function createClaim(item, claimData) {
...
     }
   
-    const claimItem = await CREATE(urlPath, parsedClaimItem)
+    const claimItem = await CREATE<ClaimItem>(urlPath, parsedClaimItem)
   
     claims.update(claims => {
-      const claim = claims.find(c => c.id === claimId) || {}
-      const claimItems = claim.claim_items || []
-      claimItems.push(claimItem)
+      const claim = claims.find(c => c.id === claimId)
```
forevermatt, line 237: It seems helpful to be able to write code with fewer `if`s... is there a way we can continue doing that (such as in the prior version of this `claims.update()` call) with TypeScript?

src/components/ClaimCard.svelte
```diff
@@ -1,13 +1,14 @@
 <script lang="ts">
 import ClaimCardBanner from './ClaimCardBanner.svelte'
 import { day } from './const'
-import { getState } from '../data/claims'
+import { getState, Claim, ClaimItem } from '../data/claims'
 import { Card, Button } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
+import type { PolicyItem } from '../data/items';
```
hobbitronics, line 7: ordering and ;

src/components/forms/ItemForm.svelte
```diff
@@ -5,9 +5,10 @@ import { categories, init, initialized as catItemsInitialized } from '../../data
 import { loadMembersOfPolicy, membersByPolicyId } from '../../data/policy-members'
 import { Button, Form, Select, TextArea, TextField } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
+import type { PolicyItem } from '../../data/items';
```
hobbitronics, line 8: ; and import order

src/data/claims.ts
```diff
@@ -2,24 +2,135 @@ import { CREATE, GET, UPDATE } from "."
 import { start, stop } from "../components/progress"
 import { convertToCents } from "../helpers/money"
 import { writable } from "svelte/store"
+import type { PolicyItem } from "./items";
```
hobbitronics, line 5: same comment as above

src/data/claims.ts
```diff
@@ -2,24 +2,135 @@ import { CREATE, GET, UPDATE } from "."
...
 
-export const claims = writable([])
-export const initialized = writable(false)
-export const warning = {
+export type PayoutOption = 'Repair' | 'Replacement' | 'FMV';
+export type ClaimItemStatus = 'Pending' | 'Approved' | 'Denied';
+export type ClaimEventType = 'Theft' | 'Impact' | 'Electrical Surge' | 'Water Damage' | 'Evacuation' | 'Other';
+export type ClaimStatus = 'Draft' | 'Pending' | 'Approved' | 'Denied';
```
hobbitronics, line 10: ;

src/data/claims.ts
```diff
@@ -2,24 +2,135 @@ import { CREATE, GET, UPDATE } from "."
...
+    name: string;
+    size: number;
+    url: string;
+    url_expiration: string /*Date*/;
+  };
+  file_id: string;
+  id: string;
+  purpose: ClaimFilePurpose;
+  updated_at: string /*Date*/;
+};
```
hobbitronics, line 29: ;

src/data/claims.ts
forevermatt: Hmm... it looks like in this case you were actually fixing a bug, where we could have been adding a ClaimItem to an array that's simply thrown away if there was no matching claim found, so my question may be less applicable here than I thought, though it still (to some degree) is relevant.

src/data/claims.ts
```diff
@@ -194,10 +313,10 @@ export async function claimsFileAttach(claimId, fileId) {
  * @export
  * @param {Number} itemId 
  */
-export function deleteClaim(itemId) {
+export function deleteClaim(itemId: string) {
   start(itemId)
 
-  claims.update(currClaims => currClaims.filter(clm => clm.itemId !== itemId))
+  claims.update(currClaims => currClaims.filter(clm => clm.id !== itemId))
```
forevermatt, line 319: We may just need to stick a todo here... this method itself seems suspect in my mind. I don't think we would ever delete a claim by specifying an Item ID.

Actually, @RussellChamp, would you mind simply renaming the `itemId` parameter to `claimId`? I think that's probably what it was supposed to be. Your current fix here, plus that change, should get this function to being (theoretically) correct.

src/data/items.ts
```diff
@@ -2,20 +2,105 @@ import { CREATE, DELETE, GET, UPDATE } from "."
...
+
+export type ItemCoverageStatus = 'Draft' | 'Pending' | 'Approved' | 'Denied';
+export type PolicyType = 'Household' | 'Corporate';
+
+export type RiskCategory = {
+  created_at: string /*Date*/;
+  id: {
+    UUID: any;
+    Valid: boolean;
+  };
```
forevermatt, line 15: This seems odd to me... let's verify with the API team that a RiskCategory's `id` is an array, rather than just a number or string.

src/data/claims.ts
```diff
@@ -2,24 +2,135 @@ import { CREATE, GET, UPDATE } from "."
...
+  repair_actual: number;
+  replace_actual: number;
+}
+
+export type State = {
+  icon: string;
+  color: string;
+  bgColor: string;
+  title: string;
+  actionLabel: string;
```
hobbitronics, line 114: I am doing away with actionLabel. Since my PR will likely merge first you can get rid of it in the type.

src/data/items.ts
forevermatt: Looks like that was a mistake in the API. This can be changed to
```suggestion
  id: string;
```

src/data/claims.ts
```diff
@@ -105,25 +216,29 @@ export async function createClaim(item, claimData) {
...
-    const parsedClaimItem = {
+    const parsedClaimItem: CreateClaimItemRequestBody = {
       fmv: convertToCents(claimItemData.fairMarketValueUSD),
       is_repairable: claimItemData.isRepairable,
       item_id: claimItemData.itemId,
       payout_option: claimItemData.payoutOption,
       repair_estimate: convertToCents(claimItemData.repairEstimateUSD),
+      repair_actual: undefined, //TODO: get this value
       replace_estimate: convertToCents(claimItemData.replaceEstimateUSD),
+      replace_actual: undefined //TODO: get this value
```
hobbitronics, line 231: "replace_actual" and "repair_actual" are not needed for this endpoint. I think they were accidentally included.

src/data/policies.ts
```diff
@@ -1,9 +1,32 @@
 import { writable } from 'svelte/store'
 import { start, stop } from '../components/progress'
+import type { Claim } from './claims'
 import { GET, UPDATE } from './index'
+import type { PolicyMember } from './policy-members'
 
-export const policies = writable([])
-export const initialized = writable(false)
+export type Policy = {
```
forevermatt, line 7: No single policy will have all of these fields, so we may later want to split this into two different subtypes, just as a heads-up.

src/data/policy-members.ts
```diff
@@ -2,20 +2,30 @@ import { GET } from './index'
...
-export const membersByPolicyId = writable({})
+export type PolicyMember = {
+  email: string;
+  email_override: string;
+  first_name: string;
+  id: string;
+  last_login_utc: string /*Date*/;
+  last_name: string;
+  location: string;
+};
```
hobbitronics, line 13: ;

hobbitronics (approved):
If this will get rid of the red squiggles I will be happy.

src/components/filteredList/FilteredList.svelte
forevermatt: Just to fill in a few more details, we try to alphabetize our imports by where they're imported from (ignoring `..` path segments) and putting our in-app imports before npm-dependency imports.

forevermatt (commented):
Okay, finally finished reviewing. Added a few comments and suggestions, but otherwise I think this will indeed help us spot bugs sooner as we code (and learn what fields are available as we code).

forevermatt (approved):

PR #131, opened 2021-09-15
created types for ui-components

RussellChamp:
* the correct way to do this would be to add types to the library
* this works as a bandaid for now

### DIFF
--- a/src/customTypes/@silintl/ui-components.d.ts
+++ b/src/customTypes/@silintl/ui-components.d.ts
@@ -1,0 +1,249 @@
+
+declare module '@silintl/ui-components' {
+  import type { SvelteComponentTyped } from 'svelte';
+  import type { writable } from 'svelte/store';
+
+  export type actions = writable<any[]>;
+  
+  interface ButtonProps extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['div']> {
+    disabled?: any;
+    outlined?: any;
+    raised?: any;
+    prependIcon?: any;
+    appendIcon?: any;
+    url?: any;
+  };
+  export class Button extends SvelteComponentTyped<ButtonProps> {};
+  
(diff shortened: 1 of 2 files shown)


src/customTypes/@silintl/ui-components.d.ts
```diff
@@ -0,0 +1,249 @@
+
+declare module '@silintl/ui-components' {
+  import type { SvelteComponentTyped } from 'svelte';
+  import type { writable } from 'svelte/store';
+
+  export type actions = writable<any[]>;
+  
+  interface ButtonProps extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['div']> {
```
forevermatt, line 8: What is JSX?

src/customTypes/@silintl/ui-components.d.ts
```diff
@@ -0,0 +1,249 @@
...
+    uppercase?: boolean;
+  };
+	export class Checkbox extends SvelteComponentTyped<CheckboxProps> {};
+  
+  interface DatatableProps extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['div']> {
+    label?: string;
+  };
+	export class Datatable extends SvelteComponentTyped<DatatableProps> {};
+
+  export namespace Datatable {
```
forevermatt, line 44: Ahhh... this looks like it may fix the errors my IDE is showing about Datatable.Header (etc.) not existing. 😁

src/customTypes/@silintl/ui-components.d.ts
RussellChamp: JSX is the thing that allows you to define html elements in javascript. eg https://reactjs.org/docs/introducing-jsx.html

forevermatt (approved):

PR #132, opened 2021-09-15
Draft uploads and file purpose

hobbitronics:
### Added
- filePurpose
- getFilePurpose()
- needsEvidence
- needsFile
- uploadLabel

### Changed
- control block around FileDropArea and MoneyInput in [claimID]
- animation delay on FilePreview
- business-rules to use ts

### Fixed
- bug: wrong payout option 'replace'

Realized we need the second "needs changes" banner since we are requiring files.
Screenshot from 2021-09-15 15-53-33

### DIFF
--- a/src/components/FilePreview/FilePreview.svelte
+++ b/src/components/FilePreview/FilePreview.svelte
@@ -50,7 +50,8 @@ function onDelete(event, id) {
 
 <div class="mt-10px py-10px {$$props.class}">
   {#each previews as preview (preview.id)}
-    <div on:click|preventDefault={() => onClick(preview.id)} transition:fly={{ y: 200, duration: 1500 }} animate:flip={{duration: 500}} class:selected={isSelected[preview.id]} class="preview flex justify-between align-items-center br-8px p-10px mb-1">
+    <!-- TODO: fix - transition has to finish before leaving page -->
+    <div on:click|preventDefault={() => onClick(preview.id)} transition:fly={{ y: 200, duration: 500 }} animate:flip={{duration: 500}} class:selected={isSelected[preview.id]} class="preview flex justify-between align-items-center br-8px p-10px mb-1">
       <div>
         <p class="white my-0">{preview.file.name}</p>
         <p class="white my-0">{formatDate(preview.created_at)}</p>
--- a/src/components/FilePreview/FilePreview.svelte
+++ b/src/components/FilePreview/FilePreview.svelte
@@ -3,7 +3,6 @@ import { formatDate } from '../dates'
 import { Button, Progress } from '@silintl/ui-components'
 import { createEventDispatcher } from "svelte"
 import { flip } from 'svelte/animate'
-import { fly } from 'svelte/transition'
 
 export let previews = []
 export let uploading = false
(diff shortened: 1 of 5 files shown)


src/components/FilePreview/FilePreview.svelte
```diff
@@ -48,9 +48,10 @@ function onDelete(event, id) {
 }
 </style>
 
+<!-- TODO: fix - transition has to finish before leaving page -->
```
forevermatt, line 51: For simplicity, we could also simply disable any transitions that cause issues for now. Those could definitely be a post-1.0 addition.

src/data/claims.ts
```diff
@@ -265,12 +265,12 @@ export async function updateClaim(claimId: string, newClaimData) {
...
 }
 
-export async function claimsFileAttach(claimId: string, fileId: string) {
+export async function claimsFileAttach(claimId: string, fileId: string, purpose: ClaimFilePurpose) {
   start(fileId)
 
   const data: ClaimsFileAttachRequestBody = {
     file_id: fileId,
-    purpose: undefined // TODO: get this value
+    purpose: undefined
```
forevermatt, line 273: Should this have been 
```suggestion
    purpose,
```
?

src/pages/claims/[claimId].svelte
```diff
@@ -54,6 +58,21 @@ const computeReplaceMaxPayout = () => computePayout(claimItem.replace_estimate,
 
 const computeCashMaxPayout = () => computePayout(item.coverage_amount, claimItem.fmv)
 
+const getFilePurpose = (claimItem: ClaimItem, needsReceipt: Boolean): ClaimFilePurpose => {
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:61]: If there's a way to keep this function from having side effects (in this case, reaching out and changing the value of `uploadLabel`), it will help debugging later. Better to have a 2nd function along the lines of `getUploadLabel()` that has nearly-duplicate logic than to have the `getFilePurpose()` function also invisibly set `uploadLabel`.

src/data/claims.ts
hobbitronics: why yes, thought I had done that.

src/components/FilePreview/FilePreview.svelte
hobbitronics: yup, we can leave that out for now.

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:61]: yeah, no problem

src/pages/claims/[claimId].svelte
```diff
@@ -127,9 +143,9 @@ function onDeleted(event) {
   </Row>
   <Row cols="9">
     <ClaimBanner claimStatus={status} />
-    {#if needsReceipt}
+    {#if needsFile}
       <ClaimBanner claimStatus={`${status}2`} >
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:147]: Is that `2` hard-coded there intentionally?

Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:147]: yeah, there is a Draft2 and NeedsChanges2 state

Comment (forevermatt) [reply src/pages/claims/[claimId].svelte:147]: But don't we want to show the claim's current status, rather than the "something2" version of the status? I'm probably just missing something here...

Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:147]: We show the second banner only when a file is required. The first status is still shown. Did you see the screenshot I added?

forevermatt (approved):

PR #134, opened 2021-09-15
Policy details page

forevermatt:
### Added
- Add formatFriendlyDate() helper
- Add basic policy-details page, with links to claim details

### Fixed
- Prevent error from trying to get properties of an item that wasn't found

### DIFF
--- a/src/helpers/date.ts
+++ b/src/helpers/date.ts
@@ -1,0 +1,5 @@
+
+export const formatFriendlyDate = (dateTimeString: string) => {
+  const date = new Date(dateTimeString)
+  return date.toLocaleDateString()
+}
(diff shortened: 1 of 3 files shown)


RussellChamp (approved):
lgtm

src/helpers/date.ts
```diff
@@ -0,0 +1,5 @@
+
+export const formatFriendlyDate = (dateTimeString: string) => {
+  const date = new Date(dateTimeString)
+  return date.toLocaleDateString()
+}
```
hobbitronics, line 5: we have a /components/dates.js file that does basically this with a few parameters. Could you see if that meets the requirement and/or move both functions to the same file.

hobbitronics (commented):
Looks good.

src/helpers/date.ts
forevermatt: Ahh, thanks! I thought we had one, but I wasn't seeing it. I'll take a look at combining those.

PR #136, opened 2021-09-16
Link to policies page, minor touch-ups

forevermatt:
### Added
- Add menu item for Policies page (for Stewards)
- Added `UserAppRole` type
- Add fun little icons by the policy type

### Changed
- Change to better icons for policies vs. FAQ (in menu)

### Fixed
- Switch to more normal "home" icon in menu


src/pages/policies.svelte
```diff
@@ -20,15 +27,22 @@ $: $initialized || loadPolicies()
     <Datatable.Data>
       {#each $policies as policy (policy.id) }
         <Datatable.Data.Row on:click={() => $goto(`/policies/${policy.id}`)} clickable>
-          <Datatable.Data.Row.Item>{policy.type || ''}</Datatable.Data.Row.Item>
+          <Datatable.Data.Row.Item>
+            {#if policy.type === 'Household' }
+              <span class="material-icons">family_restroom</span>
```
hobbitronics, line 32: interesting icon name

src/pages/policies/[policyId].svelte
```diff
@@ -103,6 +103,12 @@ th {
             <Datatable.Data.Row.Item>{formatMoney(claimItem.fmv)}</Datatable.Data.Row.Item>
           </Datatable.Data.Row>
         {/each}
+      {:else}
```
Comment (hobbitronics) [inline src/pages/policies/[policyId].svelte:106]: I am having trouble seeing where the {#if} is.

src/pages/policies.svelte
forevermatt: I agree, but it seemed the most representative of a family/household:
https://fonts.google.com/icons?selected=Material+Icons&icon.query=family

src/pages/policies/[policyId].svelte
Comment (forevermatt) [reply src/pages/policies/[policyId].svelte:106]: It's actually to the `{$each}`. Svelte uses `{:else}` to let you show different content if the list you're iterating over is empty. "Show this for each of these, but if there aren't any then show this other thing."

Comment (hobbitronics) [reply src/pages/policies/[policyId].svelte:106]: well that's nifty.

hobbitronics (approved):
Looks good.

PR #137, opened 2021-09-16
Working through claim states

hobbitronics:
Changed
- States - renamed to match api claim.status and claim state chart
- ClaimStatus type
- when 'Edit Claim' button is visible
- 'Submit button' visible for 'Receipt' status which set the status to Review2

fixed
- bug when udpating claimItem
- bug where receipType was undefined
- ClaimCardBanner and ClaimCard banner were not using claim.status_reason

removed
- 'Inactive' from State
- item from ClaimCardBanner since item.message doesn't exist


src/pages/claims/[claimId].svelte
```diff
@@ -29,14 +29,15 @@ $: $user.policy_id && loadItems($user.policy_id)
 $: item = items.find(itm => itm.id === claimItem.item_id) || {} as PolicyItem
 $: incidentDate = formatDate(claim.incident_date)
 $: status = claim.status || ''
-$: payoutOption = claimItem.payout_option
-$: needsRepairReceipt = (status === 'Needs_repair_receipt')
-$: needsReplaceReceipt = (status === 'Needs_replace_receipt')
-$: needsReceipt = (needsRepairReceipt || needsReplaceReceipt)
+$: payoutOption = claimItem.payout_option as PayoutOption
+$: isEditable = (status !== 'Approved') && (status !== 'Denied') && (status !== 'Paid')
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:33]: This might be a good opportunity to extract some of this logic into another "business rules" file. Perhaps an `isEditableStatus(status)` function?

forevermatt (approved):
Approved, with one suggestion for a future PR

src/pages/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId].svelte:33]: I like this idea, but I think I will save it for a separate PR so I can see if there is anything else I'd like to include.

PR #138, opened 2021-09-17
fixed editClaim url

hobbitronics:
There were some typos.

I thought about getting rid of the editClaim function and just giving the button the relative url. Any benefits to using $goto?

### DIFF
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -71,7 +71,7 @@ const getUploadLabel = (claimItem: ClaimItem, needsReceipt: Boolean, receiptType
   if(claimItem.fmv) return 'evidence of fair market value'
 }
 
-const editClaim = () => $goto(`claims/${claimId}/edit)`)
+const editClaim = () => $goto(`/claims/${claimId}/edit`)
 
 const onSubmit = async () => await submitClaim(claimId)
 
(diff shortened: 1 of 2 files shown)


RussellChamp:
it appears that `$goto` is wired into "svelte/store" while the Button url just created an anchor tag link. there _could_ be benefits in using $goto regarding full page reloads, etc but i'm not yet savvy enough with svelte to know that degree of inner workings.
code lgtm

RussellChamp (approved):

hobbitronics:
> it appears that `$goto` is wired into "svelte/store" while the Button url just created an anchor tag link. there _could_ be benefits in using $goto regarding full page reloads, etc but i'm not yet savvy enough with svelte to know that degree of inner workings.
> code lgtm

Yeah, routify is a plugin which uses svelte/store and has some benefits like you mentioned, but in this case I am unsure. I think Matt was going to do some component wrapping of these buttons so I'll let him decide.

forevermatt:
> > it appears that `$goto` is wired into "svelte/store" while the Button url just created an anchor tag link. there _could_ be benefits in using $goto regarding full page reloads, etc but i'm not yet savvy enough with svelte to know that degree of inner workings.
> > code lgtm
> 
> Yeah, routify is a plugin which uses svelte/store and has some benefits like you mentioned, but in this case I am unsure. I think Matt was going to do some component wrapping of these buttons so I'll let him decide.

I think we can use simple `<a ...` links as freely as using `$goto()` in Svelte. That's one of the nice things about it over, say, Vue: in Svelte, clicking a link to somewhere within the SPA just goes to that route instead of doing a full page reload; in Vue, you have to call a method on the `router` object or use a special component to avoid full page reloads (in my experience).

PR #139, opened 2021-09-17
user preferences

RussellChamp:
* Add setting nav button
* Add user setting page
* Move and update household/policy setting page
* Added croppie.js for avatar cropping purposes


RussellChamp:
There are several "TODO"s in this work as there is not yet an endpoint to update user data

src/components/FileDropArea/FileDropArea.svelte
```diff
@@ -6,11 +6,11 @@ export let outlined = false
 export let uploading = false
 export let mimeType = 'application/pdf,image/*'
 
-let fileInput = {}
+let fileInput = {} as any
 
 let highlighted = false
 
-const dispatch = createEventDispatcher()
+const dispatch = createEventDispatcher<{upload: FormData}>()
```
hobbitronics, line 13: I am a little confused by why this function which takes no parameter needs this. Maybe I am confused by what <> are typing, what the functions returns? Since this is a built in Svelte function which we won't mess with do we need to type it?

src/pages/settings.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
```
hobbitronics, line 1: can you check if you are using 2 spaces for indentation?

src/pages/settings/household/dependent.svelte
```diff
@@ -6,12 +6,12 @@ import { goto } from '@roxi/routify'
 import { Page } from '@silintl/ui-components'
 
 const onCancel = () => {
-  $goto('../settings')
+  $goto('../household')
```
forevermatt, line 9: This comment is only semi-related to this PR, but more about spreading tribal knowledge.

We've found it helpful to use root-relative paths in calls to `$goto()`, so as we're adding new calls or editing existing calls, it would be helpful to make them full paths. In this example, that would be 
```suggestion
  $goto('/settings/household')
```

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+              let data = new FormData()
+              data.append('file', result)
+
+              const file = await upload(data)
+
+              updateUser({
+                ...$user,
+                photo_file: {
+                  content_type: file.content_type,
+                  created_by_id: $user.id,
```
forevermatt, line 97: We probably should be able to specify a `created_by_id` field. If the API accepts that as an input, let's create a card about that as a bug.

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+                  id: file.id,
```
forevermatt, line 98: I suspect we also won't be able to specify the ID. That's an API-generated UUID, typically.

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+                  name: file.filename,
+                  size: file.size,
+                  url: file.url,
+                  url_expiration: undefined
+                },
+                photo_file_id: file.id
```
forevermatt, line 104: Same here: we can't (or at least shouldn't be able to) specify the ID for new content.

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+              })
+
+              setNotice('Your profile photo has been updated')
+            })
+          } catch {
+            setNotice('There was an error uploading your profile photo')
+          } finally {
+              uploading = false
```
forevermatt, line 112: Minor issue, but it looks like indentation got off by one level from just inside the `try` down to here.

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+
+  <h3 class="ml-1 mt-3">Location<span class="required">*</span></h3>
+  <p>
+    <TextField placeholder={'Enter country'} bind:value={location} on:blur={updateLocation} />
+  </p>
+
+  <h3 class="ml-1 mt-3">Receive notification emails for</h3>
+  <p>
+      {#each $policies as policy (policy.id) }
+      <Checkbox label={policy.type} checked={true} on:checked={() => handleChecked(policy.id)} on:unchecked={() => handleUnchecked(policy.id)} ></Checkbox>
```
forevermatt, line 152: Another simplification Svelte allows is that you can leave off the value of an attribute if you simply want it to be `true`, which can make things like `checked` shorter yet still clearly understandable.
```suggestion
      <Checkbox label={policy.type} checked on:checked={() => handleChecked(policy.id)} on:unchecked={() => handleUnchecked(policy.id)} ></Checkbox>
```

src/components/FileDropArea/FileDropArea.svelte
forevermatt: This is the only blocking issue I see on this PR. All my other comments are just minor cleanup.

As @hobbitronics said, this typing of the return value of `createEventDispatcher()` seems out of place.

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
+<script lang="ts">
+    import user, { updateUser } from '../../authn/user'
+    import { Breadcrumb, FileDropArea, RadioOptions } from "../../components"
+    import { upload } from '../../data'
+    import { policies, init as loadPolicies } from '../../data/policies'
+    import { Button, Checkbox, TextField, Page, Snackbar, setNotice } from "@silintl/ui-components"
```
hobbitronics, line 6: not sure if our own library gets sorted with other dependencies. @forevermatt ?

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+    import Croppie from "croppie"
+    import "croppie/croppie.css"
+    
+    const NOTIFICATION_OPTION_DEFAULT = "default_email"
+    const NOTIFICATION_OPTION_CUSTOM = "custom_email"
+
+    let uploading = false
+    let notification_email = $user.email_override ? NOTIFICATION_OPTION_CUSTOM : NOTIFICATION_OPTION_DEFAULT
+    let email_override = $user.email_override || ''
```
hobbitronics, line 15: would the above two lines work as a reactive statement? That way if $user hasn't finished loading when they are declared they will still get updated.

src/pages/settings/personal.svelte
forevermatt: If I understood you correctly, yes: our ui-components library is just another library this code uses, so it gets sorted with other libraries, after our internal imports. However, it looks like it's already correctly placed in that record.

I'm not sure what to do with the CSS import below, but otherwise these imports look (at a brief review) about right to me.

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+    let location = $user.location || ''
```
hobbitronics, line 16: same as comment above

hobbitronics:
```suggestion
    $: location = $user.location || ''
```

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+    import { policies, init as loadPolicies } from '../../data/policies'
+    import { Button, Checkbox, TextField, Page, Snackbar, setNotice } from "@silintl/ui-components"
+    import Croppie from "croppie"
+    import "croppie/croppie.css"
+    
+    const NOTIFICATION_OPTION_DEFAULT = "default_email"
+    const NOTIFICATION_OPTION_CUSTOM = "custom_email"
+
+    let uploading = false
+    let notification_email = $user.email_override ? NOTIFICATION_OPTION_CUSTOM : NOTIFICATION_OPTION_DEFAULT
```
hobbitronics, line 14:
```suggestion
    $: notification_email = $user.email_override ? NOTIFICATION_OPTION_CUSTOM : NOTIFICATION_OPTION_DEFAULT
```

src/pages/settings/personal.svelte
hobbitronics:
```suggestion
    $: email_override = $user.email_override || ''
```

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+    const handleChecked = (policyId: string) => {
+      setNotice('Notification option has been saved')
+    }
+
+    const handleUnchecked = (policyId: string) => {
+      setNotice('Notification option has been saved')
+    }
+
+    async function onFileSelect(event: CustomEvent<FormData>) {
+      croppie = croppie || new Croppie(document.getElementById('croppie-container'), {
```
hobbitronics, line 62: There is a sveltier way of doing this which make accessing components easy.
```
let croppierContainer = {}

...
croppie = croppie || new Croppie(document.getElementById('croppieContainer'), {
...

<div bind:this={croppierContainer}>
```

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+
+      if (file) {
+        let reader = new FileReader()
+        reader.onload = e => {
+          croppie.bind({
+            url: e.target.result as string
+          })
+        }
+
+        reader.readAsDataURL(file)
```
hobbitronics, line 76: I thought this code looked familiar. The fileDropArea had a preview at one point which followed this pattern except for the croppie.bind part

src/pages/settings/personal.svelte
hobbitronics: Yes, it should be available as `file.id`

src/pages/settings/personal.svelte
```diff
@@ -0,0 +1,169 @@
...
+  <h3 class="ml-1 mt-3">Profile picture</h3>
+  <p>
+      <FileDropArea mimeType="image/*" class="w-50 mt-10px" raised {uploading} on:upload={onFileSelect} />
+  </p>
+  
+  <div class="photo-preview">
+    <div id="croppie-container"></div>
+  </div>
+  {#if croppie}
+  <Button raised on:click={onUpload}>Save</Button>
```
hobbitronics, line 165: I know vs code doesn't do this, but we indent inside {#if}'s

hobbitronics (commented):
Looks good. Just some mostly cosmetic suggestions, but great work.

src/pages/settings/personal.svelte
hobbitronics: oh, I was thinking it would come after croppie

src/components/FileDropArea/FileDropArea.svelte
RussellChamp: basically what this is saying is that this particular dispatcher will dispatch an event named 'upload' that passes a value of type 'FormData'

src/pages/settings/personal.svelte
RussellChamp: as i understand, that's what happening during `const file = await upload(data)` and here we're just saying "here's the API generated ID of the file that just got uploaded" right?

hobbitronics (commented):
Approved

src/pages/settings/personal.svelte
forevermatt: Ohhh... yeah, you may well be right. Thanks.

PR #140, opened 2021-09-17
Fix edit claim

hobbitronics:
The formData was just being sent to updateClaim, but needed to be split up, parsed and sent to updateClaim and updateClaimItem.

The api will need to add fields to updateClaim.


src/data/claims.ts
```diff
@@ -298,6 +305,12 @@ export async function submitClaim(claimId: string) {
   start(claimItemId)
 
   const parsedData: UpdateClaimItemRequestBody = {
+    // fmv: claimItemData.fairMarketValueUSD,
```
forevermatt, line 308: This will probably need to be converted to dollars (used in the UI code) to cents (what the API typically expects for money fields), just as a heads-up. We have a `convertToCents()` method in our money helper for exactly that purpose.

Same with the other `...USD` local variables we are sending to the API.

src/data/claims.ts
```diff
@@ -298,6 +305,12 @@ export async function submitClaim(claimId: string) {
...
+    // is_repairable: claimItemData.isRepairable,
+    // item_id: claimItemData.stringNumber,
```
forevermatt, line 310: Looks like this `stringNumber` was supposed to be `item_id`, I think.

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -38,7 +38,29 @@ $: items = $itemsByPolicyId[$user.policy_id] || []
 $: item = items.find(anItem => anItem.id === itemId) || {} as PolicyItem
 
 const onSubmit = async event => {
-  await updateClaim(claimId, event.detail)
+  const {claimData, claimItemData} = event.detail
+
+  const editedClaimData = {
```
Comment (forevermatt) [inline src/pages/claims/[claimId]/edit.svelte:43]: I think we've been doing this mapping/adjusting of form values in the functions that actually call the API and update our stores (like `updateClaim()`). Was there an advantage to doing that here in the page's code instead?

Comment (hobbitronics) [reply src/pages/claims/[claimId]/edit.svelte:43]: since we are using updateClaimItem, but with data from the claimForm which was originally for new claims and the createClaim endpoint I had to use an intermediate parsing step. I should be able to eliminate if for updateClaim though since we only use that once.

src/data/claims.ts
hobbitronics: I've invented a new type for js.

src/pages/claims/[claimId]/edit.svelte
Comment (hobbitronics) [reply src/pages/claims/[claimId]/edit.svelte]: looks like I was able to eliminate mostly for both.

Comment (hobbitronics) [reply src/pages/claims/[claimId]/edit.svelte]: I am wondering if this would be better though, but unsure what the api will do with it.
```
    repairActual: null,
    replaceActual: null,
```

RussellChamp (commented):
code lgtm

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -32,13 +32,23 @@ $: claimItems = claim.claim_items || []
...
 
 $: $user.policy_id && loadItems($user.policy_id)
 $: items = $itemsByPolicyId[$user.policy_id] || []
 $: item = items.find(anItem => anItem.id === itemId) || {} as PolicyItem
 
 const onSubmit = async event => {
-  await updateClaim(claimId, event.detail)
+  const {claimData, claimItemData} = event.detail
+
+  const editedClaimItemData = {
```
Comment (forevermatt) [inline src/pages/claims/[claimId]/edit.svelte:44]: I think it would be good to do even this in the `updateClaimItem()` method, so that any other code that calls `updateClaimItem()` doesn't have to also implement this edit.

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -32,13 +32,23 @@ $: claimItems = claim.claim_items || []
...
+    ...claimItemData,
+    repairActual: 0,
+    replaceActual: 0,
+  }
+  
+  await updateClaim(claimId, claimData)
+  await updateClaimItem(ClaimItemId, editedClaimItemData)
```
Comment (forevermatt) [inline src/pages/claims/[claimId]/edit.svelte:51]:
```suggestion
  await updateClaimItem(claimItemId, editedClaimItemData)
```

forevermatt (commented):
Final comments added. Almost there, thanks! 🙂  (Sorry I've once again been so slow to finish reviewing this...)

forevermatt (changes requested):
The only two remaining things I see are moving the `repairActual: 0, replaceActual: 0,` bit down into `updateClaimItem()` and then lowercasing the "c" on the "ClaimItemId" variable.

forevermatt (approved):

PR #141, opened 2021-09-20
fix item coverage submission

RussellChamp:
* save as draft or pending depending on which button was clicked
* fix a bunch of breadcrumbs on pages


src/components/forms/ItemForm.svelte
```diff
@@ -85,7 +85,8 @@ const onSubmit = () => {
   dispatch('submit', getFormData())
 }
 const saveForLater = () => {
-  dispatch('save-for-later', getFormData())
+  coverageStatus = 'Draft'
+  // then onSubmit is automatically called by the form
```
RussellChamp, line 89: so what confused me in the code here was that _REGARDLESS OF WHICH BUTTON IS PRESSED_ the `onSubmit` method is always called. not sure what's up with that...

src/pages/claims.svelte
```diff
@@ -13,6 +13,7 @@ $: items = $itemsByPolicyId[policyId] || []
 </script>
 
 <Page layout="grid">
+  <Breadcrumb />
```
RussellChamp, line 16: all the figmas had breadcrumbs and they were easy to add

src/pages/claims/[claimId].svelte
```diff
@@ -132,65 +140,77 @@ function onDeleted(event) {
...
-    </Banner>
-    <div class="left-detail">{incidentDate || ''}</div>
-  </Row>
-  <Row cols="9">
-    <ClaimBanner claimStatus={status} >{claim.status_reason || ''}</ClaimBanner>
-    {#if needsFile}
-      <ClaimBanner claimStatus={`${status}2`} >
-        Upload {uploadLabel} to get reimbursed.
-      </ClaimBanner>
+  {#if !item.id }
```
Comment (RussellChamp) [inline src/pages/claims/[claimId].svelte:143]: mirrors what in the item details page

Comment (RussellChamp) [reply src/pages/claims/[claimId].svelte:143]: the other "changes" is just an indentation change

src/pages/items/[itemId]/edit.svelte
```diff
@@ -8,37 +8,22 @@ import { Page } from '@silintl/ui-components'
...
+const itemsBreadcrumb = { name: 'Items', url: '/items' }
+$: thisItemBreadcrumb = { name: itemName || 'This item', url: `/items/${itemId}` }
+const editBreadcrumb =   { name: "Edit", url: `/items/${itemId}/edit` }
+$: breadcrumbLinks = [itemsBreadcrumb, thisItemBreadcrumb, editBreadcrumb]
 
 const onSubmit = async event => {
   await updateItem(policyId, itemId, event.detail)
   $goto(`/items/${itemId}`)
 }
-const onSaveForLater = async event => {
```
Comment (RussellChamp) [inline src/pages/items/[itemId]/edit.svelte:36]: in either case we're just submitting an item to be created. the form has already specified if it is a 'Draft' or 'Pending' this code is not needed

src/pages/claims/[claimId].svelte
```diff
@@ -1,13 +1,14 @@
...
+import { Banner, ClaimBanner, ConvertCurrencyLink, FileDropArea, FilePreview, MoneyInput, Row, Breadcrumb } from '../../components'
+import { formatDate } from '../../components/dates'
+import { loading } from '../../components/progress'
 import { upload } from '../../data'
 import { loadClaims, claims, initialized, claimsFileAttach, updateClaimItem, Claim, ClaimItem, ClaimFile, ClaimFilePurpose, PayoutOption, submitClaim } from '../../data/claims'
 import { loadItems, itemsByPolicyId, PolicyItem } from '../../data/items'
 import { formatMoney } from '../../helpers/money'
 import { goto } from '@roxi/routify'
-import { Button, Page } from '@silintl/ui-components'
+import { Button, Page, } from '@silintl/ui-components'
```
Comment (hobbitronics) [inline src/pages/claims/[claimId].svelte:11]: was this comma added intentionally?

src/components/forms/ItemForm.svelte
hobbitronics: I think we were maybe anticipating different endpoints but then things changed.

hobbitronics (approved):
Looks good.

src/pages/items/[itemId]/edit.svelte
```diff
@@ -8,37 +8,22 @@ import { Page } from '@silintl/ui-components'
...
+$: itemName = item.name || ''
+
+// Dynamic breadcrumbs data:
+const itemsBreadcrumb = { name: 'Items', url: '/items' }
+$: thisItemBreadcrumb = { name: itemName || 'This item', url: `/items/${itemId}` }
+const editBreadcrumb =   { name: "Edit", url: `/items/${itemId}/edit` }
+$: breadcrumbLinks = [itemsBreadcrumb, thisItemBreadcrumb, editBreadcrumb]
 
 const onSubmit = async event => {
   await updateItem(policyId, itemId, event.detail)
```
Comment (hobbitronics) [inline src/pages/items/[itemId]/edit.svelte:24]: should we be adding the call to https://s3.amazonaws.com/static.gtis.guru/cover/api-docs.html#operation/PolicyItemsSubmit

hobbitronics (changes requested):
I think I see the issue. When we are editing a Draft, we need to use the PolicyItemSubmit after updateItem.

src/components/forms/ItemForm.svelte
forevermatt: Ahh, my mistake. I think that means the `on:click={saveForLater}` should have been `on:click|preventDefault={saveForLater}` to prevent that button click from also submitting the form.

src/components/forms/ItemForm.svelte
```diff
@@ -85,7 +85,8 @@ const onSubmit = () => {
   dispatch('submit', getFormData())
 }
 const saveForLater = () => {
-  dispatch('save-for-later', getFormData())
+  coverageStatus = 'Draft'
```
forevermatt, line 88: I don't think we get to set the status directly. We create the item (which automatically results in it being in a Draft state), and then also (if they were ready to submit it for coverage) call the [PolicyItemsSubmit](https://static.gtis.guru/cover/api-docs.html#operation/PolicyItemsSubmit) endpoint.

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -37,6 +21,13 @@ $: $user.policy_id && loadItems($user.policy_id)
 $: items = $itemsByPolicyId[$user.policy_id] || []
 $: item = items.find(anItem => anItem.id === itemId) || {} as PolicyItem
 
+// Dynamic breadcrumbs data:
+$: claimName = `${item.name} (${claim.reference_number})`
+const claimsBreadcrumb = { name: 'Claims', url: '/claims' }
+$: thisClaimBreadcrumb = { name: claimName || 'This item', url: `/claims/${claimId}` }
+const editBreadcrumb =   { name: "Edit", url: `/claims/${claimId}/edit` }
+$: breadcrumbLinks = [claimsBreadcrumb, thisClaimBreadcrumb, editBreadcrumb]
```
Comment (forevermatt) [inline src/pages/claims/[claimId]/edit.svelte:29]: This isn't any criticism of your work here, @RussellChamp, I'm just wondering...

It seems like we're ending up needing to duplicate a fair bit of code for the dynamic breadcrumbs. What if we tweak the Breadcrumbs component to simply take another parameter for an additional (non-link) item at the end of the breadcrumbs, for representing the current page? It seems odd to me to have a link in the breadcrumbs to the page the user is currently on, so we could simply remove that from the list of auto-generated links.

Thoughts?

src/pages/claims/[claimId].svelte
Comment (RussellChamp) [reply src/pages/claims/[claimId].svelte:11]: no.

src/pages/claims/[claimId]/edit.svelte
Comment (RussellChamp) [reply src/pages/claims/[claimId]/edit.svelte:29]: i do not feel strongly one way or another about this

src/pages/items/[itemId]/edit.svelte
Comment (forevermatt) [reply src/pages/items/[itemId]/edit.svelte:36]: I think this will be needed...

Comment (forevermatt) [reply src/pages/items/[itemId]/edit.svelte:36]: Oh, this is for editing an item. Hmm... yeah, you could well be right. I think we will need the separate "save-for-later" handler when creating a new item, though (below).

src/components/forms/ItemForm.svelte
```diff
@@ -74,44 +75,38 @@ const getFormData = () => {
...
+const onSubmit = (event: Event) => {
+  shouldSubmit = true
   dispatch('submit', getFormData())
 }
-const saveForLater = () => {
-  coverageStatus = 'Draft'
-  // then onSubmit is automatically called by the form
+
+const saveForLater = (event: Event) => {
+  dispatch('submit', getFormData())
```
forevermatt, line 94: This should be 'save-for-later'
```suggestion
  dispatch('save-for-later', getFormData())
```

src/components/forms/ItemForm.svelte
```diff
@@ -74,44 +75,38 @@ const getFormData = () => {
...
+  event.preventDefault()
```
forevermatt, line 95: Let's use [Svelte's way of the prevent-default](https://svelte.dev/tutorial/event-modifiers), rather than calling this manually, if you don't mind

src/components/forms/ItemForm.svelte
RussellChamp: i removed that event. do you think i should add it back?

src/components/forms/ItemForm.svelte
RussellChamp: i was getting some warning squiggly when trying to do it that way and thought i did something wrong
image

PR #142, opened 2021-09-20
simplify Menu closing

hobbitronics:
While reviewing the Menu ui-components PR I realized the syncToggler event isn't needed and doesn't fire.


RussellChamp (approved):
for my own learning, what was that code originally trying to do?

PR #147, opened 2021-09-21
add prettier support to the project

RussellChamp:
* Add prettier to node_modules and vscode recommended extensions
* Add prettier-plugin-svelte to node_modules
* Add formatting rules to settings.json and .prettierrc
* Add .prettierignore
* Format ALL the files!


.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "files.trimTrailingWhitespace": true,
```
forevermatt, line 2:
```suggestion
  "files.trimTrailingWhitespace": false,
```

.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
...
+
+  "[typescript]": {
+    "editor.formatOnSave": true,
+    "editor.formatOnPaste": true,
+  },
+
+  "[svelte]": {
+    "editor.formatOnSave": true,
+    "editor.formatOnPaste": true
+}
```
forevermatt, line 14:
```suggestion
  }
```

RussellChamp: i'll tap that when working on the eslint changes so i don't trigger a new cloudflare build

forevermatt (approved):
Approved based on partial skim. 👍

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+    "printWidth": 120,
+    "semi": false,
+    "tabWidth": 2,
+    "singleQuote": true,
+    "useTabs": false,
+
+    "svelteSortOrder": "options-scripts-styles-markup",
+    "svelteAllowShorthand": true,
+    "svelteIndentScriptAndStyle": false
```
hobbitronics, line 10: I thought the default was true, but it seems to be formatting as if ```"svelteBracketNewLine": false```, I personally don't like > on a new line

src/components/Breadcrumb.svelte
```diff
@@ -43,20 +43,26 @@ a {
...
 }
 </style>
 
 <div class="flex text-align-center align-items-center {$$props.class}">
   <!-- svelte-ignore a11y-invalid-attribute -->
-  {#if hasHome }
-  <a href="/home" class="capitalize"><i class="material-icons mdc-list-item__graphic money-icon breadcrumb-icon breadcrumb-home" aria-hidden="true">home</i></a><!--
-  --><i class="material-icons mdc-list-item__graphic money-icon breadcrumb-icon" aria-hidden="true">chevron_right</i>
+  {#if hasHome}
+    <a href="/home" class="capitalize"
```
hobbitronics, line 53: Any idea why it is adding a new line before > here, but not other places?

hobbitronics: oh, maybe its only when the width is over 120?

.prettierrc
hobbitronics: How do you guys feel about ```"arrowParens": "avoid"```, not a big deal for me, but I think it looks better

forevermatt: I'm not a overall a fan of angle-brackets being on a new line, but I could go either way. I think some of the weirdness comes from the tooling trying to avoid inserting whitespace where there wasn't any, and we can manually put those back (adding the newline ourselves).

forevermatt: My first inclination on the arrowParens option was like yours, @hobbitronics. However, the reasoning is that you have to use parens when adding a type, and so always using the parens is more consistent. And "[c]onsistent use of parentheses provides a better developer experience when editing real codebases" [[1]](https://prettier.io/docs/en/options.html#arrow-function-parentheses).

Anyway, I'm fine with either option on that one, too.

src/components/Breadcrumb.svelte
forevermatt: This looks like one where it's avoiding adding whitespace inside a "display: inline" tag (like `<a>`). I think we can manually add those line breaks and it will accept them.

PR #148, opened 2021-09-21
Extract claim actions to component, fix Policy ID bug

forevermatt:
### Added
- Extract buttons (for doing something to a claim) to new component

### Fixed
- Use `claim.policy_id` (not `$user.policy_id`) when viewing a claim's details

---

My intent is to hide the complexity of which-button-to-show-when-and-to-whom logic in it's own component. I just haven't yet actually added any buttons (such as for the steward to approve or deny a claim).


src/pages/claims/[claimId].svelte
```diff
@@ -1,14 +1,24 @@
 <script lang="ts">
-import user from '../../authn/user'
-import { Banner, Breadcrumb, ClaimBanner, ConvertCurrencyLink, FileDropArea, FilePreview, MoneyInput, Row } from '../../components'
+import user from '../../authn/user.js'
```
Comment (RussellChamp) [inline src/pages/claims/[claimId].svelte:2]: remove the extraneous '.js'
```suggestion
import user from '../../authn/user'
```

RussellChamp (commented):
The code looks fine but how useful is it to extract out a "component" that you're only going to use in one place? (unless i'm missing something) It seems like it's just creating unnecessarily-long event chains

hobbitronics (approved):
looks good.

forevermatt:
> The code looks fine but how useful is it to extract out a "component" that you're only going to use in one place? (unless i'm missing something) It seems like it's just creating unnecessarily-long event chains

The benefit is that it reduces the complexity of the claim details page itself, by extracting those 25+ (and eventually 100+) lines of code--which are a cohesive unit unto themselves--into their own space. We're hiding the "how" of this part of the page, so that the page simply shows the "what" there (which is all you need to know in the context of the page).

PR #149, opened 2021-09-21
Return an empty string from `formatMoney()` if cents are NaN

forevermatt:
### Fixed
- Return an empty string from `formatMoney()` if cents are NaN

### DIFF
--- a/src/helpers/money.ts
+++ b/src/helpers/money.ts
@@ -7,7 +7,7 @@ export const convertToCents = dollars => {
 }
 
 export const formatMoney = cents => {
-  if (cents === undefined || Number.isNaN(cents)) {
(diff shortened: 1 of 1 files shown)


hobbitronics (approved):

src/helpers/money.ts
```diff
@@ -7,7 +7,7 @@ export const convertToCents = dollars => {
 }
 
 export const formatMoney = cents => {
-  if (cents === undefined) {
+  if (cents === undefined || Number.isNaN(cents)) {
```
RussellChamp, line 10: this conditional does not fully protect you from bad data in the case that 'cents' is either `null` or a non-numeric string. a more robust check for this particular case might be as follows.
```suggestion
  if (!cents || !Number.isFinite(+cents)) {
```

PR #150, opened 2021-09-21
moved determineMaxPayout to business-rules

hobbitronics:
I might have moved more to business-rules but wanted to keep it short.

I also fixed a but where Math.min() was choking on undefined.

### DIFF
--- a/src/business-rules/claim-payout-amount.ts
+++ b/src/business-rules/claim-payout-amount.ts
@@ -1,6 +1,6 @@
 import type { PayoutOption, ClaimItem } from '../data/claims'
 
-export const deductible = .05
+export const DEDUCTIBLE = .05
 export const LOSS_REASON_EVACUATION = 'Evacuation'
 
 export const PAYOUT_OPTION_FIXED_FRACTION: PayoutOption = 'FixedFraction'
@@ -48,7 +48,7 @@ export const isUnrepairableOrTooExpensive = (isRepairable, repairCostIsTooHigh) 
 
 const computePayout = (...values) => {
   const filteredValues = [...values]?.filter(value => value !== undefined)
-  return Math.min(...filteredValues) * (1 - deductible)
+  return Math.min(...filteredValues) * (1 - DEDUCTIBLE)
 }
 
 const computeRepairMaxPayout = (claimItem: ClaimItem, coverageAmount) => computePayout(claimItem.repair_actual || claimItem.repair_estimate, coverageAmount, claimItem.fmv)
--- a/src/business-rules/claim-payout-amount.ts
+++ b/src/business-rules/claim-payout-amount.ts
@@ -57,12 +57,21 @@ const computeReplaceMaxPayout = (claimItem: ClaimItem, coverageAmount) => comput
 
 const computeCashMaxPayout = (claimItem: ClaimItem, coverageAmount) => computePayout(coverageAmount, claimItem.fmv)
 
-export const determineMaxPayout = (payoutOption, claimItem: ClaimItem, coverageAmount) => {
-  const calculationsByPayoutOption = {
-    Repair: () => computeRepairMaxPayout(claimItem, coverageAmount),
-    Replacement: () => computeReplaceMaxPayout(claimItem, coverageAmount),
-    FMV: () => computeCashMaxPayout(claimItem, coverageAmount),
-    FixedFraction: () => coverageAmount * 2/3,
+export const determineMaxPayout = (payoutOption: PayoutOption, claimItem: ClaimItem, coverageAmount) => {
+  switch (payoutOption) {
+    case PAYOUT_OPTION_REPAIR:
+      return computeRepairMaxPayout(claimItem, coverageAmount)
+      break
+    case PAYOUT_OPTION_REPLACE:
+      return computeReplaceMaxPayout(claimItem, coverageAmount)
+      break
+    case PAYOUT_OPTION_FMV:
+      return computeCashMaxPayout(claimItem, coverageAmount)
+      break
+    case PAYOUT_OPTION_FIXED_FRACTION:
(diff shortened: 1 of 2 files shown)


src/business-rules/claim-payout-amount.ts
```diff
@@ -44,3 +46,26 @@ export const isUnrepairableOrTooExpensive = (isRepairable, repairCostIsTooHigh)
...
+export const determineMaxPayout = (payoutOption, claimItem: ClaimItem, coverageAmount) => {
+  if(payoutOption === PAYOUT_OPTION_REPAIR) {
+    return computeRepairMaxPayout(claimItem, coverageAmount)
+  } else if(payoutOption === PAYOUT_OPTION_REPLACE) {
+    return computeReplaceMaxPayout(claimItem, coverageAmount)
+  } else if(payoutOption === PAYOUT_OPTION_FMV) {
+    return computeCashMaxPayout(claimItem, coverageAmount)
+  } else if(payoutOption === PAYOUT_OPTION_FIXED_FRACTION) {
+    return formatMoney(coverageAmount * 2/3) || ''
+  }
```
forevermatt, line 70: This is totally just a suggestion... I mostly wanted to see how it would look if we did it this way.

I tend to find multiple if/else-if/else's chained together to be somewhat hard to read. If you're interested, this is a way to do it as an object, so there is less boiler-plate code.:
```suggestion
  const calculationsByPayoutOption = {
    PAYOUT_OPTION_REPAIR: () => computeRepairMaxPayout(claimItem, coverageAmount),
    PAYOUT_OPTION_REPLACE: () => computeReplaceMaxPayout(claimItem, coverageAmount),
    PAYOUT_OPTION_FMV: () => computeCashMaxPayout(claimItem, coverageAmount),
    PAYOUT_OPTION_FIXED_FRACTION: () => formatMoney(coverageAmount * 2/3) || '',
  }
  return calculationsByPayoutOption[payoutOption]()
```

You are more than welcome to keep your code as-is, I'm just tossing this out there as another option.

src/business-rules/claim-payout-amount.ts
```diff
@@ -44,3 +46,26 @@ export const isUnrepairableOrTooExpensive = (isRepairable, repairCostIsTooHigh)
   
   return repairCostIsTooHigh
 }
+
+const computePayout = (...values) => {
+  const filteredValues = [...values]?.filter(value => value !== undefined)
+  return formatMoney(Math.min(...filteredValues) * (1 - deductible))
```
forevermatt, line 52: This will probably work fine, but it might be helpful to leave it as a number until actually inserting it into the HTML somewhere. That helps us keep our formatting/encoding separate from our numerical calculations.

forevermatt (approved):
Approved, with some ideas of possible tweaks if desired

src/business-rules/claim-payout-amount.ts
hobbitronics: I have formatted it in the markup now.

src/business-rules/claim-payout-amount.ts
```diff
@@ -44,3 +45,24 @@ export const isUnrepairableOrTooExpensive = (isRepairable, repairCostIsTooHigh)
...
+  return Math.min(...filteredValues) * (1 - deductible)
+}
+
+const computeRepairMaxPayout = (claimItem: ClaimItem, coverageAmount) => computePayout(claimItem.repair_actual || claimItem.repair_estimate, coverageAmount, claimItem.fmv)
+
+const computeReplaceMaxPayout = (claimItem: ClaimItem, coverageAmount) => computePayout(claimItem.replace_estimate, coverageAmount)
+
+const computeCashMaxPayout = (claimItem: ClaimItem, coverageAmount) => computePayout(coverageAmount, claimItem.fmv)
+
+export const determineMaxPayout = (payoutOption, claimItem: ClaimItem, coverageAmount) => {
```
RussellChamp, line 60: this method will throw an uncaught TypeError exception if called with a `payoutOption` not included in your options. you can create a type here to try and assist you during development time but it won't protect you during execution time so you'd still want to check that the specified value exists.
Type example: `type PayoutOption = 'Repair' | 'Replacement' | 'FMV' | 'FixedFraction'`
Exception check: `if (payoutOption in calculationsByPayoutOption) { return ... } else { return DEFAULT_VALUE }`

src/pages/claims/[claimId].svelte
```diff
@@ -43,30 +42,14 @@ $: uploadLabel = getUploadLabel(claimItem, needsReceipt, receiptType) as string
...
-$: if(payoutOption === 'Repair') {
-    maximumPayout = computeRepairMaxPayout()
-  } else if(payoutOption === 'Replacement') {
-    maximumPayout = computeReplaceMaxPayout()
-  } else if(payoutOption === 'FMV') {
-    maximumPayout = computeCashMaxPayout()
-  } else if(claim.incident_type === 'Evacuation') {
-    maximumPayout = formatMoney(item.coverage_amount * 2/3) || ''
-  }
+$: maximumPayout = determineMaxPayout(payoutOption, claimItem, item.coverage_amount) as string
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:45]: This is more of a minor thing, but now that we're waiting to format this as a money string, the return value of `determineMaxPayout()` would (I think) be either a number or possibly a nullable number (for situations where an invalid type is passed, where we may want to return null since we don't know what the max payout is in that case).

src/business-rules/claim-payout-amount.ts
```diff
@@ -1,5 +1,6 @@
-import type { PayoutOption } from '../data/claims'
+import type { PayoutOption, ClaimItem } from '../data/claims'
 
+export const deductible = .05
```
RussellChamp, line 3: CAPITALIZE the const value

forevermatt (approved):

PR #151, opened 2021-09-22
add eslint support

RussellChamp:
* Add the eslint configuration files and vscode extension recomendation
* Auto-fix existing files


tsconfig.json
```diff
@@ -9,6 +9,7 @@
         "src/customTypes",
         "node_modules/@types"
       ],
-      "resolveJsonModule": true
+      "resolveJsonModule": true,
+      "strict": true
```
hobbitronics, line 13: I am pretty sure this doesn't effect the strict mode for Svelte, but just want to be clear on the implications.

RussellChamp: this is for the typescript compiler itself. usually it should be set but i hadn't done so before. tbh i was also following along an eslint, svelte, typescript blog post https://codechips.me/eslint-svelte-typescript/ and they recommended it

forevermatt (approved):

PR #152, opened 2021-09-22
Hide file claim and refactor states

hobbitronics:
### Changed
- Moved states from claims.ts to states.ts
- gave states.title claim and item properties as they often differ
- Items view - hide File Claim button unless 'Approved' and only show Edit Item on Draft or Pending

### Fixed
- a number of ts errors

### DIFF
--- a/src/components/banners/ClaimBanner.svelte
+++ b/src/components/banners/ClaimBanner.svelte
@@ -1,15 +1,16 @@
 <script lang="ts">
 import { Banner } from '../../components'
-import { getState, State, Status } from '../../data/states'
+import { getState, State } from '../../data/states'
 
-export let claimStatus = '' as Status
+export let claimStatus = ''
+export let isClaim: boolean = false
 
 $: state = (claimStatus && getState(claimStatus)) || ({} as State)
 
 $: bgColor = state.bgColor || ''
 $: color = state.color || ''
 $: icon = state.icon || ''
-$: title = state.title || ''
+$: title = isClaim ? state.title.claim : state.title.item || ''
 </script>
 
 <style>
--- a/src/components/banners/ClaimBanner.svelte
+++ b/src/components/banners/ClaimBanner.svelte
@@ -1,8 +1,8 @@
 <script lang="ts">
 import { Banner } from '../../components'
-import { getState, State } from '../../data/claims'
+import { getState, State, Status } from '../../data/states'
 
-export let claimStatus = ''
+export let claimStatus = '' as Status
 
 $: state = (claimStatus && getState(claimStatus)) || ({} as State)
 
--- a/src/data/items.ts
+++ b/src/data/items.ts
@@ -118,7 +118,7 @@ export async function addItem(policyId: string, itemData: any) {
     serial_number: itemData.uniqueIdentifier,
   }
 
-  const addedItem = await CREATE<PolicyItem>(urlPath, parsedItemData)
+  const addedItem = await CREATE<PolicyItem>(urlPath, parsedItemData as any)
 
   itemsByPolicyId.update((data) => {
     const items = data[policyId] || []
--- a/src/data/items.ts
+++ b/src/data/items.ts
@@ -98,8 +98,8 @@ export async function loadItems(policyId: string) {
  * @param {Object} itemData
  * @return {Object}
  */
-export async function addItem(policyId: string, itemData) {
-  const urlPath = `policies/${policyId}/items`
+export async function addItem(policyId: string, itemData: any) {
+  const urlPath: string = `policies/${policyId}/items`
   start(urlPath)
 
   const parsedItemData: CreatePolicyItemRequestBody = {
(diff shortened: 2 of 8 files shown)


src/components/banners/ClaimBanner.svelte
```diff
@@ -1,15 +1,16 @@
 <script lang="ts">
 import { Banner } from '../../components'
-import { getState, State } from '../../data/claims'
+import { getState, State } from '../../data/states'
 
 export let claimStatus = ''
+export let isClaim: boolean = false
```
forevermatt, line 6: I think it may be better to split this out into an `ItemBanner` rather than showing both claim and item information in the ClaimBanner.

src/data/items.ts
```diff
@@ -118,7 +118,7 @@ export async function addItem(policyId: string, itemData) {
     serial_number: itemData.uniqueIdentifier,
   }
 
-  const addedItem = await CREATE<PolicyItem>(urlPath, parsedItemData)
+  const addedItem = await CREATE<PolicyItem>(urlPath, parsedItemData as any)
```
forevermatt, line 121: I don't think we should have to specify `any` here, since we're passing in a typed item anyway. Ohhh... it's probably because `CREATE()` expects the second parameter to be an `any`, isn't it?

src/data/states.ts
```diff
@@ -0,0 +1,79 @@
+import type { ClaimStatus } from './claims'
+import type { ItemCoverageStatus } from './items'
+
+export type Status = ClaimStatus | ItemCoverageStatus
```
forevermatt, line 4: Here as well, I think it may be worth splitting this file out to two, perhaps a `data/item-states.ts` and a `data/claim-states.ts`, since we always know which we're dealing with. That avoids the need to add code in to sort them out later.

src/pages/items/[itemId].svelte
```diff
@@ -77,7 +80,7 @@ const goToNewClaim = () => {
     </Row>
 
     <Row cols={'9'}>
```
Comment (forevermatt) [inline src/pages/items/[itemId].svelte:82]: I know this is unrelated, but I think this `cols={'9'}` could be simplified at some point to `cols="9"`.

src/data/states.ts
forevermatt: What would you think of extracting a `<StatusBanner />` component from this, and having a `<ClaimBanner />` and `<ItemBanner />` that each use that internally and thus each essentially only have the code that's different?

I'm still not comfortable with us multi-purposing the `<ClaimBanner />` to also be used for items, since it's specifically named and designed to be the banner for claims. It feels like we're combining two things unnecessarily, then having to add logic to sort them back out later, even though we always know which we're trying to use (i.e. it doesn't depend on dynamic data).

src/components/banners/ClaimBanner.svelte
hobbitronics: I was able to do this in a way that recycled most of the code by creating another element and eliminated the need for this prop.

src/data/states.ts
hobbitronics: ok, I have refactored states into claimStates and ItemStates, although they are still in the same file using some of the same resources so I didn't need to repeat myself.

RussellChamp (changes requested):
I feel moderately strongly that `null` values should not be used unless there is a very good and specific reason for the usage

hobbitronics:
> I feel moderately strongly that `null` values should not be used unless there is a very good and specific reason for the usage

Are you saying in general? I am using then here since a draft claimItem shoudn't have those values, but maybe the backend should/will take care of this.

RussellChamp:
> > I feel moderately strongly that `null` values should not be used unless there is a very good and specific reason for the usage
> 
> Are you saying in general? I am using then here since a draft claimItem shoudn't have those values, but maybe the backend should/will take care of this.

In this specific case I think they should be `undefined` and not `null`. this would also allow for better types (`foo?: string` instead of `foo: string | null`) as well unless the BE is _specifically_ expecting a value `null` (which i doubt because i think most BEs will treat `undefined` and `null` similarly)

src/data/states.ts
```diff
@@ -2,78 +2,92 @@ import type { ClaimStatus } from './claims'
...
   },
-  Draft2: warning,
   Pending: pending,
-  Receipt: success,
-  Receipt2: warning,
+}
+
+export const claimStates: { [stateName: string]: State } = {
+  ...commonStates,
+  Approved: { ...approved, title: 'Approved for payout' },
```
forevermatt, line 50: I really like how you did this. 👍

src/data/states.ts
forevermatt: Thank for splitting those. Looks great!

At this point, it looks like nothing is using `Status` (though I could easily have missed something). Instead they're all using either `ClaimStatus` or `ItemCoverageStatus`, so I suspect this const could be removed, too.

forevermatt (approved):

src/data/states.ts
hobbitronics: thanks, no class required.

src/data/states.ts
hobbitronics: yup, seems like it was not used anymore, deleted it.

forevermatt (approved):

PR #153, opened 2021-09-22
remove "Receive notification emails for" from user profile

Schparky:
Not on the roadmap, so removing it. If commenting it out is preferred, let me know.

This is what it looks like for an admin:
image


RussellChamp:
i'd probably prefer some form of a `currentDevelopment` flag that disables the code instead of deleting it outright

Schparky:
> i'd probably prefer some form of a `currentDevelopment` flag that disables the code instead of deleting it outright

Since we don't have that (yet) I'll use `#if 0` and `if (false)` for now

forevermatt (approved):

PR #154, opened 2021-09-23
remove extraneous calls to start/stop progress bar

RussellChamp:
* also improve the error message system
* also add return values on some API call that were lacking them. this will make testing easier down the road
* also add a few type in files i was editing anyways

### DIFF
--- a/src/data/claim-incident-types.ts
+++ b/src/data/claim-incident-types.ts
@@ -1,5 +1,4 @@
 import { GET } from '.'
-import { start, stop } from '../components/progress'
 import { writable } from 'svelte/store'
 
 export type ClaimIncidentType = {
@@ -9,13 +8,12 @@ export type ClaimIncidentType = {
 
 export const claimIncidentTypes = writable<ClaimIncidentType[]>([])
 
-export async function loadClaimIncidentTypes() {
+export async function loadClaimIncidentTypes(): Promise<ClaimIncidentType[]> {
   const urlPath = 'config/claim-incident-types'
-  start(urlPath)
 
   const results = await GET<ClaimIncidentType[]>(urlPath)
 
-  stop(urlPath)
-
   claimIncidentTypes.set(results)
+
+  return results
 }
--- a/src/data/claim-incident-types.ts
+++ b/src/data/claim-incident-types.ts
@@ -8,12 +8,10 @@ export type ClaimIncidentType = {
 
 export const claimIncidentTypes = writable<ClaimIncidentType[]>([])
 
-export async function loadClaimIncidentTypes(): Promise<ClaimIncidentType[]> {
+export async function loadClaimIncidentTypes(): Promise<void> {
   const urlPath = 'config/claim-incident-types'
 
   const results = await GET<ClaimIncidentType[]>(urlPath)
 
   claimIncidentTypes.set(results)
-
-  return results
 }
--- a/src/data/claims.ts
+++ b/src/data/claims.ts
@@ -134,9 +134,8 @@ export function init(): void {
  * @export
  * @param {Object} item
  * @param {Object} claimData
- * @return {Object} -- The newly created Claim
  */
-export async function createClaim(item: PolicyItem, claimData: any): Promise<Claim> {
+export async function createClaim(item: PolicyItem, claimData: any): Promise<void> {
   const urlPath = `policies/${item.policy_id}/claims`
 
   const parsedClaim: CreateClaimRequestBody = {
@@ -151,11 +150,9 @@ export async function createClaim(item: PolicyItem, claimData: any): Promise<Cla
     currClaims.push(claim)
     return currClaims
   })
-
-  return claim
 }
 
-export const createClaimItem = async (claimId: string, claimItemData: any): Promise<ClaimItem> => {
+export const createClaimItem = async (claimId: string, claimItemData: any): Promise<void> => {
   const urlPath = `claims/${claimId}/items`
 
   const parsedClaimItem: CreateClaimItemRequestBody = {
@@ -177,8 +174,6 @@ export const createClaimItem = async (claimId: string, claimItemData: any): Prom
     }
     return claims
   })
-
-  return claimItem
 }
 
 /**
@@ -188,7 +183,7 @@ export const createClaimItem = async (claimId: string, claimItemData: any): Prom
  * @param {String} claimId
  * @param {Object} newClaimData
  */
-export async function updateClaim(claimId: string, newClaimData: any): Promise<Claim> {
+export async function updateClaim(claimId: string, newClaimData: any): Promise<void> {
   const parsedData: UpdateClaimRequestBody = {
     incident_date: new Date(newClaimData.lostDate).toISOString(),
     incident_type: newClaimData.lossReason,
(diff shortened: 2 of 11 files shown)


src/data/claim-incident-types.ts
```diff
@@ -9,13 +8,12 @@ export type ClaimIncidentType = {
 
 export const claimIncidentTypes = writable<ClaimIncidentType[]>([])
 
-export async function loadClaimIncidentTypes() {
+export async function loadClaimIncidentTypes(): Promise<ClaimIncidentType[]> {
   const urlPath = 'config/claim-incident-types'
-  start(urlPath)
 
   const results = await GET<ClaimIncidentType[]>(urlPath)
```
forevermatt, line 14: Now that we don't need the `urlPath` as the parameter to start/stop, you're welcome to inline this string (or not, as you like).

src/data/claim-incident-types.ts
```diff
@@ -9,13 +8,12 @@ export type ClaimIncidentType = {
...
 
-  stop(urlPath)
-
   claimIncidentTypes.set(results)
+
+  return results
```
forevermatt, line 18: I could be wrong, but I think we intentionally didn't return the results here, since we want our code to be using the `claimIncidentTypes` (which will auto-update), not the return value of this (which won't).

RussellChamp: i think it's still worthwhile to have return types to aid in unit testing down the line

forevermatt: We do return a Promise, though, because we want to wait to redirect the user in case there was an error response from the API.

src/data/claim-incident-types.ts
RussellChamp: i actually prefer the separate 'urlPath' line because it helps to keep things neat and avoid line-wrapping with the uri gets really long

src/data/claims.ts
```diff
@@ -154,36 +152,33 @@ export async function createClaim(item: PolicyItem, claimData: any) {
...
   })
 
-  stop(urlPath)
   return claim
 }
 
-export const createClaimItem = async (claimId: string, claimItemData: any) => {
+export const createClaimItem = async (claimId: string, claimItemData: any): Promise<ClaimItem> => {
   const urlPath = `claims/${claimId}/items`
-  start(urlPath)
```
forevermatt, line 163: So happy to be losing all this duplicated, boiler-plate code. 😁

src/data/claim-incident-types.ts
RussellChamp: well some of the API call functions did have return types already. i figured it would be good to standardize the practice. if you feel that returning values from API call functions could encourage bad behavior, i'll remove the values.

src/data/claims.ts
```diff
@@ -210,31 +203,30 @@ export async function updateClaim(claimId: string, newClaimData: any) {
...
+  const response = await CREATE<ClaimsFileAttachResponseBody>(`claims/${claimId}/files`, data as any)
 
-export async function submitClaim(claimId: string) {
-  start(claimId)
+  return response
+}
 
-  await CREATE<string>(`claims/${claimId}/submit`)
+export async function submitClaim(claimId: string): Promise<string> {
+  const response = await CREATE<string>(`claims/${claimId}/submit`)
```
forevermatt, line 225: This is probably an item for a later PR, but this endpoint actually returns the submitted claim, not merely a string. It would probably even be good to update that claim in our local list of claims.

src/data/claims.ts
```diff
@@ -267,26 +257,21 @@ export async function updateClaimItem(claimItemId: string, claimItemData: any) {
...
+export async function loadClaims(): Promise<Claim[]> {
+  const response = await GET<Claim[]>('claims')
 
-  claims.set(clms)
-
-  stop('loadClaims')
+  claims.set(response)
   initialized.set(true)
+
+  return response
```
forevermatt, line 276: This is another place where by design want don't want to return the response, just update the store.

src/data/dependents.ts
```diff
@@ -124,15 +119,14 @@ export async function updateDependent(policyId: string, dependentId: string, dep
...
-  start(urlPath)
 
   const loadedDependents = await GET<PolicyDependent[]>(urlPath)
   dependentsByPolicyId.update((data) => {
     data[policyId] = loadedDependents
     return data
   })
 
-  stop(urlPath)
+  return loadedDependents
```
forevermatt, line 131: `<broken-record>`Same here. (Update store, don't return value.)`</broken-record>`

src/data/itemCategories.ts
```diff
@@ -23,13 +22,11 @@ export async function init() {
...
 
-  const catz = await GET<ItemCategory[]>('config/item-categories')
-
-  stop('itemCategories')
-
-  categories.set(catz)
+  categories.set(response)
   initialized.set(true)
+
+  return response
```
forevermatt, line 31: `<broken-record>`And here`</broken-record>`

src/data/items.ts
```diff
@@ -77,17 +76,16 @@ export const itemsByPolicyId = writable<{ [policyId: string]: PolicyItem[] }>({}
...
-  start(urlPath)
 
   const items = await GET<PolicyItem[]>(urlPath)
   itemsByPolicyId.update((data) => {
     data[policyId] = items
     return data
   })
 
-  stop(urlPath)
+  return items
```
forevermatt, line 88: `<broken-record />`

RussellChamp:
I feel like some non-void value should be returned by those API calls. how would you feel about returning status strings like 'done' or 'updated' or even returning the updated `writable` value that we just updated?

src/data/policies.ts
```diff
@@ -61,21 +60,19 @@ export async function updatePolicy(id: string, policyData) {
...
 
-  const plcs = await GET<Policy[]>('policies')
-
-  policies.set(plcs)
-
-  stop('loadPolicies')
+  policies.set(response)
   initialized.set(true)
+
+  return response
```
forevermatt, line 75: You guessed it...

src/data/policy-members.ts
```diff
@@ -21,15 +20,14 @@ export const membersByPolicyId = writable<{ [policyId: string]: PolicyMember[] }
...
-  start(urlPath)
 
   const policyMembers = await GET<PolicyMember[]>(urlPath)
   membersByPolicyId.update((data) => {
     data[policyId] = policyMembers
     return data
   })
 
-  stop(urlPath)
+  return policyMembers
```
forevermatt, line 32: Maybe I should have just messaged you directly, rather than saying the same thing in PR comments 42 times... (sorry)

RussellChamp: i'm good at detecting patterns, too

forevermatt (changes requested):
Okay, this is looking good (despite all my comments). Thanks for fixing the need for all those start/stop calls. 😁

src/data/claims.ts
```diff
@@ -134,9 +134,8 @@ export function init(): void {
  * @export
  * @param {Object} item
  * @param {Object} claimData
- * @return {Object} -- The newly created Claim
  */
-export async function createClaim(item: PolicyItem, claimData: any): Promise<Claim> {
+export async function createClaim(item: PolicyItem, claimData: any): Promise<void> {
```
forevermatt, line 138: Sorry if I misled you on this one, but we _do_ actually use the returned claim from `createClaim()`, because the code calling this has to immediately use that API-generated Claim ID to call `createClaimItem()`. So this one will still need to return the new claim.

forevermatt (approved):

PR #155, opened 2021-09-23
Claim actions for Steward at "Review1" status

forevermatt:
### Added
- Add steward action buttons/input for "Review1" claim status

### Changed
- Move definition of which claim statuses are editable into data/claims.ts

### Fixed
- Let the user (re)submit a claim with a status of "Revision"

### DIFF
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -1,29 +1,67 @@
 <script lang="ts">
 import user from '../authn/user'
 import { Claim, ClaimStatus, editableStatuses } from '../data/claims'
-import { Button } from '@silintl/ui-components'
+import { Button, TextField } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
+import Description from './Description.svelte'
 
 export let claim = {} as Claim
 
 const dispatch = createEventDispatcher()
 
+let message = ''
 let status: ClaimStatus
 $: status = claim.status
 
 let isEditable = false
 $: isEditable = editableStatuses.includes(status)
 
-const onEdit = () => dispatch('edit')
-const onSubmit = () => dispatch('submit')
+const on = (eventType) => () => dispatch(eventType)
 </script>
 
-{#if $user.app_role === 'User'}
+<style>
+.container {
+  display: grid;
+  grid-template-columns: 1fr auto;
+  grid-template-rows: 1fr auto;
+}
+.text-input {
+  grid-column: 1 / span 2;
+  grid-row-start: 1;
+}
+.left-buttons {
+  grid-column-start: 1;
+  grid-row-start: 2;
+}
+.right-buttons {
+  grid-column-start: 2;
+  grid-row-start: 2;
+  text-align: right;
+}
+</style>
+
+{#if $user.app_role === 'Steward'}
+  {#if claim.status === 'Review1'}
+    <div class="container">
+      <div class="text-input">
+        <TextField class="w-100" label="Send a message" bind:value={message} />
+        <Description>A message is required to deny or ask for changes.</Description>
+      </div>
+      <div class="left-buttons">
+        <Button on:click={on('deny')} disabled={!message} outlined>Deny</Button>
+      </div>
+      <div class="right-buttons">
+        <Button class="mx-1" on:click={on('ask-for-changes')} disabled={!message} raised>Ask for Changes</Button>
+        <Button on:click={on('approve')} raised>Approve</Button>
+      </div>
+    </div>
+  {/if}
+{:else if $user.app_role === 'User'}
   {#if isEditable}
-    <Button on:click={onEdit} outlined>Edit claim</Button>
+    <Button on:click={on('edit')} outlined>Edit claim</Button>
   {/if}
 
   {#if status === 'Draft' || status === 'Receipt'}
-    <Button raised on:click={onSubmit}>Submit claim</Button>
+    <Button raised on:click={on('submit')}>Submit claim</Button>
   {/if}
 {/if}
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -63,7 +63,7 @@ const onDeny = () => dispatch('deny', message)
     <Button on:click={on('edit')} outlined>Edit claim</Button>
   {/if}
 
-  {#if status === 'Draft' || status === 'Receipt'}
+  {#if status === 'Draft' || status === 'Receipt' || status === 'Revision'}
     <Button raised on:click={on('submit')}>Submit claim</Button>
   {/if}
 {/if}
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -18,6 +18,7 @@ $: isEditable = editableStatuses.includes(status)
 
 const on = (eventType) => () => dispatch(eventType)
 const onAskForChanges = () => dispatch('ask-for-changes', message)
+const onDeny = () => dispatch('deny', message)
 </script>
 
 <style>
@@ -49,7 +50,7 @@ const onAskForChanges = () => dispatch('ask-for-changes', message)
         <Description>A message is required to deny or ask for changes.</Description>
       </div>
       <div class="left-buttons">
-        <Button on:click={on('deny')} disabled={!message} outlined>Deny</Button>
+        <Button on:click={onDeny} disabled={!message} outlined>Deny</Button>
       </div>
       <div class="right-buttons">
         <Button class="mx-1" on:click={onAskForChanges} disabled={!message} raised>Ask for Changes</Button>
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+import user from '../authn/user'
 import { Claim, ClaimStatus, editableStatuses } from '../data/claims'
 import { Button } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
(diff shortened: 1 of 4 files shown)


src/components/ClaimActions.svelte
```diff
@@ -1,26 +1,69 @@
 <script lang="ts">
-import type { Claim } from '../data/claims'
-import { Button } from '@silintl/ui-components'
+import user from '../authn/user'
+import { Claim, ClaimStatus, editableStatuses } from '../data/claims'
+import { Button, TextField } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
+import Description from './Description.svelte'
```
hobbitronics, line 6: you can import this straight from components
```suggestion
import Description from './index
'
```

src/components/ClaimActions.svelte
```diff
@@ -1,26 +1,69 @@
...
+      <div class="text-input">
+        <TextField class="w-100" label="Send a message" bind:value={message} />
+        <Description>A message is required to deny or ask for changes.</Description>
+      </div>
+      <div class="left-buttons">
+        <Button on:click={onDeny} disabled={!message} outlined>Deny</Button>
+      </div>
+      <div class="right-buttons">
+        <Button class="mx-1" on:click={onAskForChanges} disabled={!message} raised>Ask for Changes</Button>
+        <Button on:click={on('approve')} raised>Approve</Button>
```
hobbitronics, line 57:
```suggestion
        <Button on:click={() => on('approve')} raised>Approve</Button>
```

src/components/ClaimActions.svelte
```diff
@@ -1,26 +1,69 @@
...
+    </div>
+  {/if}
+{:else if $user.app_role === 'User'}
+  {#if isEditable}
+    <Button on:click={on('edit')} outlined>Edit claim</Button>
+  {/if}
 
-{#if status === 'Draft' || status === 'Receipt'}
-  <Button raised on:click={onSubmitClaim}>Submit claim</Button>
+  {#if status === 'Draft' || status === 'Receipt' || status === 'Revision'}
```
hobbitronics, line 66: just an option to consider, not sure if it is easier to read, just a little shorter ```['Draft', 'Receipt', 'Revision'].includes(status)```

src/components/ClaimActions.svelte
```diff
@@ -1,26 +1,69 @@
...
+    <Button raised on:click={on('submit')}>Submit claim</Button>
```
hobbitronics, line 67:
```suggestion
    <Button raised on:click={() => on('submit')}>Submit claim</Button>
```

src/data/claims.ts
```diff
@@ -201,6 +210,68 @@ export async function updateClaim(claimId: string, newClaimData: any): Promise<v
...
+ * Approve a claim.
+ *
+ * @export
+ * @param {String} claimId
+ */
+export const approveClaim = async (claimId: string): Promise<Claim> => {
+  const approvedClaim = await CREATE<Claim>(`claims/${claimId}/approve`)
+
+  claims.update((claims) => {
+    const i = claims.findIndex((claim) => claim.id === claimId)
```
hobbitronics, line 223: looks like you have this code three times, maybe time to make a function?

hobbitronics:
```const updateClaimStore = (claims, updatedClaim)=> claims.update((claims) => {
    const i = claims.findIndex((claim) => claim.id === claimId)
    claims[i] = updatedClaim
    return claims
  })```

**Review by hobbitronics — CHANGES_REQUESTED** [review]
Looks good except a few possible bugs unless I am mistaken.

File: src/components/ClaimActions.svelte (hunk 3 above)
Comment (RussellChamp) [reply src/components/ClaimActions.svelte:66]: i've used the 'includes' pattern as well and it works pretty well (especially when your list of conditionals is dynamic)

File: src/data/claims.ts [hunk 6]
```diff
@@ -201,6 +210,68 @@ export async function updateClaim(claimId: string, newClaimData: any): Promise<v
...
+    claims[i] = approvedClaim
+    return claims
+  })
+
+  return approvedClaim
```
RussellChamp, line 228: i thought we were against returning values from API calls? should this instead be saved to a store like 'currentClaim' or 'currentClaimId' or something?

hobbitronics (approved):
I was mistaken about the bugs. I see what you did there, returning a a function as a value.

src/data/claims.ts
```diff
@@ -201,6 +210,68 @@ export async function updateClaim(claimId: string, newClaimData: any): Promise<v
   })
 }
```
hobbitronics, line 212:
```suggestion
/**
 * Update claims store.
 *
 * @param {Claim} changedClaim
 * @param {String} claimId
 */
const updateClaimsStore = (changedClaim: Claim, claimId: string) => {
  claims.update((claims) => {
    const i = claims.findIndex((claim) => claim.id === claimId)
    claims[i] = changedClaim
    return claims
  })
}
```

src/data/claims.ts
hobbitronics: `updateClaimsStore(approvedClaim, claimId)`

src/data/claims.ts
```diff
@@ -201,6 +210,68 @@ export async function updateClaim(claimId: string, newClaimData: any): Promise<v
...
+ * @param {String} claimId
+ */
+export const approveClaim = async (claimId: string): Promise<Claim> => {
+  const approvedClaim = await CREATE<Claim>(`claims/${claimId}/approve`)
+
+  claims.update((claims) => {
+    const i = claims.findIndex((claim) => claim.id === claimId)
+    claims[i] = approvedClaim
+    return claims
+  })
```
hobbitronics, line 226: you could call the suggested fucntion `  updateClaimsStore(approvedClaim, claimId)
`

src/data/claims.ts
```diff
@@ -201,6 +210,68 @@ export async function updateClaim(claimId: string, newClaimData: any): Promise<v
...
+  const requestBody: ClaimsRequestRevisionRequestBody = {
+    status_reason: reason,
+  }
+  const modifiedClaim = await CREATE<Claim>(`claims/${claimId}/revision`, requestBody)
+
+  claims.update((claims) => {
+    const i = claims.findIndex((claim) => claim.id === claimId)
+    claims[i] = modifiedClaim
+    return claims
+  })
```
hobbitronics, line 248: you could call the suggested function `  updateClaimsStore(modifiedClaim, claimId)
`

src/data/claims.ts
```diff
@@ -201,6 +210,68 @@ export async function updateClaim(claimId: string, newClaimData: any): Promise<v
...
+  const requestBody: DenyClaimRequestBody = {
+    status_reason: reason,
+  }
+  const deniedClaim = await CREATE<Claim>(`claims/${claimId}/deny`, requestBody)
+
+  claims.update((claims) => {
+    const i = claims.findIndex((claim) => claim.id === claimId)
+    claims[i] = deniedClaim
+    return claims
+  })
```
hobbitronics, line 270: you could call the suggested funciton `updateClaimsStore(deniedClaim, claimId)`

src/data/claims.ts
```diff
@@ -118,8 +118,17 @@ export type UpdateClaimItemRequestBody = {
...
+  status_reason: string
+}
+
+export type DenyClaimRequestBody = {
+  status_reason: string
+}
+
 export const claims = writable<Claim[]>([])
 export const initialized = writable<boolean>(false)
+export const editableStatuses: ClaimStatus[] = ['Draft', 'Review1', 'Review2', 'Review3', 'Revision', 'Receipt']
```
hobbitronics, line 132:
```suggestion
/**
 * Update claims store.
 *
 * @param {Claim} changedClaim
 * @param {String} claimId
 */
const updateClaimsStore = (changedClaim: Claim, claimId: string) => {
  claims.update((claims) => {
    const i = claims.findIndex((claim) => claim.id === claimId)
    claims[i] = changedClaim
    return claims
  })
}

```

hobbitronics: can be used on line 206 as well

src/components/ClaimActions.svelte
forevermatt: I'm leaning towards avoiding importing straight from components _within other components_ because of the circular-import warnings from Svelte. Do you all have any thoughts on that general practice?

src/components/ClaimActions.svelte
forevermatt: Agreed. That was just a minor bug-fix here, but switching to an `includes()` pattern is a good direction to go.

src/data/claims.ts
forevermatt: That crossed my mind when I was reviewing the changes when submitting the PR. I think we need to establish some convention for when to return them, and when not to.

So far…
1. We shouldn’t return them from API calls that load a list of things into a Svelte store.
   - Reason: We want UI code to use the store (auto-updating) rather than the return value (static).
2. We should (at least _sometimes_) return them from “create” API calls.
   - Reason: Sometimes we need the new item’s UUID right away, such as when creating a claim then immediately adding an item to that claim.
   - Whether we always return them from create calls… maybe so?
3. All other cases (updates, deletes, etc.)… maybe not?

What are your thoughts? (@RussellChamp, @hobbitronics, and/or @Schparky)

src/components/ClaimActions.svelte
forevermatt: I'll plan to make that change in a future PR

src/data/claims.ts
Schparky: 1. Agreed, do not return the data that's saved into a store
2. I think "always" is better, if only for consistency. Just a thought, maybe only the ID should be returned, requiring the caller to fetch the new data from the store.
3. Probably not, just make sure the store is correctly updated

forevermatt: @Schparky, that makes sense. Thanks. And your response to point 2 is an interesting one... I think I like that idea. 🙂

src/components/ClaimActions.svelte
forevermatt: For now, I'll continue importing components directly, when doing so within other components, to avoid circular imports. Feel free to bring this back up for further discussion if desired.

forevermatt:
Okay, I believe I've addressed all the concerns. @hobbitronics, do you want to take another look before I merge?

src/data/claims.ts
hobbitronics: I agree with Schparky

src/components/ClaimActions.svelte
hobbitronics: oh yeah I forgot about those warnings.

hobbitronics (approved):
I think you can use that updateClaimsStore function in one more spot (should be 4 times) on line 219 now.

PR #156, opened 2021-09-23
Remove coverage

hobbitronics:
Added a dialog and called the remove endpoint
Screenshot from 2021-09-24 12-09-55

### DIFF
--- a/src/customTypes/@silintl/ui-components.d.ts
+++ b/src/customTypes/@silintl/ui-components.d.ts
@@ -4,13 +4,6 @@ declare module '@silintl/ui-components' {
 
   export type actions = writable<any[]>
 
-  //Dialog.Alert Buttons
-  export type AlertButton = {
-    label: string
-    action: strin
-    class: string
-  }
-
   interface ButtonProps extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['div']> {
     disabled?: any
     outlined?: any
--- a/src/customTypes/@silintl/ui-components.d.ts
+++ b/src/customTypes/@silintl/ui-components.d.ts
@@ -4,6 +4,13 @@ declare module '@silintl/ui-components' {
 
   export type actions = writable<any[]>
 
+  //Dialog.Alert Buttons
+  export type AlertButton = {
+    label: string
+    action: strin
+    class: string
+  }
+
   interface ButtonProps extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['div']> {
     disabled?: any
     outlined?: any
@@ -77,6 +84,7 @@ declare module '@silintl/ui-components' {
       open?: boolean
       title?: string
       defaultAction?: string
+      buttons?: AlertButton[]
     }
     export class Alert extends SvelteComponentTyped<AlertProps> {}
 
--- a/src/customTypes/@silintl/ui-components.d.ts
+++ b/src/customTypes/@silintl/ui-components.d.ts
@@ -73,6 +73,11 @@ declare module '@silintl/ui-components' {
   }
 
   export namespace Dialog {
+    export type AlertButton = {
+      label: string
+      action: string
+      class: string
+    }
     interface AlertProps extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap['div']> {
       open?: boolean
       title?: string
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -64,8 +64,7 @@ async function customFetch<T>(method: FetchMethod, uri: string, body: any = unde
   } finally {
     stop(url)
   }
-
-  const results = await response.json()
+  const results = response.status === 204 /* No Content */ ? {} : await response.json()
 
   // reminder: fetch does not throw exceptions for non-200 responses (https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch)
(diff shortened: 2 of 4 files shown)


RussellChamp:
implement the fix on the json parsing we spoke of then this lgtm

hobbitronics:
> implement the fix on the json parsing we spoke of then this lgtm

worked like a charm

src/customTypes/@silintl/ui-components.d.ts
```diff
@@ -4,6 +4,13 @@ declare module '@silintl/ui-components' {
 
   export type actions = writable<any[]>
 
+  //Dialog.Alert Buttons
```
forevermatt, line 7: If this is a Dialog.Alert, should it be in a `Dialog` namespace? (see the Datatable example below)

src/pages/items/[itemId].svelte
```diff
@@ -4,22 +4,27 @@ import { Banner, Breadcrumb, ItemBanner, Row } from '../../components'
...
-import { Button, Page } from '@silintl/ui-components'
+import { Button, Page, Dialog, AlertButton } from '@silintl/ui-components'
 
-export let itemId
+export let itemId: string
 
 const now = Date.now()
+const buttons: AlertButton[] = [
+  { label: 'Yes, I’m Sure', action: 'remove', class: 'error-button' },
+  { label: 'cancel', action: 'cancel', class: 'mdc-dialog__button' },
```
Comment (forevermatt) [inline src/pages/items/[itemId].svelte:18]: It seems odd that we would need to pass in an `mdc-dialog__buton` class. That may be perfectly correct, but somehow it seems like an implementation detail that would have been handled within the Dialog.Alert component.

src/pages/items/[itemId].svelte
```diff
@@ -59,7 +74,15 @@ const goToNewClaim = () => {
...
         <Breadcrumb links={breadcrumbLinks} />
         <div>
-          <Button class="remove-button mx-5px" url={`/items/${itemId}/delete`}>Remove</Button>
+          <Button class="remove-button mx-5px" on:click={() => (open = true)}>Remove</Button>
+          <Dialog.Alert
+            {open}
+            {buttons}
+            defaultAction="cancel"
+            title="Remove Coverage"
+            on:chosen={(e) => handleDialog(e.detail)}
```
Comment (forevermatt) [inline src/pages/items/[itemId].svelte:83]: Rather than calling the same method in response to both different types of events that this component can dispatch, and then having to add an `if` in that method to sort out the difference, this seems like a situation where they could each call their own method, that just does what is applicable in response to that type of event.

This certainly isn't a blocking concern, merely a suggestion of what seems like a cleaner design. If you would like me to suggest actual code changes, I'll be happy to do so.

src/data/index.ts
```diff
@@ -64,8 +64,7 @@ async function customFetch<T>(method: FetchMethod, uri: string, body: any = unde
   } finally {
     stop(url)
   }
-
-  const results = await response.json()
+  const results = response.status === 204 /* No Content */ ? {} : await response.json()
```
forevermatt, line 67: Nice solution! 👍

src/pages/items/[itemId].svelte
Comment (hobbitronics) [reply src/pages/items/[itemId].svelte:18]: it is actually the default value. I had to give non defaults for the other button properties so also include this. We really want a solid background color but without changing the Alert component I cant just pass the button the raised prop. Any ideas on a better solution?

src/pages/items/[itemId].svelte
Comment (hobbitronics) [reply src/pages/items/[itemId].svelte:83]: The `if` needs to be there because one of the choices `cancel` from `event.detail` should not run that code. Both handler functions need to set `menuOpen = false` so I think this is about the most efficient way I can think of doing things. I basically copied and pasted this code from TF.

src/customTypes/@silintl/ui-components.d.ts
hobbitronics: sorry, what example?

src/customTypes/@silintl/ui-components.d.ts
```diff
@@ -4,6 +4,13 @@ declare module '@silintl/ui-components' {
...
+  export type AlertButton = {
+    label: string
+    action: strin
+    class: string
+  }
+
```
forevermatt, line 13: (Moving down into the "Dialog" namespace)
```suggestion
```

forevermatt: Hmm... GitHub isn't letting me start a suggestion on the `export namespace Dialog {` line because it's too far away from lines changed in this PR.

src/pages/items/[itemId].svelte
Comment (forevermatt) [reply src/pages/items/[itemId].svelte:83]: Given the constraints of Dialog.Alert's events (all buttons dispatch the same type of event, with details about which button was clicked), I think your solution is fine.

Ideally, Dialog.Alert could submit an event type of whatever the button's action was, to allow for explicitly handling the separate buttons separately. That's a change for another day, though. 🙂

forevermatt (approved):

src/pages/items/[itemId].svelte
Comment (forevermatt) [reply src/pages/items/[itemId].svelte:18]: My only thought in that regard would be to modify Dialog.Alert (in that separate library) and put the `mdc-dialog__button` class on each button, hard-coded, then also add the button.class (if one was provided). I don't know the full effects of that, though

PR #157, opened 2021-09-24
fix claim form to load money values

RussellChamp:
* update save claim item to persist to store
* add more typings
* fix some typings

Some of the components rely on a bit of a "trinary" system where they manually set values to 'undefined'. this is a bit of a code smell to me but beyond the current scope.

### DIFF
--- a/src/components/MoneyInput.svelte
+++ b/src/components/MoneyInput.svelte
@@ -5,9 +5,9 @@ import { generateRandomID } from '@silintl/ui-components/random'
 import { onMount } from 'svelte'
 
 export let label = ''
-export let value = null
+export let value: number | undefined
 export let placeholder = ''
-export let maxlength = undefined
+export let maxlength: number = 524288 /* default */
 export let autofocus = false
 export let disabled = false
 
--- a/src/components/forms/ClaimForm.svelte
+++ b/src/components/forms/ClaimForm.svelte
@@ -12,7 +12,7 @@ import {
 } from '../../business-rules/claim-payout-amount'
 import { ConvertCurrencyLink, Description, RadioOptions, DateInput, MoneyInput } from '../../components'
 import { claimIncidentTypes, loadClaimIncidentTypes } from '../../data/claim-incident-types'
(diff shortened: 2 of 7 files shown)


src/components/MoneyInput.svelte
```diff
@@ -5,9 +5,9 @@ import { generateRandomID } from '@silintl/ui-components/random'
 import { onMount } from 'svelte'
 
 export let label = ''
-export let value = null
+export let value: number | undefined
 export let placeholder = ''
-export let maxlength = undefined
+export let maxlength: number = 524288 /* default */
```
hobbitronics, line 10: I'm curious where you got this number.

RussellChamp: from w3cSchools https://www.w3schools.com/tags/att_input_maxlength.asp

src/components/forms/ClaimForm.svelte
```diff
@@ -143,8 +150,14 @@ const setInitialValues = (claim, claimItem) => {
   if (claimItem.is_repairable !== undefined) {
     repairableSelection = claimItem.is_repairable ? 'repairable' : 'not_repairable'
   }
+
   payoutOption = claimItem.payout_option || payoutOption
+
+  repairEstimateUSD = claimItem.repair_estimate / 100
+  replaceEstimateUSD = claimItem.replace_estimate / 100
+  fairMarketValueUSD = claimItem.fmv / 100
```
hobbitronics, line 158: could we use formatMoney from money.ts here?

RussellChamp: formatMoney returns a string with a '$' symbol. we just want a number here

RussellChamp: but we could add a 'convertToDollars' method if you'd like

hobbitronics: since we should be dealing with round figures here I don't see any issue really.

hobbitronics (approved):
Looks good

forevermatt:
@RussellChamp - Did you mean to merge this, rather than close it?

PR #158, opened 2021-09-24
fix the auth redirect issue

RussellChamp:
a change in error type introduced a regression in the authentication redirect code when a 401 is hit. better type data was added. ;)

mea culpa! mea culpa! I break it, I fix it.


hobbitronics (approved):
lgtm

PR #160, opened 2021-09-27
Show accountable person on home page

hobbitronics:
### Fixed
- accountable person was not showing


forevermatt (approved):
At some point I expect we'll want to extract out this Datatable usage to its own component (maybe a `PolicyItems` component?) to move some of this complexity out of the home page's code, but these changes look fine to me.

PR #165, opened 2021-09-28
accountablePerson in home, item and itemForm

hobbitronics:
### Added
- accountablePersons.ts with helper functions
### Fixed
- accountable person not showing in item view and home item table
- accountable person not loading in edit/item

I was playing around with different ways of doing the same thing here so we may want to pick one and get rid of any unnecessary helpers

### DIFF
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -144,7 +144,6 @@ const setInitialValues = (item: PolicyItem) => {
     <Description>Required for mobile items.</Description>
   </p>
   <p>
-    <!-- TODO: Set the initial value here (like on Category) once the API is providing it. -->
     <Select
       label="Accountable person"
       on:change={onAccountablePersonChange}
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { ConvertCurrencyLink, Description, MoneyInput } from '../../components'
+import { getAccountablePerson } from '../../data/accountablePersons'
 import { dependentsByPolicyId, loadDependents } from '../../data/dependents'
 import type { ItemCoverageStatus, PolicyItem } from '../../data/items'
 import { categories, init, initialized as catItemsInitialized } from '../../data/itemCategories'
@@ -8,29 +9,29 @@ import { Button, Form, Select, TextArea, TextField } from '@silintl/ui-component
 import { createEventDispatcher } from 'svelte'
 
 export let item = {} as PolicyItem
-export let policyId: string = undefined
+export let policyId: any = undefined
 
 const dispatch = createEventDispatcher<{ submit: any; 'save-for-later': any }>()
 
 // Set default values.
-let accountablePersonId = ''
-let categoryId = ''
-let country = ''
-let marketValueUSD = ''
-let coverageStartDate = ''
+let accountablePersonId: string = ''
+let categoryId: string = ''
+let country: string = ''
+let marketValueUSD: string = ''
+let coverageStartDate: string = ''
 let coverageStatus: ItemCoverageStatus
-let itemDescription = ''
+let itemDescription: string = ''
 let inStorage = false
-let make = ''
-let model = ''
-let shortName = ''
-let purchaseDate = ''
-let uniqueIdentifier = ''
+let make: string = ''
+let model: string = ''
+let shortName: string = ''
+let purchaseDate: string = ''
+let uniqueIdentifier: string = ''
 
 // Set initial values based on the provided item data.
 $: setInitialValues(item)
 
-let initialCategoryId = undefined
+let initialCategoryId: any = undefined
 let today = new Date()
 
 $: dependents = $dependentsByPolicyId[policyId] || []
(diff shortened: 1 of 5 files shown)


src/components/forms/ItemForm.svelte
```diff
@@ -1,61 +1,57 @@
 <script lang="ts">
 import { ConvertCurrencyLink, Description, MoneyInput } from '../../components'
-import { dependentsByPolicyId, loadDependents } from '../../data/dependents'
+import { getDependentOptions, getPolicyMemberOptions } from '../../data/accountablePersons'
+import { dependentsByPolicyId, loadDependents, PolicyDependent } from '../../data/dependents'
```
forevermatt, line 4: It has sometimes caused problems to have components reach out and load data, as that can result in repeated-but-identical calls for data as a page loads data and a component on that page loads data.

As a general practice, if we can limit "load ___" API calls to only happen with page code (not component code), I think that will help us.

src/components/forms/ItemForm.svelte
```diff
@@ -1,61 +1,57 @@
...
+let initialAccountablePersonId = undefined
+let initialCategoryId: any = undefined
 let today = new Date()
 
 $: dependents = $dependentsByPolicyId[policyId] || []
-$: dependentOptions = dependents.map((dependent) => ({
-  id: dependent.id,
-  name: dependent.name,
-}))
+$: dependentOptions = getDependentOptions(dependents)
```
forevermatt, line 39: Nic extractions (here and on `getPolicyMemberOptions()`) 👍

src/components/forms/ItemForm.svelte
```diff
@@ -140,7 +144,13 @@ const setInitialValues = (item: PolicyItem) => {
   </p>
   <p>
     <!-- TODO: Set the initial value here (like on Category) once the API is providing it. -->
```
forevermatt, line 146: I believe this `TODO` can be deleted now

src/data/accountablePersons.ts
```diff
@@ -0,0 +1,41 @@
...
+export type AccountablePerson = {
+  id: string
+  name: string
+}
+
+export const getPolicyMemberOptions = (policyMembers: any[]): AccountablePerson[] =>
+  policyMembers?.map((policyMember) => ({
+    id: policyMember.id,
+    name: policyMember.first_name + ' ' + policyMember.last_name,
+  }))
```
forevermatt, line 14: Something's odd about the indentation here... I think we're missing the opening `{` at the end of line 10 and a closing `}` after line 14 (which necessitates a `return` when fixed):
```suggestion
export const getPolicyMemberOptions = (policyMembers: any[]): AccountablePerson[] => {
  return policyMembers?.map((policyMember) => ({
    id: policyMember.id,
    name: policyMember.first_name + ' ' + policyMember.last_name,
  }))
}
```

src/data/accountablePersons.ts
```diff
@@ -0,0 +1,41 @@
...
+import { dependentsByPolicyId } from './dependents'
+import { membersByPolicyId } from './policy-members'
+
+export type AccountablePerson = {
+  id: string
+  name: string
+}
+
+export const getPolicyMemberOptions = (policyMembers: any[]): AccountablePerson[] =>
+  policyMembers?.map((policyMember) => ({
```
forevermatt, line 11: We may have an issue of this not returning an array if `policyMembers` is undefined... is there a need for the `?.` syntax here?

src/data/accountablePersons.ts
```diff
@@ -0,0 +1,41 @@
...
+  dependents?.map((dependent) => ({
+    id: dependent.id,
+    name: dependent.name,
+  }))
+
+export const getAccountablePersonsByPolicyId = (id: string): AccountablePerson[] => {
+  let dependents_value: any
+  let members_value: any
+
+  const unsubscribe_dependents = dependentsByPolicyId.subscribe((value) => (dependents_value = value[id] || []))
```
forevermatt, line 26: If all you're after here is to get the current value, there is a `get()` method exported by svelte for getting the current value of a store.

However, I'd be interested in talking through options here... something isn't sitting right for me about this approach.

src/data/items.ts
```diff
@@ -13,7 +13,9 @@ export type RiskCategory = {
 }
 
 export type PolicyItem = {
-  accountable_person: string
+  accountable_person_id?: string
+  accountable_dependent_id?: string
+  accountable_user_id?: string
```
forevermatt, line 18: Also worth discussing, as I don't think (long term) we want to keep receiving all three values, just either the 1st one or the 2nd-and-3rd.

src/data/accountablePersons.ts
hobbitronics: this was probably added when I was debugging and had lots of type errors muddying things up. Everything seems fine without them.

src/data/accountablePersons.ts
hobbitronics: this is javascript trying to be python

src/components/forms/ItemForm.svelte
hobbitronics: since these calls were already there and I don't wan't to make this too much longer should we made a card for moving these calls to the page level?

forevermatt: Yes, let's do that. I also don't want to add more to this PR. 👍

forevermatt: Okay, I've created a card for that

src/data/items.ts
hobbitronics: it is reducing the number of error messages for now.

src/components/forms/ItemForm.svelte
```diff
@@ -1,61 +1,58 @@
...
+let model: string = ''
+let shortName: string = ''
+let purchaseDate: string = ''
+let uniqueIdentifier: string = ''
 
 // Set initial values based on the provided item data.
 $: setInitialValues(item)
 
-let initialCategoryId = undefined
+let accountablePersons = [] as (PolicyMember | PolicyDependent)[]
```
forevermatt, line 34: I think I was wrong about this. The only way we're using `accountablePersons` is as options to a Select, so I think this would be `let accountablePersons: AccountablePersonOptions[]`, like you have in `items/[itemId].svelte`

forevermatt: That will probably require another import in this file, just FYI.

forevermatt (approved):

PR #167, opened 2021-09-29
Re-load the user data (after a delay) if we didn't get their policy_id

forevermatt:
### Fixed
- Ensure "Accountable Persons" list has entries, by re-retrieving the user data if it didn't come back with a policy_id


hobbitronics (approved):
Gotta love those 5 second timeouts!

PR #168, opened 2021-09-29
Claim of status: "Draft" redirects to /claimID/edit, "save for later" button redirects to /home

hobbitronics:
### Added
- ClaimBanner to /[claimID]/edit.svelte
- `$: claimStatus === 'Draft' && editClaim()` to [ClaimID].svelte
- `$$props.class` to ClaimBanner.svelte
### Changed
- onSaveForLater to redirect to /home
- canProvideEvidenceNow to only include 'Review1'
draftForm

### DIFF
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -55,6 +55,7 @@ $: item = items.find((itm) => itm.id === claimItem.item_id) || ({} as PolicyItem
 
 $: incidentDate = formatDate(claim.incident_date)
 $: claimStatus = (claim.status || '') as ClaimStatus
+$: claimStatus === 'Draft' && editClaim()
 $: payoutOption = claimItem.payout_option as PayoutOption
 $: needsRepairReceipt = needsReceipt && payoutOption === 'Repair'
 $: needsReplaceReceipt = needsReceipt && payoutOption === 'Replacement'
--- a/src/pages/claims/[claimId].svelte
+++ b/src/pages/claims/[claimId].svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+import user from '../../authn/user'
 import { determineMaxPayout, isEvidenceNeeded } from '../../business-rules/claim-payout-amount'
 import {
   Banner,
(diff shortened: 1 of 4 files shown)


merchako:
I'd like the list item to be more prominent in the list. 
- How are you determining the sort order of items in the list. Could we go by recently changed, so the draft would wind up at the top?
- Could you bold and color (blue) the "Draft" status cell for customers?
- Put focus on the draft item after the redirect

Also, there's too much shifting during page load. Could we do something about that? (certainly a different ticket)

merchako (approved):

src/pages/claims/[claimId].svelte
```diff
@@ -55,6 +55,7 @@ $: item = items.find((itm) => itm.id === claimItem.item_id) || ({} as PolicyItem
 
 $: incidentDate = formatDate(claim.incident_date)
 $: claimStatus = (claim.status || '') as ClaimStatus
+$: claimStatus === 'Draft' && editClaim()
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:58]: Eventually, this will need to take into account whether it's the Customer viewing it or it's a Steward / Signator viewing it. Maybe if the claim's user_id matches $user.id and it's in Draft status?

src/pages/claims/[claimId]/edit.svelte
```diff
@@ -69,6 +69,7 @@ const onSubmit = async (event: CustomEvent) => {
   <!-- @todo Handle situations where the user isn't allowed to edit this claim. -->
   <Page>
     <Breadcrumb links={breadcrumbLinks} />
+    <ClaimBanner claimStatus={'Draft'} class="my-2" />
```
Comment (forevermatt) [inline src/pages/claims/[claimId]/edit.svelte:72]: Don't we want this to show the claim's actual status, not just always show "Draft"?

Comment (hobbitronics) [reply src/pages/claims/[claimId]/edit.svelte:72]: we could get the status from the claim but in my understanding a claim being edited will always have a Draft status.

src/pages/claims/[claimId].svelte
```diff
@@ -55,7 +56,7 @@ $: item = items.find((itm) => itm.id === claimItem.item_id) || ({} as PolicyItem
 
 $: incidentDate = formatDate(claim.incident_date)
 $: claimStatus = (claim.status || '') as ClaimStatus
-$: claimStatus === 'Draft' && editClaim()
+$: claimStatus === 'Draft' && $user.app_role == 'User' && editClaim()
```
Comment (forevermatt) [inline src/pages/claims/[claimId].svelte:59]:
```suggestion
$: claimStatus === 'Draft' && $user.app_role === 'User' && editClaim()
```

forevermatt (approved):

PR #170, opened 2021-09-30
Add landing page for "invitation" links

forevermatt:
### Added
- Add "/invite/[uuid].svelte" page
- Do a slightly different login call for that page

### Changed
- Only call `loadUser()` for non-public pages

### DIFF
--- a/src/components/App.svelte
+++ b/src/components/App.svelte
@@ -5,8 +5,7 @@ import type { CustomError } from '../error'
 import './mdc/_index.scss'
 import t from '../i18n'
 import { parse, stringify } from 'qs'
-import { Router } from '@roxi/routify'
-import { onMount } from 'svelte'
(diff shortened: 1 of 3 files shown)


src/components/App.svelte
```diff
@@ -19,12 +18,21 @@ const queryHandler = {
   parse: (params: string) => parse(params, { ignoreQueryPrefix: true }),
   stringify,
 }
-onMount(() => {
+
+const publicRoutes = ['/invite/:uuid']
+
+const authenticateUser = async () => {
```
hobbitronics, line 24: is anything returned here?

hobbitronics: nevermind

hobbitronics (approved):
I'm not totally clear on how the afterPageLoad stuff works but everything else looks good.

PR #171, opened 2021-09-30
Extract `UserHomePage` component

forevermatt:
### Changed
- Extract home page content to a new UserHomePage, to prepare for showing different content to Stewards and Signators

---

Since this moved a lot of content to a different file, the diff won't tell you much, so I kept changes to an absolute minimum. You can look at the individual commits to see what actually changed, but the commit messages should be a pretty good summary.


hobbitronics (approved):
looks good

PR #173, opened 2021-09-30
Increase button and fallback font weight

hobbitronics:
### Added
- --mdc-typography-button-font-weight to _theme.scss
- .mdc-button rule to _theme.scss
### Changed
- _fallback.svelte content font-weight to use --mdc-typography-button-font-weight t
Screenshot from 2021-09-30 16-19-04
o

### DIFF
--- a/src/pages/_fallback.svelte
+++ b/src/pages/_fallback.svelte
@@ -1,7 +1,5 @@
 <script lang="ts">
 import { notFound } from '../analytics'
-import { Row } from '../components'
-import { Page } from '@silintl/ui-components'
 import { onMount } from 'svelte'
 
 onMount(notFound)
@@ -11,15 +9,6 @@ onMount(notFound)
 p {
   text-align: center;
 }
-.typography-button-font-weight {
-  font-weight: var(--mdc-typography-button-font-weight);
-}
 </style>
 
-<Page>
-  <div class="typography-button-font-weight">
-    <Row>
-      <p>These aren't the droids you're looking for... 🤖</p>
-    </Row>
-  </div>
-</Page>
+<p>These aren't the droids you're looking for... 🤖</p>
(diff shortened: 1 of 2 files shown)


src/pages/_fallback.svelte
```diff
@@ -9,6 +11,15 @@ onMount(notFound)
...
+}
 </style>
 
-<p>These aren't the droids you're looking for... 🤖</p>
+<Page>
+  <div class="typography-button-font-weight">
+    <Row>
+      <p>These aren't the droids you're looking for... 🤖</p>
+    </Row>
+  </div>
```
forevermatt, line 24: Was this just to demonstrate the effect? (I'm a little confused, since this isn't a button... 🙂 )

merchako (approved):
Just what I asked for. Thanks!

It looks like this commit changed a bunch of the hex colors from uppercase to lowercase. Is that intentional? There's nothing wrong with it, but it does give the impression that the PR is much bigger than it really is.

src/pages/_fallback.svelte
hobbitronics: this was another request of Alex's to have just this page font-weight: 700. I wasn't sure about the best way to target one page but this worked. I could have just added it to the `<p>` I suppose.

hobbitronics:
> Just what I asked for. Thanks!
> 
> It looks like this commit changed a bunch of the hex colors from uppercase to lowercase. Is that intentional? There's nothing wrong with it, but it does give the impression that the PR is much bigger than it really is.

That is just prettier reformatting on save. Eventually we will convert these to HSLA I think.

src/pages/_fallback.svelte
forevermatt: It just seems like a mis-application of the class. Isn't the class intended for buttons?

merchako: Huh. I don’t remember asking for this. Maybe it was something else?

I do agree with Matt that we shouldn’t use a button class for this.

hobbitronics: looks like I should get rid of this change

forevermatt (approved):

PR #174, opened 2021-10-01
fix overflowing fields in claimId & itemId

hobbitronics:
### Added
- break-word class 
- apply to claimId & itemId fields to fix overflowing

### DIFF
--- a/src/pages/items/[itemId].svelte
+++ b/src/pages/items/[itemId].svelte
@@ -113,7 +113,7 @@ const handleDialog = async (choice: string) => {
     </Row>
 
     <Row cols="3">
-      <h2>{item.name || ''}</h2>
+      <h2 class="break-word">{item.name || ''}</h2>
       <b>Covered value</b>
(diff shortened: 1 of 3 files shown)


src/pages/items/[itemId].svelte
```diff
@@ -113,7 +113,7 @@ const handleDialog = async (choice: string) => {
     </Row>
 
     <Row cols="3">
-      <h2>{item.name || ''}</h2>
+      <h2 class="break-word">{item.name || ''}</h2>
```
Comment (forevermatt) [inline src/pages/items/[itemId].svelte:116]: Any ideas why this didn't automatically wrap the text, as the typical flow layout normally does?

I wonder if there's a tweak we could make to our page structure / grid / something that would avoid the need for this class to be added to multiple elements.

Comment (hobbitronics) [reply src/pages/items/[itemId].svelte:116]: This only happens when there are very long strings with no spaces, otherwise it does wrap just fine.

forevermatt (approved):

PR #175, opened 2021-10-04
Move customer home page

forevermatt:
### Changed
- Move normal User (aka "Customer") home page content to `/customer/home/[policyId]`
- Update redirects to get customers to new home page route

### Fixed
- Remove breadcrumbs from customer home page


hobbitronics (approved):
looks good at least to my Morning brain.

PR #177, opened 2021-10-04
Move customer claims pages

forevermatt:
### Changed
- Move customer-focused "claims" pages under `/customer/`


src/components/ClaimCards.svelte
```diff
@@ -10,7 +10,7 @@ export let claims: Claim[]
 export let items: PolicyItem[]
 export let accountablePersons = [] as AccountablePersonOptions[]
 
-const onGotoClaim = (event) => $goto(`/claims/${event.detail}`)
+const onGotoClaim = (event) => $goto(`/customer/claims/${event.detail}`)
```
hobbitronics, line 13: Perhaps a we should be passing in `user.app_role` as a prop to construct the appropraite url. This could all be TODO for now though.

hobbitronics (approved):
Approved with a suggested Change/TODO

src/components/ClaimCards.svelte
hobbitronics: I suppose it would be `$isSteward`.

forevermatt: I'd like to actually change this component to dispatch an event, and let the actual page handle it. I just didn't to include that in this PR, since I try to keep move-a-page PRs as short as I reasonably can.

forevermatt: TODO added

PR #180, opened 2021-10-04
Update UI-components and some other deps

hobbitronics:
Updated deps to minor versions
removed Menu from components and imported from UI-components


src/customTypes/@silintl/ui-components.d.ts
```diff
@@ -130,6 +130,17 @@ declare module '@silintl/ui-components' {
   }
   export class List extends SvelteComponentTyped<ListProps> {}
 
+  type MenuItem = {
+    icon: string
+    label: string
+    url: string
+  }
```
forevermatt, line 137: I think there's also a `hide` property, if you want to include that.

forevermatt: At least, there was in our local Menu... I don't see it in the ui-components library.

forevermatt: Ah, there it is. It's just in the Drawer, so never mind:
https://github.com/silinternational/ui-components/blob/develop/components/mdc/Drawer/Drawer.svelte#L125

forevermatt (approved):

PR #184, opened 2021-10-06
Validate item form and email

hobbitronics:
### Added
- /validation/assertions.ts, assertEmailAddress() and assertHas()
- validate() to ItemForm
- location to AccountablePersonOptions
- accountablePerson to get country from after an accountablePerson is selected

### Changed
- updateCustomEmail to use assertEmailAddress()


src/validation/assertions.ts
```diff
@@ -0,0 +1,15 @@
+import { throwError } from '../error'
+
+export function assertHas(value: any, errorMessage: string): void {
+  if (!value) {
+    throwError(errorMessage)
+  }
+}
+
+export function assertEmailAddress(email: string, errorMessage: string): void {
+  const simpleEmailRegex = /\S+@\S+\.\S+/
```
RussellChamp, line 10: much consternation is had over whether or not you should use regex to validate email addresses. eg https://stackoverflow.com/questions/48055431/can-it-cause-harm-to-validate-email-addresses-with-a-regex (short story is that you shouldn't.) yours should work for 99.9% of cases, though

hobbitronics: will put a TODO here

PR #185, opened 2021-10-07
add validation to DependentForm

hobbitronics:
- Validated the required field and give appropriate message
- Disabled 'Spouse' as an option if one already exists when adding a new dependent (not when editing, may be a discussion here)


src/components/forms/DependentForm.svelte
```diff
@@ -25,6 +27,24 @@ let formData = {
...
+$: alreadyHasSpouse && (relationshipOptions[0].disabled = true)
+
+const validateAndSanitize = () => {
+  assertHas(formData.name, 'Please specify a name')
+  assertHas(formData.location, 'Please specify a location')
+  assertHas(formData.relationship, 'Please select "Spouse" or "Child"')
+  if (formData.relationship === 'Child') {
+    assertHas(formData.childBirthYear, "Please specify your child's birthyear")
+  } else {
+    delete formData.childBirthYear
```
RussellChamp, line 43: i'm a little concerned about doing data manipulation in a validation method. should this be done here or during the `onSubmit` event in the case that a user toggles back and forth between 'Spouse' and 'Child'?

src/components/forms/ItemForm.svelte
```diff
@@ -43,7 +44,9 @@ $: policyMembers = $membersByPolicyId[policyId] || []
 $: policyMemberOptions = getPolicyMemberOptions(policyMembers)
 
 $: accountablePersons = [...policyMemberOptions, ...dependentOptions]
+$: accountablePerson = accountablePersons.find( person => person.id === (accountablePersonId || initialAccountablePersonId)) as AccountablePersonOptions
```
RussellChamp, line 47: The `as AccountablePersonOptions` here should be unnecessary as `accountablePersons` is already set to `AccountablePersonOptions[]` and the types should "flow" correctly. I think i saw you use this pattern elsewhere as well.

src/components/forms/ItemForm.svelte
```diff
@@ -86,12 +89,33 @@ const onCategorySelectPopulated = () => {
   }
 }
 
+const validateOnSave = (formData: any) => {
```
RussellChamp, line 92: same question as last PR about different validation methods

PR #190, opened 2021-10-11
Feature/add routes improve imports

RussellChamp:
* fix the broken "New Claim" link
* add centralized routes file
* replaced hardcoded routes with references to routes file
* add components, data, and helpers to includePaths
* update import references for components, data, and helpers
* improve data types and resolve ts warnings


hobbitronics (approved):
Looks good.

PR #201, opened 2021-10-14
Add "Recent Activity" table to Steward home page

forevermatt:
### Added
- Add "Recent Activity" table to Steward home page

### DIFF
--- a/src/components/RecentActivityTable.svelte
+++ b/src/components/RecentActivityTable.svelte
@@ -1,0 +1,83 @@
+<script lang="ts">
+import type { Claim } from 'data/claims'
+import { formatMoney } from 'helpers/money'
+import { Datatable } from '@silintl/ui-components'
+
+export let dependents = []
+export let loading = false
+export let policyMembers = []
+export let recentChanges = []
+
+const HeaderRow = Datatable.Header
+const HeaderItem = Datatable.Header.Item
+const TableBody = Datatable.Data
+const DataRow = Datatable.Data.Row
+const RowItem = Datatable.Data.Row.Item
+
+$: people = { policyMembers, dependents }
+
+const getClaimItemName = (claim: Claim): string => {
+  const claimItem = claim.claim_items[0]
+  return claimItem?.item?.name || ''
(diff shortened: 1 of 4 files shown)


src/pages/steward/home.svelte
```diff
@@ -1,15 +1,19 @@
 <script lang="ts">
-import { ClaimCards, Row } from 'components'
+import { ClaimCards, RecentActivityTable, Row } from 'components'
+import { loading } from 'components/progress'
```
hobbitronics, line 3: in other places I've notices some wonky loading when using $loading rather than isLoadingById('steward/recent'). I pulled the code an didn't notice a difference, but just something to be aware of if multiple things are loading you can get some jumpiness.

hobbitronics (approved):
Looks like you are still waiting on a value for Recent Activity. I also noticed some misalignment on my machine but otherwise looks good.

src/components/RecentActivityTable.svelte
```diff
@@ -0,0 +1,83 @@
...
+  return formatMoney(claimItem?.item?.annual_premium)
+}
+</script>
+
+<Datatable>
+  <HeaderRow>
+    <HeaderItem>Item</HeaderItem>
+    <HeaderItem>Recent Activity</HeaderItem>
+    <HeaderItem>From</HeaderItem>
+    <HeaderItem numeric>Value</HeaderItem>
```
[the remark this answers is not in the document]
forevermatt: Yep. I submitted a PR on ui-components to fix that.

src/pages/steward/home.svelte
forevermatt: As a side note, I suspect our `isLoadingById()` isn't working the way we expect it to, because it's currently including the protocol-relative URL (including domain name) in those IDs. Example: `//cover.local:3000/users/me`, which isn't what we tend to pass to that method to see if it's loading.

I have some experimental code to fix that (so that the IDs match the the URL paths we use in our "data" files), but I haven't PR'd that.

PR #247, opened 2021-10-29
Treat any policyId in the URL as authoritative

forevermatt:
### Changed
- Change our Role/Policy menu (and `$rolePolicySelection` store) to react to URL changes that include a `policyId` parameter _if_ the user is acting as a customer

### Fixed
- Fix some automatic redirects to avoid leaving browser history entries

---

This does not yet change the various places in the application that we use `$user.policy_id` to instead get it from the `$rolePolicySelection.selectedPolicyId`, but that's my next step.

### DIFF
--- a/src/data/role-policy-selection.ts
+++ b/src/data/role-policy-selection.ts
@@ -36,3 +36,12 @@ export const recordPolicySelection = (policyId: string) => {
     selectedPolicyId: policyId,
   })
 }
+
+const isAdminRole = (role: UserAppRole | undefined) => role && ['Signator', 'Steward'].includes(role)
+
+export const reactToUrlChanges = (urlPolicyId: string | undefined) => {
+  if (urlPolicyId) {
+    const { selectedRole } = get(rolePolicySelection)
+    isAdminRole(selectedRole) || recordPolicySelection(urlPolicyId)
+  }
+}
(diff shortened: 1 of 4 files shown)


src/data/role-policy-selection.ts
```diff
@@ -36,3 +36,12 @@ export const recordPolicySelection = (policyId: string) => {
     selectedPolicyId: policyId,
   })
 }
+
+const isAdminRole = (role: UserAppRole | undefined) => role && ['Signator', 'Steward'].includes(role)
+
+export const reactToUrlChanges = (urlPolicyId: string | undefined) => {
```
RussellChamp, line 42: i don't think that's quite what we want. although the `RoleAndPolicyMenu.svelte` component shows up universally so that shouldn't be an immediate issue. for the moment this probably works, but i'll work on a readable store as a more permanent solution

RussellChamp (approved):
see comments

hobbitronics (approved):
lgtm

src/data/role-policy-selection.ts
forevermatt: Okay, thanks.

forevermatt: Feel free to do yours in another feature branch coming off of this one, PR'd back into this one when you're ready. I think that'll be the cleanest way for us to collaborate on this.

PR #248, opened 2021-11-01
add adminClaimStates and fix actual cost update bug

hobbitronics:
- I noticed the customer and admin see different banner while looking at the same claim
adminNeedsReviewClaim

### DIFF
--- a/src/data/states.ts
+++ b/src/data/states.ts
@@ -99,12 +99,15 @@ export const itemStates: { [stateName: string]: State } = {
 }
 
 export const getClaimState = (status: ClaimStatus, isAdmin = false): State => {
-  if (claimStates[status] === undefined && adminClaimStates[status] === undefined) {
-    console.error('No such state (for claim status):', status, Object.keys({ ...claimStates, ...adminClaimStates }))
-  }
   if (isAdmin) {
+    adminClaimStates[status] === undefined &&
+      console.error('No such state (for claim status):', status, Object.keys(adminClaimStates))
+
     return adminClaimStates[status] || ({} as State)
   } else {
+    claimStates[status] === undefined &&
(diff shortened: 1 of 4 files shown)


src/data/states.ts
```diff
@@ -73,18 +80,33 @@ export const claimStates: { [stateName: string]: State } = {
...
   Approved: { ...approved, title: 'Approved' },
   Inactive: { ...pending, title: 'This item has no coverage', icon: 'umbrella' },
   Pending: { ...pending, title: 'Awaiting item coverage review' },
 }
 
-export const getClaimState = (status: ClaimStatus): State => {
-  if (claimStates[status] === undefined) {
-    console.error('No such state (for claim status):', status, Object.keys(claimStates))
+export const getClaimState = (status: ClaimStatus, isAdmin = false): State => {
+  if (claimStates[status] === undefined && adminClaimStates[status] === undefined) {
```
forevermatt, line 102: Doing the check here, before narrowing down to whether `isAdmin` is true or false, means we could ask for an admin state but pass `false` for isAdmin and see neither the console error nor the state we were expecting.

What do you think of doing an `assertHas()` here? It's not what we typically use them for... just a thought.

Feel free to change this or not, as seems best to you for now.

forevermatt (approved):
Approved, with a comment/question

forevermatt:
Looks like the build failed due to a type error:
```
17:49:03.426 | src/pages/policies/[policyId]/claims/[claimId].svelte
17:49:03.428 | Error: [svelte-preprocess] Encountered type error
17:49:03.428 | at throwError (/opt/buildhome/repo/node_modules/svelte-preprocess/dist/modules/errors.js:5:11)
```

src/data/states.ts
hobbitronics: Good point. I would rather have better error handling than this.

PR #249, opened 2021-11-01
Enable admins to search for someone's policies

forevermatt:
### Changed
- Move `/policies` page to `/admin/policies`

### Added
- Enable an admin to search for policies by member name (first or last)
- Add readable data store with query string data


RussellChamp (commented):
lgtm

src/pages/admin/policies.svelte
```diff
@@ -1,43 +1,89 @@
...
-import { Datatable, Page } from '@silintl/ui-components'
+import { Datatable, Form, Page, TextField } from '@silintl/ui-components'
+import { onMount } from 'svelte'
+
+metatags.title = formatPageTitle('Policies')
 
-$: $initialized || loadPolicies()
-$: metatags.title = formatPageTitle('Policies')
+let matchingPolicies = []
+let searchFieldContents = ''
```
hobbitronics: Trying to figure out why you have both, are you trying to protect searchedFor from being changed by the binding?

hobbitronics (approved):
Just one question for my own understanding. Looks good.

src/pages/admin/policies.svelte
forevermatt: Correct. With our `Form` and `TextField` components, there doesn't seem to be an easy way to get the value when the submit event fires. Splitting out these two fields let me simply bind the value to the search field, and then when the event fires, do things with it. Otherwise it was showing "Policies matching something", as the user changed what they'd typed, even though I hadn't done new search yet.

PR #250, opened 2021-11-02
Use policyId from the url for everything

RussellChamp:
* Create a readable store for policyId that automatically updates when the url changes
* Use selectedPolicyId instead of $user.policy_id on appropriate pages
* Use this store to create derived stores for selectedPolicy, selectedPolicyItems, selectedPolicyClaims and use them as appropriate instead of manually figuring that out in each component
* Moved Group Settings under /policies/{policyId} so that you can edit the group settings of corporate policies
* Fix a variety of typos


src/components/ItemsTable.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { ItemDeleteModal } from 'components'
+import ItemDeleteModal from './ItemDeleteModal.svelte'
```
RussellChamp, line 2: resolves a 'circular dependency' issue. when in a component, DO NOT import from the 'components' index

src/components/RoleAndPolicyMenu.svelte
```diff
@@ -84,38 +78,31 @@ const getEntriesForRole = (role: UserAppRole | undefined): MenuItem[] => {
 
 const isAdminRole = (role: UserAppRole) => ['Signator', 'Steward'].includes(role)
 
-const tryToSetInitialRolePolicySelection = (
```
RussellChamp, line 87: do not manually set the policyId. we get that from the route and policyId store

src/components/RoleAndPolicyMenu.svelte
```diff
@@ -84,38 +78,31 @@ const getEntriesForRole = (role: UserAppRole | undefined): MenuItem[] => {
...
-      recordRoleSelection(actualRole)
-    } else if (corporatePolicies.length > 0) {
-      recordPolicySelection(corporatePolicies[0].id)
-    } else if (householdPolicies.length > 0) {
-      recordPolicySelection(householdPolicies[0].id)
-    }
+const setInitialRolePolicySelection = (actualRole: UserAppRole, policyId: string, myPolicies: Policy[]) => {
+  // If the current user is an admin but the policy selected ON STARTUP was one of their own,
+  // assume that they are editing their own policy as a user
+  if (myPolicies.some((p) => p.id === policyId)) {
```
RussellChamp, line 84: i changed some of the assumptions here. does this still fit the initial understanding?

src/components/RoleAndPolicyMenu.svelte
```diff
@@ -34,18 +28,18 @@ let roleEntries: MenuItem[]
...
 $: roleEntries = getEntriesForRole(role)
 $: corporatePolicyEntries = getCorporatePolicyEntries(myCorporatePolicies)
 $: householdPolicyEntries = getHouseholdEntries(myHouseholdPolicies)
 
 $: menuItems = [...roleEntries, ...corporatePolicyEntries, addCorporatePolicyEntry, ...householdPolicyEntries]
 
-const selectPolicy = (policyId: string) => {
-  recordPolicySelection(policyId)
+const selectPolicy = (userRole: UserAppRole, policyId: string) => {
+  recordRoleSelection(userRole)
```
forevermatt, line 42: This looks like a bug: selecting a policy used to force the role to be "User" (aka "Customer")... it may be helpful for you to walk me/us through your changes, to help us see the big picture design now.

src/components/RoleAndPolicyMenu.svelte
```diff
@@ -63,8 +57,8 @@ const getHouseholdEntries = (policies: Policy[]): MenuItem[] => {
   return policies.map((policy: Policy): MenuItem => {
     return {
       icon: 'family_restroom',
-      label: 'Household', // TODO: Replace with name, when available
-      action: () => selectPolicy(policy.id),
+      label: policy.household_id || 'Household',
```
forevermatt, line 60: When we get to real data, the `household_id` will just be a 6-digit number, so probably leaving this as always "Household" (until we have a better option) would be good at this point.

src/components/RoleAndPolicyMenu.svelte
forevermatt: Okay. We'll need some way to set a customer's initial policy ID to their first Corporate policy, or if they don't have one, to their Household policy, though.

src/pages/steward/home.svelte
```diff
@@ -19,7 +19,7 @@ $: $claimsInitialized || loadClaims()
 $: claimsAwaitingSteward = $claims.filter(isAwaitingSteward)
 $: claimsAwaitingSteward.map((claim) => claim.policy_id).forEach(loadDataOnce)
 
-$: items = [].concat(...Object.values($itemsByPolicyId))
+$: items = $allPolicyItems
 $: dependents = [].concat(...Object.values($dependentsByPolicyId))
```
RussellChamp, line 23: we can do the same to these later

src/components/RoleAndPolicyMenu.svelte
forevermatt: Hmm... I think it would be preferable for an admin to always start as their admin role, for consistency's sake. They'll know/learn how to switch to acting as a Customer, but that will be far less often than their desire to act as an admin, I would think.

src/components/RoleAndPolicyMenu.svelte
```diff
@@ -1,19 +1,13 @@
...
+import type { UserAppRole } from '../authn/user'
+import type { Policy } from 'data/policies'
+import { roleSelection, recordRoleSelection, selectedPolicyId } from 'data/role-policy-selection'
 import { POLICY_NEW_CORPORATE } from 'helpers/routes'
 import { Button, Menu, MenuItem } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
 
 export let myPolicies: Policy[]
-export let role: UserAppRole | undefined
+export let role: UserAppRole
```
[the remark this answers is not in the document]
RussellChamp: `undefined` is kinda an implicitly understood value for types but you usually only say it explicitly if you want to allow the value to be _assigned_ undefined.
eg if you wanted to be able to say `role = undefined` (which isn't the case here).

src/components/RoleAndPolicyMenu.svelte
RussellChamp: i can see how you'd think that. i can rework this.

src/components/RoleAndPolicyMenu.svelte
RussellChamp: i defaulted to what the $user.policy_id has

src/components/RoleAndPolicyMenu.svelte
RussellChamp: changed

src/components/RoleAndPolicyMenu.svelte
RussellChamp: changed

src/components/RoleAndPolicyMenu.svelte
RussellChamp: This is the only open "issue" on this PR that can maybe be resolve in the next one

src/components/RoleAndPolicyMenu.svelte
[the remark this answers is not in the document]
forevermatt: Ohhh... okay, thanks.

src/components/RoleAndPolicyMenu.svelte
```diff
@@ -33,23 +25,21 @@ let myCorporatePolicies: Policy[]
 let myHouseholdPolicies: Policy[]
 let roleEntries: MenuItem[]
 
-$: reactToUrlChanges($params.policyId) // TEMP
-
 $: myCorporatePolicies = myPolicies.filter(isCorporatePolicy)
 $: myHouseholdPolicies = myPolicies.filter(isHouseholdPolicy)
 
-$: $haveSetRolePolicySelection || tryToSetInitialRolePolicySelection(role, myCorporatePolicies, myHouseholdPolicies)
+$: setInitialRoleSelection(role)
```
forevermatt, line 31: Oh, clever. Since the `role` will presumably only change once (when the `/users/me` call comes back), then this will effectively only be called twice: once when `role` is undefined, and once when it has a real value.

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -41,20 +41,21 @@ import {
...
 import { assertHas } from '../../../../validation/assertions'
 import { onMount } from 'svelte'
-import { goto, metatags, params } from '@roxi/routify'
+import { goto, metatags } from '@roxi/routify'
 import { Page } from '@silintl/ui-components'
 import { formatDistanceToNow } from 'date-fns'
 
 export let claimId: string
-export let policyId: string = $params.policyId
+export let policyId = $selectedPolicyId
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:58]: Would there be a benefit to no longer exporting this property? As-is, it's value could come through two different routes, which feels concerning.

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -77,11 +78,11 @@ $: claimItem = claim.claim_items?.[0] || ({} as ClaimItem) //For now there will
 $: setInitialValues(claimItem)
 $: statusText = getClaimStatusText(claim, claimItem)
 
-$: items = $itemsByPolicyId[policyId] || []
+$: items = $selectedPolicyItems
 $: policyId && loadItems(policyId)
 $: item = items.find((itm) => itm.id === claimItem.item_id) || ({} as PolicyItem)
 
-$: isMemberOfPolicy = itemBelongsToPolicy($user.policy_id, item)
+$: isMemberOfPolicy = memberBelongsToPolicy($user.id, $policies, item.policy_id)
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:85]: It looks like this page can get its Policy ID from three different sources:
1. from routify injecting it (`export let policyId...`),
2. from our store (`... let policyId = $selectedPolicyId`), or
3. from an item (`memberBelongsToPolicy(..., ..., item.policy_id)`, derived from the claimId in the URL).

If we can reduce the number of places we look for a Policy ID, I think it will reduce odd bugs and ease troubleshooting.
```suggestion
$: isMemberOfPolicy = memberBelongsToPolicy($user.id, $policies, policyId)
```

Thoughts?

src/pages/policies/[policyId]/settings.svelte
```diff
@@ -168,5 +171,5 @@ p {
       </li>
     {/each}
   </ul>
-  <Button prependIcon="add" url="household/dependent" outlined>Add dependent</Button>
+  <Button prependIcon="add" url={addDependentUrl} outlined>Add dependent</Button>
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/settings.svelte:174]: I like how you did this: getting the policy-specific URL above, then passing it to the Button (rather than using a click event).

src/pages/policies/[policyId]/settings/dependents/new.svelte
```diff
@@ -1,25 +1,25 @@
...
 
 const onCancel = () => {
-  $goto(SETTINGS_HOUSEHOLD)
+  $goto(settingsPolicy(policyId))
 }
 const onSubmit = async (event: CustomEvent<string>) => {
   const formData = event.detail
-  await addDependent($user.policy_id, formData)
-  $goto(SETTINGS_HOUSEHOLD)
+  await addDependent(policyId, formData)
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/settings/dependents/new.svelte:21]: Nice fix

src/pages/settings.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
+import user from '../authn/user'
+import { settingsPolicy } from 'helpers/routes'
 import { redirect } from '@roxi/routify'
-import { SETTINGS_HOUSEHOLD } from 'helpers/routes'
 
-$redirect(SETTINGS_HOUSEHOLD)
+$: $user.policy_id && $redirect(settingsPolicy($user.policy_id))
```
forevermatt, line 6: This may end up needing to use `$selectedPolicyId`, so that if a customer is viewing one of their corporate policies, this will go to that policy's settings.

src/pages/steward/home.svelte
forevermatt: Agreed, though we might even be able to refactor these out at some point. I think they're just here to try to ensure the admin has all the related data, for populating their ClaimCards.

forevermatt (changes requested):
Several comments added (sorry), but I would only consider a couple of them blockers.

src/data/policies.ts
```diff
@@ -37,6 +38,9 @@ export type UpdatePolicyRequestBody = {
 }
 
 export const policies = writable<Policy[]>([])
+export const selectedPolicy = derived([policies, selectedPolicyId], ([policies, selectedPolicyId]) => {
+  return policies.find((p) => p.id === selectedPolicyId) || policies[0] || {}
```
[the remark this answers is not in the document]
RussellChamp: fixed

src/data/policies.ts
```diff
@@ -135,3 +139,10 @@ export const getPolicyById = (policyId: string): Policy => {
   const policy = get(policies).find((policy) => policy.id === policyId) || ({} as Policy)
   return policy
 }
+
+export const memberBelongsToPolicy = (memberId: string, policies: Policy[], policyId: string): boolean => {
+  const policy = policies?.find((p) => p.id === policyId)
+  const member = policy?.members?.find((m) => m.id === memberId)
```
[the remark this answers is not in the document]
RussellChamp: gonna keep it this way. i don't think there is a performance difference and this helps when debugging

src/pages/policies/[policyId]/claims/[claimId].svelte
Comment (RussellChamp) [reply src/pages/policies/[policyId]/claims/[claimId].svelte:85]: fixed. i was trying to think if there was any case when the api might return us policy items where the policy_id is not what we just retrieved from. probably not worthwhile worrying about. :)

src/pages/_layout.svelte
```diff
@@ -1,14 +1,16 @@
...
+import { initialized as policiesInitialized, loadPolicies, policies, Policy } from 'data/policies'
 import * as routes from 'helpers/routes'
 import { goto } from '@roxi/routify'
+import { roleSelection, selectedPolicyId } from 'data/role-policy-selection'
 
 $: $policiesInitialized || loadPolicies()
 
-$: myPolicies = ($user.policies || []) as Policy[]
-$: myHouseholdPolicyId = $user.policy_id
+$: myPolicies = $policies.filter((p) => p.members?.some((m) => m.id === $user.id))
```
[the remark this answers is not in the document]
RussellChamp: resolved. had to add policies to the user type

src/pages/settings.svelte
RussellChamp: no, this is the "shortcut" route when a policy isn't specified at all. we want to default to something in that case

forevermatt (approved):

PR #251, opened 2021-11-02
improve file preview

hobbitronics:
- add and use isFileUploadedByPurpose instead of just checking claimFiles length
- show a banner with purpose to admin instead of Delete button
- open files in new tab
- make cursor and background color change when hovering over FilePreview to show its clickable
- fix a number of ts errors/warnings

### DIFF
--- a/src/components/FilePreview/FilePreview.svelte
+++ b/src/components/FilePreview/FilePreview.svelte
@@ -2,6 +2,7 @@
 import type { ClaimFile } from 'data/claims'
 import { formatDate } from '../dates'
 import { Button, Progress } from '@silintl/ui-components'
+import { Banner } from 'components'
 import { createEventDispatcher } from 'svelte'
 import { flip } from 'svelte/animate'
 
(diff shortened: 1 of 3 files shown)


src/components/FilePreview/FilePreview.svelte
```diff
@@ -1,36 +1,39 @@
 <script lang="ts">
+import { Banner } from 'components'
```
forevermatt, line 2: Since this import is within a component, it's probably better to import the `Banner` component directly (rather than `from 'components'`) to avoid circular dependencies.
```suggestion
import Banner from 'components/Banner.svelte'
```

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -328,7 +334,17 @@ const getClaimStatusText = (claim: Claim, item: ClaimItem) => {
         {/if}
       {/if}
 
-      <FilePreview class="w-50" previews={claimFiles} on:deleted={onDeleted} on:preview={onPreview} />
+      {#if showImg}
+        <img class="receipt" src={previewFile.file?.url} alt="receipt" on:error={onImgError} />
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:338]: Is this only used for receipts, or is it also used for evidence of FMV? Mostly just wondering about that `alt="receipt"`.

forevermatt (approved):
Approved, with one suggestion and one question

src/pages/policies/[policyId]/claims/[claimId].svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/claims/[claimId].svelte:338]: its used for both, so perhaps "document" is better

PR #252, opened 2021-11-02
Use new `policy.name` field for corporate policies

forevermatt:
### Changed
- Use new `policy.name` field for corporate policies


hobbitronics (approved):
Ahh, a nice short PR.

PR #254, opened 2021-11-02
Fix `isLoadingById()`

forevermatt:
### Fixed
- Refactor isLoadingById() data to use relative API paths
- Fix usage of isLoadingById() to have correct paths and react better

### DIFF
--- a/src/pages/policies/[policyId]/claims/[claimId]/edit.svelte
+++ b/src/pages/policies/[policyId]/claims/[claimId]/edit.svelte
@@ -60,7 +60,7 @@ const onSubmit = async (event: CustomEvent) => {
 }
 </script>
 
-{#if $loading && isLoadingById(`claimss/${claimId}`)}
+{#if $loading && isLoadingById(`claims/${claimId}`)}
   Loading...
 {:else if claims && !claim.id}
   We could not find that claim. Please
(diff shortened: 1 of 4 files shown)


hobbitronics (approved):
Ya fixed it!

src/pages/policies/[policyId]/claims/[claimId]/edit.svelte
```diff
@@ -60,7 +60,7 @@ const onSubmit = async (event: CustomEvent) => {
 }
 </script>
 
-{#if isLoadingById(claimId)}
+{#if $loading && isLoadingById(`claimss/${claimId}`)}
```
Comment (RussellChamp) [inline src/pages/policies/[policyId]/claims/[claimId]/edit.svelte:63]: personally i don't think we should be hand-coding the uri like this as it requires your component to be aware of the API implementation. it also allows for typos like the above `claimss` vs `claims`

RussellChamp (changes requested):

src/pages/policies/[policyId]/claims/[claimId]/edit.svelte
Comment (forevermatt) [reply src/pages/policies/[policyId]/claims/[claimId]/edit.svelte:63]: Agreed. We'll refactor this later. 👍

RussellChamp (approved):

PR #255, opened 2021-11-03
rename customer user role from 'User' to 'Customer'

RussellChamp:
* remove AdminAppRole type and make UserAppRole an enum
* Update all hard-coded referenecs of app roles to use enum

### DIFF
--- a/src/authn/user.ts
+++ b/src/authn/user.ts
@@ -2,8 +2,11 @@ import { get, writable } from 'svelte/store'
 import { CREATE, GET, UPDATE } from 'data'
 import type { Policy } from 'data/policies'
 
-export type AdminAppRole = 'Steward' | 'Signator'
-export type UserAppRole = 'User' | AdminAppRole
+export enum UserAppRole {
+  Customer = 'Customer',
+  Steward = 'Steward',
+  Signator = 'Signator',
+}
 
 export type User = {
   app_role: UserAppRole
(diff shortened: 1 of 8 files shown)


src/authn/user.ts
```diff
@@ -2,8 +2,11 @@ import { get, writable } from 'svelte/store'
 import { CREATE, GET, UPDATE } from 'data'
 import type { Policy } from 'data/policies'
 
-export type AdminAppRole = 'Steward' | 'Signator'
-export type UserAppRole = 'User' | AdminAppRole
+export enum UserAppRole {
```
RussellChamp, line 5: enums in typescript are really great! https://www.typescriptlang.org/docs/handbook/enums.html
under the covers it basically creates an object that can be used to convert back and forth from an enum's name and value.
By default, enums values are number but it is easy to give them string values instead (as seen below).
You can also use an enum as a type! This makes enums something of a "super" type that doesn't "wash away" during the ts transpilation step.
Using string-value enums means that you have a single "source of truth" for those values and are even more resistant to fat-finger bugs and typos. It also means that you can make sweeping changes by just updating the single enum value!

RussellChamp: I think we could do a similar thing for the item/claim status values but that's for another PR. :)

forevermatt (approved):

hobbitronics (approved):
nice!

PR #256, opened 2021-11-03
Fix admins' "policy search" feature

forevermatt:
### Fixed
- Fix admins' "policy search" feature

### Changed
- Rename store holding query string data to urlQuery

### DIFF
--- a/src/data/query-string.ts
+++ b/src/data/query-string.ts
@@ -1,7 +1,7 @@
-import { params } from '@roxi/routify'
+import { route } from '@roxi/routify'
 import { derived } from 'svelte/store'
 
-export const query = derived(params, () => {
+export const urlQuery = derived(route, () => {
   const searchParams = new URLSearchParams(window.location.search)
   const results: any = {}
   for (const [key, value] of searchParams) {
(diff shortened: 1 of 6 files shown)


src/data/query-string.ts
```diff
@@ -1,7 +1,7 @@
-import { params } from '@roxi/routify'
+import { route } from '@roxi/routify'
 import { derived } from 'svelte/store'
 
-export const query = derived(params, () => {
+export const urlQuery = derived(route, () => {
```
hobbitronics, line 4: maybe I've forgotten the context but what is route doing here?

hobbitronics (approved):
Looks nice. Just one question.

src/data/query-string.ts
forevermatt: No, good question. That's the store that triggers changes to this derived store. Any time the Svelte `route` store changes (aka the URL changes), I want to recalculate what the query string parameters are. I happen to not use Svelte's route store for that calculation, but it's the event I want to listen for, so to speak.

hobbitronics: ok, that makes sense.

PR #257, opened 2021-11-03
Fix `account_detail` to be optional in `CreatePolicyRequestBody`

forevermatt:
### Fixed
- Fix `account_detail` to be optional in `CreatePolicyRequestBody`


hobbitronics (approved):
This is probably a tie for shortest PR ever.

PR #259, opened 2021-11-03
use role selection instead of app role

hobbitronics:
- added 'Admin' to UserAppRole as its in the api (not sure what its use case is)
- added ItemDetails to index
- updated adminItemStates.Pending
- use $roleSelection for isAdmin

### DIFF
--- a/src/authn/user.ts
+++ b/src/authn/user.ts
@@ -6,6 +6,7 @@ export enum UserAppRole {
   Customer = 'Customer',
   Steward = 'Steward',
   Signator = 'Signator',
+  Admin = 'Admin',
 }
 
 export type User = {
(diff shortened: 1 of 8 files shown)


RussellChamp (approved):
lgtm but i would expect we only have export params like 'isAdmin' on 'ItemDetails' if you want the ability to toggle the admin-ness of that component apart from the rest of the app. else you have a 'two sources of truth' issue again. thoughts?

src/authn/user.ts
```diff
@@ -68,6 +69,7 @@ export const isUserSteward = (user: User): boolean => user.app_role === UserAppR
 
 export const isSignator = (user: User): boolean => user.app_role === UserAppRole.Signator
```
RussellChamp, line 70: even though these work perfectly fine as-is, i was thinking they could also be written as derived stores of the current user (unless we're also doing checks on the non-current user) but that may be a bit overkill

hobbitronics: I like this idea of using more derived stores. A more general question now that we have `$roleSelection` do we even need to be using `$user.app_role`? It seems like we are just using it to double check the role.

PR #262, opened 2021-11-04
Receipt status cause bugs on Fmv & Evacuation

hobbitronics:
- Was getting an error for adding actual values
- was getting a blank page after uploading a file
- The api needs to stop giving status: Receipt for FMV and Evacuation payoutOptions


src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -183,6 +183,10 @@ const onPreview = (event: CustomEvent<string>) => {
 const onImgError = () => (showImg = false)
 
 const onBlur = () => {
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:185]: It might be worth renaming `onBlur` to include something about which field the blur happened on, but that's just a suggestion.

forevermatt (approved):

PR #263, opened 2021-11-04
Fix steward claim cards

forevermatt:
### Fixed
- Fix steward home page to show the appropriate claim cards
- Fix list of claims awaiting each admin role to include Review3

### Added
- Enable getting (from the API) the claims awaiting a specific admin role

---

There's a bug in the API causing an admin's call to "GET /claims" to not return anything, but it works if we specify which status claims we want. We already have a list of which claims are awaiting each admin role (steward, signator), so this simply makes use of that and may help when merging the steward and signator home pages.

### DIFF
--- a/src/pages/steward/home.svelte
+++ b/src/pages/steward/home.svelte
@@ -1,15 +1,9 @@
 <script lang="ts">
+import { UserAppRole } from '../../authn/user'
 import { ClaimCards, RecentActivityTable, Row } from 'components'
 import { loading } from 'components/progress'
 import { getDependentOptions, getPolicyMemberOptions } from 'data/accountablePersons'
-import {
-  Claim,
-  claims,
-  getClaimsAwaitingAdmin,
-  initialized as claimsInitialized,
-  getClaimsAwaitingAdmin,
-  statusesAwaitingSteward,
-} from 'data/claims'
+import { Claim, getClaimsAwaitingAdmin } from 'data/claims'
 import { dependentsByPolicyId, loadDependents } from 'data/dependents'
 import { allPolicyItems, itemsByPolicyId, loadItems } from 'data/items'
 import { loadMembersOfPolicy, membersByPolicyId } from 'data/policy-members'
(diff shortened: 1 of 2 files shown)


src/pages/steward/home.svelte
```diff
@@ -2,21 +2,29 @@
...
+} from 'data/claims'
 import { dependentsByPolicyId, loadDependents } from 'data/dependents'
 import { allPolicyItems, itemsByPolicyId, loadItems } from 'data/items'
 import { loadMembersOfPolicy, membersByPolicyId } from 'data/policy-members'
 import { loadRecentActivity, recentChanges } from 'data/recent-activity'
+import { roleSelection } from 'data/role-policy-selection'
 import { customerClaimDetails } from 'helpers/routes'
 import { goto } from '@roxi/routify'
 import { Page } from '@silintl/ui-components'
+import { UserAppRole } from '../../authn/user'
```
hobbitronics, line 21: looks like this probably was autoimported and not sorted.

hobbitronics (approved):
I am expecting a similar PR for the signator home page. Does the signator not get to pre-Approve?

src/pages/steward/home.svelte
forevermatt: Yep, thanks. Cleaned up imports.

forevermatt:
> I am expecting a similar PR for the signator home page. Does the signator not get to pre-Approve?

Hmm... that's odd. I think I must have simply left that our accidentally. I'll add "Review1" to the list for Signator.

src/pages/steward/home.svelte
```diff
@@ -27,8 +28,10 @@ $: dependentOptions = getDependentOptions(dependents)
 $: policyMemberOptions = getPolicyMemberOptions(policyMembers)
 $: accountablePersons = [...policyMemberOptions, ...dependentOptions]
 
-const isAwaitingSteward = (claim: Claim): boolean => {
-  return statusesAwaitingSteward.includes(claim.status)
+const loadClaimsAwaitingAdmin = async (role: UserAppRole) => {
+  if (role === UserAppRole.Steward) {
```
RussellChamp, line 32: did this need to pass `role` or could it have referenced `roleSelection` directly? alternately, `getClaimsAwaitingAdmin` could have been written to reference that store directly

forevermatt: I really like to keep functions "pure" (aka. they only use data that's handed to them, among other things) to reduce complexity due to coupled code.

However, from a pragmatic standpoint, yes: this `load...()` method is small and clear enough it could have simply used `$roleSeleciton` internally. Good observation.

PR #265, opened 2021-11-04
Clean up claim cards

forevermatt:
### Fixed
- Refactor claim cards to use the item data provided with the claim data
- Update two uses of ClaimCards to stop passing in items 

I didn't update the admin home pages' uses because they'll be (re)moved in a forthcoming PR.


hobbitronics (approved):
looks good

PR #266, opened 2021-11-04
Add missing "Review3a" and "Review3b" statuses, where applicable

forevermatt:
### Fixed
- Add missing "Review3a" and "Review3b" statuses, where applicable

---

At the moment the backend is using "Review3", but the state diagram (and some discussion awhile back) indicates that it will be replaced with "Review3a" and "Review3b" statuses, to ensure that two different people sign off on payouts.

This PR just fills in the gaps so that we support all three options. Once the backend has been updated, we can remove the "Review3" status.

### DIFF
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -28,6 +28,8 @@ $: switch (status) {
     actionLabel = action
     break
   case 'Review3':
+  case 'Review3a':
+  case 'Review3b':
     action = 'approve'
     actionLabel = 'give final approval'
     break
@@ -72,7 +74,7 @@ const onDeny = () => dispatch('deny', message)
 </style>
 
 {#if isAdmin}
-  {#if ['Review1', 'Review2', 'Review3'].includes(status)}
+  {#if ['Review1', 'Review2', 'Review3', 'Review3a', 'Review3b'].includes(status)}
     <div class="container">
       <div class="text-input">
         <TextField class="w-100" label="Send a message" bind:value={message} />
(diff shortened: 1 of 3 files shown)


src/components/ClaimActions.svelte
```diff
@@ -82,10 +84,10 @@ const onDeny = () => dispatch('deny', message)
         <Button on:click={onDeny} disabled={!message} outlined>deny</Button>
       </div>
       <div class="right-buttons">
-        {#if ['Review1', 'Review3'].includes(status)}
+        {#if ['Review1', 'Review3', 'Review3a', 'Review3b'].includes(status)}
           <Button on:click={onAskForChanges} disabled={!message} outlined>ask for changes</Button>
         {/if}
-        {#if ['Review2', 'Review3'].includes(status)}
+        {#if ['Review2', 'Review3', 'Review3a', 'Review3b'].includes(status)}
```
hobbitronics, line 90: I wonder if its worth moving these to a const named something like `FixReceiptClaimStatus` and `RequestRevisionClaimStatus`. Just something to consider.

hobbitronics (approved):
Looks good as is, just one comment.

src/components/ClaimActions.svelte
forevermatt: I would say so. I wasn't sure of the purpose/meaning of these arrays, so extracting them to a variable with a meaningful name would probably be very helpful. I'll leave that for you/someone to do in a later PR, as it sounds like you have more understanding of what these conditions are intended to check.

PR #268, opened 2021-11-05
get entity codes from api and suggest in corporate policy form

hobbitronics:
- used SearchableSelect to suggest entity codes
- fixed circular dependency

### DIFF
--- a/src/data/entityCodes.ts
+++ b/src/data/entityCodes.ts
@@ -1,0 +1,18 @@
+import { GET } from '.'
+import { writable } from 'svelte/store'
+
+export type EntityCode = {
+  code: 'string'
+  id: 'string'
+  name: 'string'
+}
+
+export const entityCodes = writable<EntityCode[]>([])
+
+export async function loadEntityCodes(): Promise<void> {
+  const urlPath = 'config/entity-codes'
+
+  const results = await GET<EntityCode[]>(urlPath)
+
+  entityCodes.set(results)
+}
--- a/src/data/entityCodes.ts
+++ b/src/data/entityCodes.ts
@@ -2,9 +2,9 @@ import { GET } from '.'
 import { writable } from 'svelte/store'
 
 export type EntityCode = {
-  code: 'string'
-  id: 'string'
-  name: 'string'
+  code: string
+  id: string
+  name: string
 }
 
 export const entityCodes = writable<EntityCode[]>([])
--- a/src/pages/policies/[policyId]/settings.svelte
+++ b/src/pages/policies/[policyId]/settings.svelte
@@ -13,7 +13,7 @@ import { Button, TextField, IconButton, Page, setNotice } from '@silintl/ui-comp
 
 const policyData = {} as Policy
 
-loadEntityCodes()
+$: $entityCodes.length || loadEntityCodes()
 
 $: policyId = $selectedPolicyId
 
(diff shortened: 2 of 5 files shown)


src/pages/policies/new.svelte
```diff
@@ -11,8 +12,16 @@ let account = ''
...
+let entityOptions: any = {}
+
+loadEntityCodes()
 
 $: metatags.title = formatPageTitle('New Corporate Policy')
+$: $entityCodes.length
+  ? $entityCodes.forEach((code) => {
+      entityOptions[code.name] = code.code
+    })
+  : (entityOptions.ABC = 'ABC')
```
hobbitronics, line 24: I used this since there weren't any active entity codes being returned by the api, but I should probably get rid of it.

src/pages/policies/[policyId]/settings.svelte
```diff
@@ -124,17 +131,18 @@ p {
 
 <Page>
   <Breadcrumb links={breadcrumbLinks} />
+  {#if policy.type === PolicyType.Household}
+    <p>
+      <span class="header">Household ID<span class="required">*</span></span>
+      <TextField placeholder={'1234567'} autofocus bind:value={householdId} on:blur={updateHouseholdId} />
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/settings.svelte:137]: Let's remove the autofocus on this field. Any time I visit the Settings page, it tries to update my Household ID, which seems a little too "proactive". 🙂

forevermatt (commented):
Approved, with one code suggestion

src/data/entityCodes.ts
```diff
@@ -0,0 +1,18 @@
+import { GET } from '.'
+import { writable } from 'svelte/store'
+
+export type EntityCode = {
+  code: 'string'
```
RussellChamp, line 5: mistake! you want this to be `code: string` else you are saying that the type allowed to `code` is the string literal "s-t-r-i-n-g"

src/pages/policies/[policyId]/settings.svelte
```diff
@@ -12,9 +13,12 @@ import { Button, TextField, IconButton, Page, setNotice } from '@silintl/ui-comp
 
 const policyData = {} as Policy
 
+loadEntityCodes()
```
Comment (RussellChamp) [inline src/pages/policies/[policyId]/settings.svelte:16]: should this load be called every time or should we used cached result for entity codes?

src/pages/policies/new.svelte
```diff
@@ -11,9 +12,14 @@ let account = ''
 let costCenter = ''
 let groupName = ''
 let entityCode = ''
+let entityOptions: any = {}
 
-$: metatags.title = formatPageTitle('New Corporate Policy')
+loadEntityCodes()
```
RussellChamp, line 17: same here

RussellChamp (changes requested):

src/pages/policies/[policyId]/settings.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/settings.svelte]: no need to load if the store is already full. I just pushed a change

src/pages/policies/[policyId]/settings.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/settings.svelte:137]: I added this change to the subsequent PR which relies on code from this PR.

RussellChamp (approved):

PR #269, opened 2021-11-05
add fields to settings for corporate policies

hobbitronics:
- added all the fields that might be accepted by the endpoint
- made them optional in the Type
- show the key and dispatch the value in SearchableSelect

### DIFF
--- a/src/data/policies.ts
+++ b/src/data/policies.ts
@@ -36,10 +36,12 @@ export type CreatePolicyRequestBody = {
 }
 
 export type UpdatePolicyRequestBody = {
-  account: string
-  cost_center: string
-  entity_code: string
-  household_id: any /* This seemed complicated */
+  account?: string
+  account_detail?: string
+  cost_center?: string
+  entity_code?: string
+  household_id?: any /* This seemed complicated */
+  name?: string
 }
 
 export const policies = writable<Policy[]>([])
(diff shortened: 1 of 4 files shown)


src/data/policies.ts
```diff
@@ -69,15 +74,8 @@ const updatePoliciesStore = (changedPolicy: Policy) => {
  * @param {Number} id
  * @param {Object} policyData
  */
-export async function updatePolicy(id: string, policyData: any): Promise<void> {
-  const parsedPolicyData: UpdatePolicyRequestBody = {
```
RussellChamp, line 73: i'd been hoping we could remove this in-between step for a while now. ;)

src/pages/policies/[policyId]/settings.svelte
```diff
@@ -37,20 +40,46 @@ $: $entityCodes.forEach((code) => {
...
 $: policy.household_id && setPolicyHouseholdId()
 $: policy.cost_center && setPolicyCostCenter()
-$: policy.entity_code && setAffiliation()
+$: policy.entity_code && setEntityCode()
+$: policy.name && setPolicyName()
+
 $: addDependentUrl = householdSettingsNewDependent(policyId)
 
-const setAffiliation = () => (affiliationChoice = entityOptions[policy.entity_code])
+const setAccount = () => (account = policy.account || '')
```
Comment (RussellChamp) [inline src/pages/policies/[policyId]/settings.svelte:53]: all these functions seem superfluous. wouldn't it be more 'svelty' to say like `$: account = policy.account || ''` and why are these variables used at all instead of just referencing `policy.account` etc?

RussellChamp (changes requested):

RussellChamp (approved):

PR #271, opened 2021-11-05
Don't show the "Add Item" button until we have a Policy ID

forevermatt:
### Fixed
- Don't show the "Add Item" button until we have a Policy ID

---

This prevents the user form going to the add-item page using a `null` Policy ID.


hobbitronics (approved):
I was bracing myself for long one, lol

PR #273, opened 2021-11-05
add account detail to new policy form

hobbitronics:


src/data/policies.ts
```diff
@@ -96,6 +96,7 @@ export async function createPolicy(policyFormData: any): Promise<Policy> {
     cost_center: policyFormData.costCenter,
     entity_code: policyFormData.entityCode,
     name: policyFormData.groupName,
+    account_detail: policyFormData.accountDetail,
```
forevermatt, line 99: Would you mind alphabetizing this?

src/pages/policies/new.svelte
```diff
@@ -69,5 +71,10 @@ const validateForm = (formData: any) => {
     <TextField placeholder="12345" bind:value={account} />
   </p>
 
+  <p>
+    <span class="header">Account Detail</span>
+    <TextField placeholder="details" bind:value={accountDetail} />
```
forevermatt, line 76: Would you mind moving the `<Description>Appears in your statements</Description>` line to here (and removing it from the Name field)? Currently, this is the field that appears in their statements, though that may change later.

hobbitronics: yup

src/data/policies.ts
hobbitronics: no problemo

forevermatt (approved):

PR #274, opened 2021-11-05
Filter claim cards to those that are recent or incomplete

forevermatt:
### Fixed
- Only show claim cards that are about incomplete and/or recent claims
- Update ClaimItemStatus possible values


hobbitronics (approved):
this looks good to me

PR #275, opened 2021-11-05
validate entity code

hobbitronics:
- Warn user to choose a valid entity code in settings (already happens at submit for new policy)
- clear Searchable Select input on focus so all datalist options are visible
- tweak styling
- add on:change to provide some user feedback when no selection is made


RussellChamp (approved):
lgtm

forevermatt (approved):

PR #280, opened 2021-11-08
Feat/add policy items to cards grid

RussellChamp:
* Create a CardsGrid component that includes data for both claims and policy items
* Populate with data from recent activity endpoint

This PR introduced the TypeScript concept of [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) (or type guards). This is quite useful when working with arrays of data that could contain multiple types (such as the store we use for recent activity)

### DIFF
--- a/src/components/CardsGrid.svelte
+++ b/src/components/CardsGrid.svelte
@@ -3,7 +3,7 @@ import ClaimCard from './ClaimCard.svelte'
 import ItemCard from './ItemCard.svelte'
 import type { AccountablePersonOptions } from 'data/accountablePersons'
 import { Claim, ClaimItem, incompleteClaimItemStatuses, isClaimItem } from 'data/claims'
-import { incompleteItemCoverageStatus, PolicyItem } from 'data/items'
+import { incompleteItemCoverageStatuses, PolicyItem } from 'data/items'
 import { isRecentClaim, RecentChange } from 'data/recent-activity'
 
 type CardItem = { data: ClaimItem | PolicyItem; claim?: Claim }
@@ -14,12 +14,12 @@ export let policyItems: PolicyItem[] = []
 export let accountablePersons: AccountablePersonOptions[] = []
 export let isAdmin: boolean = false
 
-$: items = recentChanges.length ? getRecentChanges(recentChanges) : mergeClaimsAndPolicyItems(claims, policyItems)
+$: items = recentChanges.length ? parseRecentChanges(recentChanges) : parseClaimsAndPolicyItems(claims, policyItems)
 
 const isIncomplete = (card: CardItem) => {
   return isClaimItem(card.data)
     ? incompleteClaimItemStatuses.includes(card.data.status)
-    : incompleteItemCoverageStatus.includes(card.data.coverage_status)
+    : incompleteItemCoverageStatuses.includes(card.data.coverage_status)
 }
 
 const isRecent = (card: CardItem) => {
@@ -34,7 +34,7 @@ const isRecentOrIncomplete = (card: CardItem) => isIncomplete(card) || isRecent(
 const sortByUpdateDate = (cardA: CardItem, cardB: CardItem) =>
   +new Date(cardB.data.updated_at) - +new Date(cardA.data.updated_at)
 
-const getRecentChanges = (changes: RecentChange[]): CardItem[] => {
+const parseRecentChanges = (changes: RecentChange[]): CardItem[] => {
   let cards: CardItem[] = []
   for (let change of changes) {
     if (isRecentClaim(change)) {
@@ -45,7 +45,8 @@ const getRecentChanges = (changes: RecentChange[]): CardItem[] => {
   }
   return cards
 }
-const mergeClaimsAndPolicyItems = (claims: Claim[], policyItems: PolicyItem[]): CardItem[] => {
+
+const parseClaimsAndPolicyItems = (claims: Claim[], policyItems: PolicyItem[]): CardItem[] => {
   // Add all the policy items and claim items together then filter and sort them
   let cards: CardItem[] = policyItems.map((item) => ({ data: item })) || []
   for (let claim of claims) {
--- a/src/components/CardsGrid.svelte
+++ b/src/components/CardsGrid.svelte
@@ -1,0 +1,77 @@
+<script lang="ts">
+import ClaimCard from './ClaimCard.svelte'
+import ItemCard from './ItemCard.svelte'
+import type { AccountablePersonOptions } from 'data/accountablePersons'
+import { Claim, ClaimItem, incompleteClaimItemStatuses, isClaimItem } from 'data/claims'
+import { incompleteItemCoverageStatus, PolicyItem } from 'data/items'
+import { isRecentClaim, RecentChange } from 'data/recent-activity'
+
+type CardItem = { data: ClaimItem | PolicyItem; claim?: Claim }
+
+export let recentChanges: RecentChange[] = []
+export let claims: Claim[] = []
+export let policyItems: PolicyItem[] = []
+export let accountablePersons: AccountablePersonOptions[] = []
+export let isAdmin: boolean = false
+
+$: items = recentChanges.length ? getRecentChanges(recentChanges) : mergeClaimsAndPolicyItems(claims, policyItems)
+
+const isIncomplete = (card: CardItem) => {
+  return isClaimItem(card.data)
+    ? incompleteClaimItemStatuses.includes(card.data.status)
+    : incompleteItemCoverageStatus.includes(card.data.coverage_status)
+}
+
+const isRecent = (card: CardItem) => {
+  const updatedAt = new Date(card.data.updated_at)
+  const today = new Date()
+  const aWeekAgo = new Date(today.getFullYear(), today.getMonth(), today.getDate() - 7)
+  return Number(updatedAt) > Number(aWeekAgo)
+}
+
+const isRecentOrIncomplete = (card: CardItem) => isIncomplete(card) || isRecent(card)
+
+const sortByUpdateDate = (cardA: CardItem, cardB: CardItem) =>
+  +new Date(cardB.data.updated_at) - +new Date(cardA.data.updated_at)
+
+const getRecentChanges = (changes: RecentChange[]): CardItem[] => {
+  let cards: CardItem[] = []
+  for (let change of changes) {
+    if (isRecentClaim(change)) {
+      cards.push({ data: change.Claim?.claim_items[0], claim: change.Claim })
+    } else {
+      cards.push({ data: change.Item })
+    }
+  }
+  return cards
+}
+const mergeClaimsAndPolicyItems = (claims: Claim[], policyItems: PolicyItem[]): CardItem[] => {
+  // Add all the policy items and claim items together then filter and sort them
+  let cards: CardItem[] = policyItems.map((item) => ({ data: item })) || []
+  for (let claim of claims) {
+    for (let claimItem of claim.claim_items) {
+      cards.push({ data: claimItem, claim })
+    }
+  }
+  return cards.filter(isRecentOrIncomplete).sort(sortByUpdateDate)
+}
+</script>
+
+<style>
+.card {
+  margin: 8px 8px 8px 0;
+  flex-basis: 330px;
+}
+</style>
+
+<div class="flex justify-start flex-wrap {$$props.class}">
+  {#each items as item (item.data.id)}
+    <div class="card">
+      {#if isClaimItem(item.data)}
+        <ClaimCard claim={item.claim} claimItem={item.data} {accountablePersons} {isAdmin} on:goto-claim />
+      {:else}
+        <ItemCard item={item.data} {accountablePersons} {isAdmin} on:goto-item />
+      {/if}
+    </div>
+  {/each}
+</div>
--- a/src/components/CardsGrid.svelte
+++ b/src/components/CardsGrid.svelte
@@ -14,7 +14,7 @@ export let policyItems: PolicyItem[] = []
 export let accountablePersons: AccountablePersonOptions[] = []
 export let isAdmin: boolean = false
 
-$: items = recentChanges.length ? parseRecentChanges(recentChanges) : parseClaimsAndPolicyItems(claims, policyItems)
+$: cardItems = recentChanges.length ? parseRecentChanges(recentChanges) : parseClaimsAndPolicyItems(claims, policyItems)
 
 const isIncomplete = (card: CardItem) => {
   return isClaimItem(card.data)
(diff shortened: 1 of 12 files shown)


src/components/CardsGrid.svelte
```diff
@@ -0,0 +1,77 @@
...
+.card {
+  margin: 8px 8px 8px 0;
+  flex-basis: 330px;
+}
+</style>
+
+<div class="flex justify-start flex-wrap {$$props.class}">
+  {#each items as item (item.data.id)}
+    <div class="card">
+      {#if isClaimItem(item.data)}
```
RussellChamp, line 70: using the type predicate here means that the compiler can now reason about what the concrete type of your variable actually is. in the below code, the compiler is able to figure out that the `item.data` in the first block is actually a `ClaimItem` and the one in the else block is a `PolicyItem`. yay!

src/data/claims.ts
```diff
@@ -71,6 +71,10 @@ export type ClaimItem = {
   updated_at: string /*Date*/
 }
 
+export function isClaimItem(obj: any): obj is ClaimItem {
```
hobbitronics, line 74: just checking you was "obj is ClaimItem" as I am used to seeing "as" but then is seems you are returning a boolean so maybe "is" is correct.

src/components/ItemCard.svelte
```diff
@@ -0,0 +1,74 @@
...
+  -webkit-line-clamp: 2; /* number of lines to show */
+}
+
+.ml-50px {
+  margin-left: 50px;
+}
+</style>
+
+<Card noPadding class="h-300 py-0 {$$props.class}">
+  <ClaimCardBanner {statusReason} {state} {showRevisionMessage} />
```
forevermatt, line 53: This would probably be best done in a separate PR, but if `ClaimCardBanner` is truly general-purpose enough to work both for a ClaimCard and for an ItemCard, let's rename it to `CardBanner` to avoid confusion and reduce potential bugs.

src/data/recent-activity.ts
```diff
@@ -1,42 +1,56 @@
...
 import type { Claim } from './claims'
 import type { PolicyItem } from './items'
-import { writable } from 'svelte/store'
+import { derived, writable } from 'svelte/store'
 
 export type RecentClaim = {
   Claim: Claim
   StatusUpdatedAt: string /* Date */
 }
+export function isRecentClaim(obj: any): obj is RecentClaim {
```
forevermatt, line 10: In the docs on type predicates, they use type unions as the argument type. Is there an advantage to using `any` here rather than `RecentChange`?

(Similarly below on `isRecentItem(obj: any)`)

src/data/recent-activity.ts
```diff
@@ -1,42 +1,56 @@
...
-type RecentChange = RecentClaim | RecentItem
+export type RecentChange = RecentClaim | RecentItem
 
 export type RecentActivity = {
   Claims: RecentClaim[]
   Items: RecentItem[]
 }
 
-export const recentChanges = writable<RecentChange[]>([])
+export const recentActivity = writable<RecentActivity>()
```
forevermatt, line 29: What was the advantage of splitting out a separate `recentActivity` store instead of only having a `recentChanges` store?

(If it's worth staying with this split--which it may well be--would you mind if we don't export `recentActivity`? It seems helpful to hide the messiness within this single file, and only expose the sorted list of `recentChanges` so that anywhere we use it, we always use that sorted list.)

src/components/CardsGrid.svelte
```diff
@@ -0,0 +1,77 @@
...
+  const aWeekAgo = new Date(today.getFullYear(), today.getMonth(), today.getDate() - 7)
+  return Number(updatedAt) > Number(aWeekAgo)
+}
+
+const isRecentOrIncomplete = (card: CardItem) => isIncomplete(card) || isRecent(card)
+
+const sortByUpdateDate = (cardA: CardItem, cardB: CardItem) =>
+  +new Date(cardB.data.updated_at) - +new Date(cardA.data.updated_at)
+
+const getRecentChanges = (changes: RecentChange[]): CardItem[] => {
```
forevermatt, line 37: (This comment only applies if it is still beneficial to have a separate `CardItem` type--see comment above):

Though this is named `getRecentChanges()`, it looks like it's actually being given `RecentChange`s and is getting back `CardItem`s.

What would you think of renaming it to indicate that what we're actually getting back is `CardItem`s?

src/data/claims.ts
forevermatt: Yeah, I think [the docs](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) that @RussellChamp linked to give a nice explanation of that, which helped me take a couple of steps forward in my understanding of this syntax.

src/components/CardsGrid.svelte
```diff
@@ -0,0 +1,77 @@
+<script lang="ts">
+import ClaimCard from './ClaimCard.svelte'
+import ItemCard from './ItemCard.svelte'
+import type { AccountablePersonOptions } from 'data/accountablePersons'
+import { Claim, ClaimItem, incompleteClaimItemStatuses, isClaimItem } from 'data/claims'
+import { incompleteItemCoverageStatus, PolicyItem } from 'data/items'
+import { isRecentClaim, RecentChange } from 'data/recent-activity'
+
+type CardItem = { data: ClaimItem | PolicyItem; claim?: Claim }
```
forevermatt, line 9: I'm still having trouble understanding the benefit of adding a new `CardItem` type, rather than just using `RecentChange[]` for the could-be-either type that we're passing around. Mind elaborating on that?

(I feel like we're getting enough types that are similar, but not quite identical, that it's getting harder to follow, and so I was wondering about reducing that.)

forevermatt (commented):
I'd like your input on a few things before this is merged.

src/components/CardsGrid.svelte
RussellChamp: the list of cards can be hydrated either through passing it the list of recent changes OR by passing it a list of claims and policy items. while i could mash the claims/items into the format of `RecentChange[]` I'd rather not since there are some values that I do not have (like 'StatusUpdatedAt').

src/data/items.ts
```diff
@@ -3,7 +3,19 @@ import { throwError } from '../error'
...
-export type ItemCoverageStatus = 'Draft' | 'Pending' | 'Approved' | 'Denied' | 'Revision' | 'Inactive'
+export enum ItemCoverageStatus {
+  Draft = 'Draft',
+  Pending = 'Pending',
+  Approved = 'Approved',
+  Denied = 'Denied',
+  Revision = 'Revision',
+  Inactive = 'Inactive',
+}
+export const incompleteItemCoverageStatus = [
```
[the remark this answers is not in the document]
RussellChamp: huh....... so i looked it up to make sure i did it right and it appears that 'status' can serve as both singular and plural. https://grammarhow.com/plural-of-status/ Thoughts on that?

src/data/recent-activity.ts
RussellChamp: in the examples, they are limiting the scope of their `isFooType` to only ever differentiate between the supplied union types. by using an `any` I am saying "pass me in anything at all and i'll tell you if it is a RecentClaim". this is more flexible but also potentially vulnerable to false positives since `{ Claim: 'foo', StatusUpdatedAt: 'bar', anythingElseIWant: 42 }` will also be classified as a 'RecentClaim'. i think this is a worthwhile tradeoff

src/data/recent-activity.ts
RussellChamp: while working on CardsGrid I initially was using the raw RecentActivity before switching over. part of the reasoning was to keep the logic of loadRecentActivity cleaner by moving data munging into a helper function (in this case, the derived state). an additional benefit is that we do not pay the computational cost of the data munging on retrieval time but only when there is an actual subscriber to the data store.
but i can appreciate the desire to expose a unified interface and will remove the export on recentActivity

src/components/CardsGrid.svelte
forevermatt: What if we simply extracted the PolicyItems and Claims from the RecentChanges, so that in both cases we're simply passing in PolicyItems and Claims?

RussellChamp: because RecentChanges has already mashed the two together into a single array

src/data/dependents.ts
```diff
@@ -25,6 +25,9 @@ export type UpdatePolicyDependentRequestBody = {
 }
 
 export const dependentsByPolicyId = writable<{ [policyId: string]: PolicyDependent[] }>({})
+export const allPolicyDependents = derived(dependentsByPolicyId, (dependentsByPolicyId) => {
+  return Object.values(dependentsByPolicyId).flat()
```
hobbitronics, line 29:
```suggestion
  return Object.values($dependentsByPolicyId).flat()
```

src/data/dependents.ts
```diff
@@ -25,6 +25,9 @@ export type UpdatePolicyDependentRequestBody = {
 }
 
 export const dependentsByPolicyId = writable<{ [policyId: string]: PolicyDependent[] }>({})
+export const allPolicyDependents = derived(dependentsByPolicyId, (dependentsByPolicyId) => {
```
hobbitronics, line 28:
```suggestion
export const allPolicyDependents = derived(dependentsByPolicyId, ($dependentsByPolicyId) => {
```

src/data/recent-activity.ts
```diff
@@ -1,42 +1,56 @@
...
+export const recentChanges = derived(recentActivity, (recentActivity) => {
```
hobbitronics, line 30:
```suggestion
export const recentChanges = derived(recentActivity, ($recentActivity) => {
```

src/data/recent-activity.ts
```diff
@@ -1,42 +1,56 @@
...
+  if (!recentActivity) return []
```
hobbitronics, line 31:
```suggestion
  if (!$recentActivity) return []
```

src/data/policies.ts
```diff
@@ -46,7 +46,7 @@ export type UpdatePolicyRequestBody = {
 
 export const policies = writable<Policy[]>([])
 export const selectedPolicy = derived([policies, selectedPolicyId], ([policies, selectedPolicyId]) => {
```
hobbitronics, line 48: when other stores are referenced in a derived store they should be prepended with a $
```suggestion
export const selectedPolicy = derived([policies, selectedPolicyId], ([$policies, $selectedPolicyId]) => {
```

src/data/policies.ts
```diff
@@ -46,7 +46,7 @@ export type UpdatePolicyRequestBody = {
...
-  return policies.find((p) => p.id === selectedPolicyId) || {}
+  return policies.find((p) => p.id === selectedPolicyId) || ({} as Policy)
```
hobbitronics, line 49:
```suggestion
  return $policies.find((p) => p.id === $selectedPolicyId) || ({} as Policy)
```

src/data/recent-activity.ts
```diff
@@ -1,42 +1,56 @@
...
+/**
+ * Load the list of recent activity.
+ *
+ * @export
+ */
+export async function loadRecentActivity(): Promise<void> {
+  const urlPath = 'steward/recent'
+  const result = await GET<RecentActivity>(urlPath)
+
+  recentActivity.set(result)
```
hobbitronics, line 55: I am confused why this chunk of code got moved down here.

hobbitronics (changes requested):
Looks like some new (to me) typescript magic and I see you used uniq from lodash. Fun stuff. I noticed the derived stores didn't have $ when referencing other stores so I made suggestions to fix that. Otherwise looks good.

src/data/recent-activity.ts
```diff
@@ -1,42 +1,56 @@
...
 import type { PolicyItem } from './items'
-import { writable } from 'svelte/store'
+import { derived, writable } from 'svelte/store'
 
 export type RecentClaim = {
   Claim: Claim
   StatusUpdatedAt: string /* Date */
 }
+export function isRecentClaim(obj: any): obj is RecentClaim {
+  return !!((obj as RecentClaim)?.Claim && (obj as RecentClaim)?.StatusUpdatedAt)
```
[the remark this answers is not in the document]
RussellChamp: we're basically saying "if the random object we were passed has a 'Claim' and a 'StatusUpdatedAt' then it's a duck, er, it's a RecentClaim.

src/data/dependents.ts
RussellChamp: addressed all the instances

src/data/recent-activity.ts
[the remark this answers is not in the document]
RussellChamp: in this case we are trying to determine if the arbitrary passed in object is, in fact, a 'RecentClaim' or not. they are guilty until proven innocent so 'obj' seemed more appropriate.

src/data/recent-activity.ts
RussellChamp: i basically moved the data munging out of the loadRecentActivity function

src/data/claims.ts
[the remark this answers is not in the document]
hobbitronics: ahh thanks.

src/data/recent-activity.ts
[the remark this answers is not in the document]
hobbitronics: ok ok, I see now.

src/data/recent-activity.ts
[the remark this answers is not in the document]
hobbitronics: oh cool. Makes perfect sense when I look at it now.

src/data/recent-activity.ts
hobbitronics: ok, I was being confused by the PR review view I was using in github.

src/components/CardsGrid.svelte
```diff
@@ -0,0 +1,78 @@
...
+
+export let recentChanges: RecentChange[] = []
+export let claims: Claim[] = []
+export let policyItems: PolicyItem[] = []
+export let accountablePersons: AccountablePersonOptions[] = []
+export let isAdmin: boolean = false
+
+$: items = recentChanges.length ? parseRecentChanges(recentChanges) : parseClaimsAndPolicyItems(claims, policyItems)
```
forevermatt, line 17: Would you be okay with renaming this `items` to `cardItems`? The rest of the codebase uses `items` to refer to PolicyItems, and I think it will be helpful to clarify here that this is CardItems.

src/data/recent-activity.ts
forevermatt: That works

hobbitronics (approved):
Thanks for clarifying some things for me. Fun fact, if we weren't using lodash we could still use `uniqueArr = [...new Set(arr)]` instead of uniq. Not sure which is faster.

src/components/CardsGrid.svelte
```diff
@@ -0,0 +1,78 @@
...
+import ClaimCard from './ClaimCard.svelte'
+import ItemCard from './ItemCard.svelte'
+import type { AccountablePersonOptions } from 'data/accountablePersons'
+import { Claim, ClaimItem, incompleteClaimItemStatuses, isClaimItem } from 'data/claims'
+import { incompleteItemCoverageStatuses, PolicyItem } from 'data/items'
+import { isRecentClaim, RecentChange } from 'data/recent-activity'
+
+type CardItem = { data: ClaimItem | PolicyItem; claim?: Claim }
+
+export let recentChanges: RecentChange[] = []
```
forevermatt, line 11: This can totally be a change for a later PR, but I'd like us to push the `RecentChanges` concept back up out of this `CardsGrid` component, to decouple those two concepts.

This CardsGrid is great, and it simply needs to have a list of claims (technically, claim items with some claim data) and a list of policy items in order to do it's thing.

At the moment, we're using the RecentChanges data to as an available source of that data for admins, which we would otherwise lack right now. That works, but I think it would help reduce coupling and complexity to avoid letting that implementation detail (that we're using recent changes) leak into this component's code.

src/data/items.ts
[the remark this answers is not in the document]
forevermatt: Let's go with "statuses", as it's unambiguous.

PR #281, opened 2021-11-08
remove for of loop

hobbitronics:
I realized after one of recent PR's I could eliminate the for of loop


forevermatt (approved):
Wow, nice!

RussellChamp (changes requested):
This isn't quite equivalent because your new code will dispatch an event with an undefined value if choice isn't found in options. if that is not the intent, add a quick guard.

RussellChamp (approved):

PR #282, opened 2021-11-08
Fix "/home" redirects

forevermatt:
### Fixed
- Wait to set initial role selection until we have the user's actual role
- Send admins to the admin home page by default

### Added
- Default the customer to a corporate policy if they have one, household if not


hobbitronics (approved):
Looks good

PR #284, opened 2021-11-09
add freshdesk widget and types

RussellChamp:
* adds the freshdesk widget along with some hand-written types
* populates the feedback form with the current user's name and email


src/components/FreshdeskWidget.svelte
```diff
@@ -0,0 +1,27 @@
...
+  window.fwSettings = { widget_id: 73000000445 }
+  // @ts-ignore
+  // prettier-ignore
+  !function(){if("function"!=typeof window.FreshworksWidget){var n=function(){n.q.push(arguments)};n.q=[],window.FreshworksWidget=n}}()
+})
+
+$: setUserData($user)
+
+const setUserData = (user: User) => {
+  if (user && typeof FreshworksWidget !== 'undefined') {
```
RussellChamp, line 16: this was necessary because our empty user store loads before the FreshworksWidget is setup.
There are probably timing hisenbugs hidden here but the works case scenario is that the widget doesn't get properly pre-populated with the current user name and email.

src/customTypes/freshdesk.d.ts
```diff
@@ -0,0 +1,68 @@
...
+  status?: Status[]
+  group_id?: number[]
+  type?: string[]
+}
+
+type OpenArticleOptions = {
+  id: number
+}
+
+declare function FreshworksWidget(command: 'hide', target?: 'launcher'): void
```
RussellChamp, line 55: i read through their entire widget usage API documentation and this seems like all the commands they support. i got lazy near the end but i don't expect we're actually gonna call a lot of these.

src/components/FreshdeskWidget.svelte
```diff
@@ -0,0 +1,27 @@
+<script lang="ts">
+import user, { User } from '../authn/user'
+import { onMount } from 'svelte'
+
+onMount(() => {
+  // Here be dragons
+  window.fwSettings = { widget_id: 73000000445 }
```
forevermatt, line 7: It would probably be better to pass in the FreshDesk Widget ID via an environment variable.

src/components/FreshdeskWidget.svelte
```diff
@@ -0,0 +1,27 @@
...
+    FreshworksWidget('identify', 'ticketForm', {
+      name: user?.first_name,
+      email: user?.email_override || user?.email,
+    })
+  }
+}
+</script>
+
+<svelte:head>
+  <script type="text/javascript" src="https://widget.freshworks.com/widgets/73000000445.js" async defer></script>
```
forevermatt, line 26: Note that the Widget ID (if made dynamic) would need to be injected here, too.

src/customTypes/freshdesk.d.ts
```diff
@@ -0,0 +1,68 @@
+declare interface Window {
+  fwSettings: {
+    widget_id: number
+    locale?: string
+  }
+}
+
+type IdentifyOptions = { name: string; email: string }
+enum Priority {
+  Low = 1,
```
forevermatt, line 10: Does this syntax then cause the subsequent entries to have consecutive values (e.g. Medium: 2, etc.)?

src/customTypes/freshdesk.d.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+type OpenArticleOptions = {
+  id: number
+}
+
+declare function FreshworksWidget(command: 'hide', target?: 'launcher'): void
+declare function FreshworksWidget(command: 'show', target?: 'launcher'): void
+declare function FreshworksWidget(command: 'open', target?: 'ticketForm'): void
+declare function FreshworksWidget(command: 'close' | 'destory' | 'boot'): void
```
forevermatt, line 58:
```suggestion
declare function FreshworksWidget(command: 'close' | 'destroy' | 'boot'): void
```

forevermatt (changes requested):

src/customTypes/freshdesk.d.ts
RussellChamp: yes https://www.typescriptlang.org/docs/handbook/enums.html

.env.example
```diff
@@ -1,3 +1,4 @@
 GA_MEASUREMENT_ID=G-XXXXXXXXXX
+FRESHDESK_ID=73000000445
```
forevermatt, line 2: Sorry to ask for one more change, but let's avoid putting any real IDs in our repo. Would you mind doing something similar with this to what we did with the Google Analytics ID, above?

forevermatt (approved):

PR #287, opened 2021-11-09
remove delete button from FilePreview for now

hobbitronics:
- remove the button and show file type to customer instead
- comment out code not being used yet

### DIFF
--- a/src/components/FilePreview/FilePreview.svelte
+++ b/src/components/FilePreview/FilePreview.svelte
@@ -33,12 +33,13 @@ const onClick = (id: string) => {
   dispatch('preview', id)
 }
 
-function onDelete(event: CustomEvent, id: string) {
-  event.preventDefault()
-  event.stopPropagation()
+//Todo bring back delete for 1.1
+// function onDelete(event: CustomEvent, id: string) {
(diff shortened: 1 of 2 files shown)


src/components/FilePreview/FilePreview.svelte
```diff
@@ -65,9 +66,7 @@ function onDelete(event: CustomEvent, id: string) {
         <p class="white my-0">{preview.file.name}</p>
         <p class="white my-0">{formatDate(preview.created_at)}</p>
       </div>
-      {#if isMemberOfPolicy}
-        <Button class="delete-button" raised on:click={(evt) => onDelete(evt, preview.id)}>Delete</Button>
-      {:else if preview.purpose}
+      {#if preview.purpose}
```
forevermatt, line 69: We used to only show this if `isMemberOfPolicy` was also false. Do we still want that condition here?

hobbitronics: I think showing the filetype to both roles is fine. Phillip had commented its more useful than the filename earlier

forevermatt (approved):

PR #288, opened 2021-11-09
Handle item-less policy display better

forevermatt:
### Added
- Show friendlier text and an "Add Item" button for itemless policies

### Fixed
- If we already have items, show them even while reloading the items data
- Update policy items page to use `$selectedPolicyId` store


RussellChamp (approved):
lgtm

PR #290, opened 2021-11-09
Fix policy-name code

forevermatt:
### Changed
- Only calculate a policy name client-side if none was provided by the API

### Fixed
- Update UI code to be consistent about how it obtains a policy's name

### DIFF
--- a/src/data/policies.ts
+++ b/src/data/policies.ts
@@ -112,12 +112,16 @@ export function clear(): void {
 
 export const getNameOfPolicy = (policy: Policy): string => {
   let policyName = ''
-  if (policy?.type === 'Team') {
-    policyName = policy?.name || 'Team'
-  } else if (policy?.type === 'Household') {
-    const members = policy?.members || []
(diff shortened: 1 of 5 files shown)


src/data/policies.ts
```diff
@@ -112,12 +112,16 @@ export function clear(): void {
...
-  if (policy?.type === 'Team') {
-    policyName = policy?.name || 'Team'
-  } else if (policy?.type === 'Household') {
-    const members = policy?.members || []
-    const lastName = members[0]?.last_name || ''
-    policyName = lastName + ' Household'
+  if (policy?.name) {
+    policyName = policy.name
+  } else {
+    if (policy?.type === 'Team') {
```
RussellChamp, line 118: make sure these changes don't overlap with your prior PR that uses the policy type enums

RussellChamp (changes requested):

RussellChamp (approved):

PR #291, opened 2021-11-09
Use policy type enums

forevermatt:
### Fixed
- Use `PolicyType` enum values for `policy.type` comparisons

---

**NOTE:** This leaves a few places unfixed because they will be removed in the policy-name-fix PR (#290), to reduce merge conflicts.


RussellChamp (approved):
lgmt. enums banzai!

PR #293, opened 2021-11-10
Feat/add claims route

RussellChamp:
add root-level claims route that redirects to default policy claims


src/helpers/routes.ts
```diff
@@ -14,6 +13,7 @@ export const CHAT = '/chat'
 
 export const CUSTOMER_HOME = '/customer/home'
 
+export const CLAMS = '/claims'
```
forevermatt, line 16:
```suggestion
export const CLAIMS = '/claims'
```

forevermatt (approved):
One minor fix, but otherwise looks good

forevermatt (approved):

hobbitronics (approved):
looks good.

PR #295, opened 2021-11-10
Actually call dependents API endpoints for edits and deletes

forevermatt:
### Fix
- Actually call the API when we edit a policy dependent
- Actually call the API when we delete a policy dependent


hobbitronics (approved):
Looks good.

RussellChamp (approved):
lgtm

PR #296, opened 2021-11-10
aligned numeric Datatable Items right

hobbitronics:
- The Header Item as well as Row Item needs to have `numeric` to align right.
- Material handles alignment as long as we do this.

### DIFF
--- a/src/components/RecentActivityTable.svelte
+++ b/src/components/RecentActivityTable.svelte
@@ -85,7 +85,7 @@ const getFormattedClaimItemPremium = (claim: Claim): string => {
       {/if}
     {:else}
       <DataRow>
-        <RowItem colspan="7">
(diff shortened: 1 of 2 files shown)


RussellChamp (approved):
lgtm

src/components/RecentActivityTable.svelte
```diff
@@ -85,7 +85,7 @@ const getFormattedClaimItemPremium = (claim: Claim): string => {
       {/if}
     {:else}
       <DataRow>
-        <RowItem colspan="7">
+        <RowItem colspan={7}>
```
forevermatt, line 88: Was this change intentional?

hobbitronics: yes, colspan takes a number

PR #297, opened 2021-11-10
remove accountablePersons from most components

RussellChamp:
The response body for PolicyItems has changed such that 'accountable_person' now contains both the id and user-friendly name of the accountable person (instead of just an id). This means that we can rip up TONS of now-dead code that was doing fangaly data loads and look-up in order to retrieve this data.
The remaining places where we still need to do a load'n'look-up for policy dependent/member data is in the following cases
* when we are creating a new policy item
* when we are editing a policy item
* when we are editing policy settings and policy dependent settings

### DIFF
--- a/src/pages/policies/[policyId]/items.svelte
+++ b/src/pages/policies/[policyId]/items.svelte
@@ -2,32 +2,22 @@
 import { CardsGrid, ItemsTable, Row } from 'components'
 import { isLoadingPolicyItems, loading } from 'components/progress'
 import { Claim, loadClaimsByPolicyId, selectedPolicyClaims } from 'data/claims'
-import { AccountablePersonOptions, getDependentOptions, getPolicyMemberOptions } from 'data/accountablePersons'
-import { dependentsByPolicyId, loadDependents } from 'data/dependents'
 import { deleteItem, loadItems, PolicyItem, selectedPolicyItems } from 'data/items'
 import { getNameOfPolicy, selectedPolicy } from 'data/policies'
-import { loadMembersOfPolicy, membersByPolicyId } from 'data/policy-members'
 import { selectedPolicyId } from 'data/role-policy-selection'
 import * as routes from 'helpers/routes'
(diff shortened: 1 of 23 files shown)


src/pages/policies/[policyId]/items.svelte
```diff
@@ -2,32 +2,22 @@
...
 import { deleteItem, loadItems, PolicyItem, selectedPolicyItems } from 'data/items'
 import { getNameOfPolicy, selectedPolicy } from 'data/policies'
-import { loadMembersOfPolicy, membersByPolicyId } from 'data/policy-members'
 import { selectedPolicyId } from 'data/role-policy-selection'
 import * as routes from 'helpers/routes'
 import { formatPageTitle } from 'helpers/pageTitle'
 import { goto, metatags } from '@roxi/routify'
 import { Button, Page } from '@silintl/ui-components'
-
-let policyId: string
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items.svelte:16]: Overall, I love these changes, but why this declaration of variable and type?

src/pages/policies/[policyId]/items.svelte
```diff
@@ -2,32 +2,22 @@
...
 $: policyId = $selectedPolicyId
 
-$: policyId && loadItems(policyId)
-$: policyId && loadClaimsByPolicyId(policyId)
-
-$: policyId && loadDependents(policyId)
-$: dependents = $dependentsByPolicyId[policyId] || []
-$: dependentOptions = getDependentOptions(dependents)
+onMount(() => {
+  loadItems(policyId)
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items.svelte:17]: I don't know, but I wonder if we're going to hit a situation where `policyId` doesn't have a value by the time these calls are made. Something to keep in mind if we ever hit weird errors with this.

forevermatt (approved):

src/pages/policies/[policyId]/items.svelte
Comment (RussellChamp) [reply src/pages/policies/[policyId]/items.svelte:16]: this was redundant with the line below

PR #300, opened 2021-11-11
Filedrop alignment and margin

hobbitronics:
The button had top margin throwing off alignment
When flex items were scrunched up they had no margin
switch to flex row for mobile size screens


RussellChamp (approved):
lgtm

PR #304, opened 2021-11-11
Remove `Review3a`/`Review3b` and update claim statuses awaiting admins

forevermatt:
### Changed
- Remove `Review3a` and `Review3b` statuses
- Update which claims statuses we show as awaiting each admin role

The new design is that Steward can do Review1 and Review2, and Signator can do Review3.

---

@hobbitronics, I did not modify the ClaimActions component, since it sounded like you would be handling those changes and they looked more involved that just removing entries from a list.

### DIFF
--- a/src/data/claims.ts
+++ b/src/data/claims.ts
@@ -10,9 +10,7 @@ export type ClaimItemStatus =
   | 'Draft'
   | 'Review1'
   | 'Review2'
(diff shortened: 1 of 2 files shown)


hobbitronics (approved):
Looks good. I will take care of the ClaimActions piece of the puzzle.

PR #305, opened 2021-11-11
added Team item details

hobbitronics:
- show Team details when policy.type === 'Team'

### DIFF
--- a/src/components/B.svelte
+++ b/src/components/B.svelte
@@ -1,1 +1,0 @@
-<div class="mdc-bold-font {$$props.class}"><slot /></div>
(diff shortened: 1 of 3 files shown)


src/components/B.svelte
```diff
@@ -0,0 +1 @@
+<div class="mdc-bold-font {$$props.class}"><slot /></div>
```
RussellChamp, line 1: this seems like unnecessary complications. what's wrong with just directly adding this class to the components you want? it doesn't even cover all use cases (eg if you wanted an inline bold element you would want a span)

RussellChamp (changes requested):

src/components/B.svelte
hobbitronics: yeah, I suppose so. I was just having problems using `<b>` and getting margins to cooperate, but we can just add the class.

PR #306, opened 2021-11-11
show approve or preapprove to appropriate admin type

hobbitronics:
- Signator and Steward have different "approve" actions at different states

### DIFF
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -1,6 +1,13 @@
 <script lang="ts">
 import { UserAppRole } from '../authn/user'
-import { Claim, ClaimStatus, editableStatuses, PayoutOption } from 'data/claims'
+import {
+  Claim,
+  ClaimStatus,
+  editableStatuses,
+  PayoutOption,
+  statusesAwaitingSignator,
+  statusesAwaitingSteward,
+} from 'data/claims'
 import { roleSelection } from 'data/role-policy-selection'
 import Description from './Description.svelte'
 import { throwError } from '../error'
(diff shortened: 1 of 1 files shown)


src/components/ClaimActions.svelte
```diff
@@ -45,6 +52,9 @@ $: switch (status) {
 
 $: isEditable = editableStatuses.includes(status)
 $: showSubmit = ['Receipt', 'Revision'].includes(status) || (status === 'Draft' && needsFile)
+$: showLowerButtons =
```
RussellChamp, line 55: is there a more descriptive name we can give this? like 'showActionButton'

src/components/ClaimActions.svelte
```diff
@@ -82,7 +92,7 @@ const onDeny = () => dispatch('deny', message)
 </style>
 
 {#if isAdmin}
-  {#if ['Review1', 'Review2', 'Review3', 'Review3a', 'Review3b'].includes(status)}
+  {#if ['Review1', 'Review2', 'Review3'].includes(status)}
```
RussellChamp, line 95: it would be nice to get these arrays out of the view templates, but that is probably another PR

RussellChamp (changes requested):

src/components/ClaimActions.svelte
hobbitronics: they are all action buttons, but this is either preapprove or approve so we could just use `showApprovalButtons`

src/components/ClaimActions.svelte
hobbitronics: I couldn't find them already in a const or enum yet, but I agree

PR #307, opened 2021-11-11
use "I have reviewed and agree to..."

hobbitronics:
- update the agree to terms verbiage.

### DIFF
--- a/src/components/Checkout.svelte
+++ b/src/components/Checkout.svelte
@@ -64,7 +64,9 @@ const handleDialog = (event: CustomEvent<string>) => {
 <div class="flex align-items-center my-2">
   <Checkbox on:checked={() => (checked = true)} on:unchecked={() => (checked = false)} />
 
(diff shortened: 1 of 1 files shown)


src/components/Checkout.svelte
```diff
@@ -64,7 +64,9 @@ const handleDialog = (event: CustomEvent<string>) => {
 <div class="flex align-items-center my-2">
   <Checkbox on:checked={() => (checked = true)} on:unchecked={() => (checked = false)} />
 
-  <span class="mr-3px">I have read and agree to</span><a target="_blank" href={TERMS_OF_SERVICE}> Terms of Service</a>
+  <span class="mr-3px">I have reviewed and agree to</span><a target="_blank" href={TERMS_OF_SERVICE}>
```
RussellChamp, line 67: kinda weird spacing on those tags but ok

RussellChamp (approved):
lgtm

src/components/Checkout.svelte
hobbitronics: do you mean that they are on the same line or that it is using margin to make a space?

PR #308, opened 2021-11-11
add customer items page

hobbitronics:
- Added home page with active items only
- Show active and inactive items on items page
- Admin is not affected by these changes


RussellChamp (approved):
nice

PR #309, opened 2021-11-12
add session resume feature

RussellChamp:
* keep track of the last route the user has visited and the time they visited
* the first time we arrive at the default home page (which is where the post-login process will drop a user) check if we should resume the prior "session"

### DIFF
--- a/src/authn/session.ts
+++ b/src/authn/session.ts
@@ -1,0 +1,20 @@
+import * as routes from 'helpers/routes'
+import { writable } from 'svelte/store'
+
+export const initialized = writable<boolean>(false)
+
+export const getLastPath = (): { path: string; lastSet: number } => {
+  return {
+    path: localStorage.getItem('last-path') || '',
+    lastSet: Number(localStorage.getItem('last-path-set')),
+  }
+}
+
+const lastPathBlacklist = [routes.ROOT, routes.HOME, routes.LOGOUT, routes.LOGGEDOUT]
+export const setLastPath = (path: string): void => {
+  if (!lastPathBlacklist.includes(path)) {
+    initialized.set(true)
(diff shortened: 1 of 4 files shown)


src/authn/session.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const getLastPath = (): { path: string; lastSet: number } => {
+  return {
+    path: localStorage.getItem('last-path') || '',
+    lastSet: Number(localStorage.getItem('last-path-set')),
+  }
+}
+
+const lastPathBlacklist = [routes.ROOT, routes.HOME, routes.LOGOUT, routes.LOGGEDOUT]
+export const setLastPath = (path: string): void => {
```
RussellChamp, line 14: we do NOT want to "resume" to certain routes but we DO want to track that an attempt was made.
say the user was on the settings page for 3 hours. when they come back to that same session, they will be logged out but we will want to resume this "stale" session

src/pages/home.svelte
```diff
@@ -1,10 +1,23 @@
...
 import { roleSelection } from 'data/role-policy-selection'
 
-$: sendToRoleHome($roleSelection)
+$: {
+  const lastPath = getLastPath()
+  // if the last path was set withing the last 5 minutes assume we want to resume the session
+  // only ever redirect to the last path once per session
+  const shouldResumeSession = lastPath && !$initialized && isBefore(Date.now(), add(lastPath.lastSet, { minutes: 5 }))
+  if (shouldResumeSession) {
+    initialized.set(true)
```
hobbitronics, line 15:
```suggestion
    $initialized = true
```

hobbitronics (changes requested):
interesting, just one change requested.

hobbitronics (approved):
Nice

PR #312, opened 2021-11-12
filter recent PolicyItem cards with isIncomplete

hobbitronics:
- This removes "Inactive" ItemCards from the Admin home page


RussellChamp (approved):
i'm not sure how i feel about doing too much data munging in a display component. however, this solve a "now" problem and is probably good enough. we can generic-ize this component if the need ever arises

PR #313, opened 2021-11-15
add country to AccountablePerson

RussellChamp:
use the country value found on accountable_person by default


hobbitronics (approved):
LOoks GoOd

PR #314, opened 2021-11-15
Accountable Person -> Assigned To

hobbitronics:
We are going with "Assigned To" instead of "Accountable Person"


RussellChamp (approved):
Approved To

PR #315, opened 2021-11-15
set default policies limit to 20, search to 50

hobbitronics:
- added a `limit` parameter  to `loadPolicies` and `searchPoliciesFor` and defaulted to '20'

### DIFF
--- a/src/data/policies.ts
+++ b/src/data/policies.ts
@@ -143,7 +143,7 @@ export async function loadPolicy(policyId: string): Promise<Policy> {
   return response
 }
 
-export async function searchPoliciesFor(searchText: string, limit = '50'): Promise<Policy[]> {
+export async function searchPoliciesFor(searchText: string, limit = 50): Promise<Policy[]> {
   const queryString = qs.stringify({ search: searchText, limit })
   const response = await GET<{ data: Policy[]; meta: any }>(`policies?${queryString}`)
   return response.data
--- a/src/data/policies.ts
+++ b/src/data/policies.ts
@@ -127,7 +127,7 @@ export const getNameOfPolicy = (policy: Policy): string => {
 }
 
 //claims or members/dependents fields from this endpoint are deprecated
-export async function loadPolicies(limit = '20'): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/data/policies.ts
```diff
@@ -127,8 +127,9 @@ export const getNameOfPolicy = (policy: Policy): string => {
 }
 
 //claims or members/dependents fields from this endpoint are deprecated
-export async function loadPolicies(): Promise<void> {
-  const response = await GET<{ data: Policy[]; meta: any }>('policies')
+export async function loadPolicies(limit = '20'): Promise<void> {
```
RussellChamp, line 130: the `limit` parameter should be a number instead of a string. i expect it should be stringified properly by qs

src/data/policies.ts
```diff
@@ -142,8 +143,8 @@ export async function loadPolicy(policyId: string): Promise<Policy> {
   return response
 }
 
-export async function searchPoliciesFor(searchText: string): Promise<Policy[]> {
-  const queryString = qs.stringify({ search: searchText })
+export async function searchPoliciesFor(searchText: string, limit = '50'): Promise<Policy[]> {
```
RussellChamp, line 146: the `limit` parameter should be a number instead of a string. i expect it should be stringified properly by qs

src/data/policies.ts
RussellChamp, line 130: the `limit` parameter should be a number instead of a string. i expect it should be stringified properly by qs

RussellChamp (changes requested):

src/data/policies.ts
hobbitronics:
```suggestion
export async function searchPoliciesFor(searchText: string, limit = 50): Promise<Policy[]> {
```

src/data/policies.ts
hobbitronics:
```suggestion
export async function loadPolicies(limit = 20): Promise<void> {
```

forevermatt (approved):

PR #316, opened 2021-11-15
was not redirecting appRole: Admin to correct url

hobbitronics:
- found a missing case


RussellChamp (approved):
lgtm

forevermatt (approved):
I think that "Admin" role is ambiguous enough that we (at one point) were considering removing it, but I see no harm in this change for as long as it's still here.

hobbitronics:
> I think that "Admin" role is ambiguous enough that we (at one point) were considering removing it, but I see no harm in this change for as long as it's still here.

The problem is that the api was giving me "Admin" as my user.app_role so as long as its possible we should handle it.

PR #318, opened 2021-11-15
fix(Card): Allow each policy card in each row to auto expand it's height

rdotg:
Now each card will auto expand to match the largest sibling's height in same row:

image


hobbitronics (approved):
Nice, I didn't realize it would be that easy.

forevermatt (approved):

PR #320, opened 2021-11-16
removed faq, Items and Claims from Admin roles

hobbitronics:
- remove Drawer menu items for admins as they all go to admin/home and cause the wrong item to be highlighted
- remove faq

- we may want to show a different view for them at some point but for now they seem to serve no purpose

### DIFF
--- a/src/pages/_layout.svelte
+++ b/src/pages/_layout.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
-import user, { isAdmin, UserAppRole } from '../authn/user'
+import user, { isAdmin } from '../authn/user'
 import { AppDrawer } from 'components'
-import { initialized as policiesInitialized, loadPolicies, policies, Policy } from 'data/policies'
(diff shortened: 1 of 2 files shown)


src/pages/_layout.svelte
```diff
@@ -33,16 +32,19 @@ $: menuItems = [
     url: routes.ITEMS,
     icon: 'umbrella',
     label: 'items',
+    hide: inAdminRole,
   },
   {
     url: inAdminRole ? routes.ADMIN_HOME : routes.customerClaims(policyId),
```
RussellChamp, line 38: if we're going to remove this from the 'inAdminRole' then remove the admin_home option from here

RussellChamp (changes requested):

src/pages/_layout.svelte
forevermatt, line 38:
```suggestion
    url: routes.customerClaims(policyId),
```

RussellChamp (approved):

PR #321, opened 2021-11-16
rename to Policy Settings and hide from Admin

hobbitronics:
- It seems we don't the Policy Settings for Admin

### DIFF
--- a/src/pages/settings.svelte
+++ b/src/pages/settings.svelte
@@ -4,6 +4,5 @@ import { settingsPolicy } from 'helpers/routes'
 import { redirect } from '@roxi/routify'
 import { selectedPolicyId } from 'data/role-policy-selection'
 
-$: $user.policy_id &&
-  ($selectedPolicyId ? $redirect(settingsPolicy($selectedPolicyId)) : $redirect(settingsPolicy($user.policy_id)))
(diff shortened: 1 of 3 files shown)


src/pages/settings.svelte
```diff
@@ -2,6 +2,7 @@
 import user from '../authn/user'
 import { settingsPolicy } from 'helpers/routes'
 import { redirect } from '@roxi/routify'
+import { selectedPolicyId } from 'data/role-policy-selection'
 
-$: $user.policy_id && $redirect(settingsPolicy($user.policy_id))
+$: $selectedPolicyId ? $redirect(settingsPolicy($selectedPolicyId)) : $redirect(settingsPolicy($user.policy_id))
```
RussellChamp, line 7: by not checking for policy_id this _might_ introduce a heisenbug when policy_id doesn't yet exist when a user enters the app for the very first time (because of how the BE does stuff). extremely rare case, though.

RussellChamp (approved):
probably good enough. possible heisenbug

src/pages/settings.svelte
hobbitronics: good point, i added some protection against that.

PR #322, opened 2021-11-16
stay on current page when changing role/policy

hobbitronics:
- this allows the customer/admin to stay on a claim/item but change roles or stay on settings and change policy
- the customer will not be able to view the same claim/item when switching policies so will be sent to home

### DIFF
--- a/src/components/RoleAndPolicyMenu.svelte
+++ b/src/components/RoleAndPolicyMenu.svelte
@@ -65,7 +65,7 @@ const getHouseholdEntries = (policies: Policy[]): MenuItem[] => {
 
 const selectRole = (role: UserAppRole) => {
   recordRoleSelection(role)
-  dispatch('role', role)
+  dispatch('role', { role, policyId: $selectedPolicyId })
 }
 
 const getEntriesForRole = (role: UserAppRole): MenuItem[] => {
--- a/src/pages/_layout.svelte
+++ b/src/pages/_layout.svelte
@@ -3,7 +3,7 @@ import user, { isAdmin } from '../authn/user'
 import { AppDrawer } from 'components'
 import { initialized as policiesInitialized, loadPolicies } from 'data/policies'
 import * as routes from 'helpers/routes'
-import { goto } from '@roxi/routify'
+import { goto, params, route, url } from '@roxi/routify'
 import { roleSelection, selectedPolicyId } from 'data/role-policy-selection'
 
 // TODO: Avoid trying to load the policies until the user has authenticated (to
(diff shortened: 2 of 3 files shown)


src/components/RoleAndPolicyMenu.svelte
```diff
@@ -65,7 +65,7 @@ const getHouseholdEntries = (policies: Policy[]): MenuItem[] => {
 
 const selectRole = (role: UserAppRole) => {
   recordRoleSelection(role)
-  dispatch('role', role)
+  dispatch('role', { role, policyId: $selectedPolicyId })
```
forevermatt, line 68: What's the advantage of including the policyId in the event data, rather than just using `$selectedPolicyId` in the code listening for this event?

src/pages/_layout.svelte
```diff
@@ -66,11 +66,33 @@ $: menuItems = [
...
+}
+const goToAdminView = (event: CustomEvent) => {
+  if ($params.policyId && ($params.claimId || $params.itemId)) {
+    const claimOrItemIdObj = $params.claimId ? { claimId: $params.claimId } : { itemId: $params.itemId }
+    gotoPath(event.detail.policyId, claimOrItemIdObj)
+  } else if ($params.policyId) {
+    gotoPath(event.detail.policyId)
+  } else {
+    $goto(routes.ADMIN_HOME)
+  }
```
forevermatt, line 92: It seems like it might simplify the code some here to check for URL params more individually:
```suggestion
  if ($params.policyId) {
    if ($params.claimId) {
      gotoPath(event.detail.policyId, { claimId: $params.claimId })
    } else if ($params.itemId) {
      gotoPath(event.detail.policyId, { itemId: $params.itemId })
    } else {
      gotoPath(event.detail.policyId)
    }
  } else {
    $goto(routes.ADMIN_HOME)
  }
```

src/pages/_layout.svelte
```diff
@@ -66,11 +66,33 @@ $: menuItems = [
     hide: inAdminRole || !policyId,
   },
 ]
+const isCustomerOnOwnPolicy = (policyId: string) => policyId === $selectedPolicyId
 
-const goToPolicyAsCustomer = (event: CustomEvent) => $goto(routes.policyHome(event.detail))
-const goToAdminHome = () => $goto(routes.ADMIN_HOME)
+const gotoPath = (policyId: string, claimOrItemIdObj = {}) =>
+  $goto($url($route.path, { policyId, ...claimOrItemIdObj }))
```
forevermatt, line 72: `$goto()` will handle the same parameters as `$url()`, so you can remove the call to `$url()` here:
```suggestion
  $goto($route.path, { policyId, ...claimOrItemIdObj })
```

src/pages/_layout.svelte
```diff
@@ -66,11 +66,33 @@ $: menuItems = [
...
+const goToCustomerView = (event: CustomEvent) => {
+  if ($params.policyId && !$params.claimId && !$params.itemId) {
+    gotoPath(event.detail)
+  } else if ($params.policyId && ($params.claimId || $params.itemId) && isCustomerOnOwnPolicy(event.detail)) {
+    const claimOrItemIdObj = $params.claimId ? { claimId: $params.claimId } : { itemId: $params.itemId }
+    gotoPath(event.detail, claimOrItemIdObj)
+  } else {
+    $goto(routes.policyHome(event.detail))
+  }
+}
```
forevermatt, line 83: I'm having trouble following the logic in this method. Are there ways you could clarify what's going on here?

We may also just need better defined requirements of what-should-happen-when, and then we can extract these if-conditions into methods named accordingly.

I don't know... maybe you can just talk me through this logic, and that might help.

src/components/RoleAndPolicyMenu.svelte
hobbitronics: this is the "old" policy as the url hasn't changed yet, and I need to compare it to the "new" one

src/pages/_layout.svelte
hobbitronics: sure, the first condition is for urls with only policyId (e.g. policy/settings),
the second is for viewing a claim or item (switching from Admin to Customer) but not when customer is trying to view the wrong policy,
The third as any url with no params.

src/components/RoleAndPolicyMenu.svelte
hobbitronics: `isCustomerOnOwnPolicy()` compares the new `$selectedPolicy` to the old one.

forevermatt: Since this is dispatching the value of `$selectedPolicyId` and the code handling that dispatched event seems like it would be invoked immediately (before `$selectedPolicyId` has changed, since it's before any calls to `$goto()`), it seems like the value would still be the same... or am I missing something?

forevermatt (approved):
I think there are some aspects of that may be worth fine-tuning later, but it seems like it basically accomplishes the desired purpose, so let's get this merged.

src/components/RoleAndPolicyMenu.svelte
hobbitronics: I can't remember why/where/how but `$selectedPolicyId` is updated before going to $goto() is called.

PR #323, opened 2021-11-16
Validate country

forevermatt:
### Added
- Use Google Places Autocomplete API for a user's personal settings "country" input


src/components/LocationInput.svelte
```diff
@@ -0,0 +1,46 @@
...
+const options = {
+  fields: ['address_components'],
+  types: ['(regions)'],
+}
+const placeholder = 'Enter country'
+
+/* This intermediate (internal) value is to cause the GooglePlacesAutocomplete
+ * component to be updated when our incoming `value` parameter is changed.
+ * Otherwise, this would continue showing the selected value rather than the
+ * value we dispatched. */
```
hobbitronics, line 18: I'm don't get the mechanics of how this is different than passing value directly to the googlePlacesApiKey, but trust it works.

hobbitronics: or maybe I am thinking if you had done a bind:value

forevermatt: I think it has to do with the way the Google Places Autocomplete library handles updates to the actual `<input />` element, but I'm not sure. Either way, when I pass in the `value` directly, the input field continues to show the value actually selected by the user (which can be more specific than a country), rather than the country data we extracted from their selection and are actually using.

PR #324, opened 2021-11-17
show policy to customer

hobbitronics:
- add a link to policy details page for customers
- when switching role to Admin prevent viewing pages they should not be on

### DIFF
--- a/src/helpers/routes.ts
+++ b/src/helpers/routes.ts
@@ -46,8 +46,8 @@ export const householdSettingsNewDependent = (policyId: string) => `/policies/${
 export const TERMS_OF_SERVICE = '/terms'
 export const PRIVACY_POLICY = '/privacy'
 
-export const pathIsforAdmins = (path: string): boolean =>
-  ![
+export const pathIsforCustomerOnly = (path: string): boolean =>
+  [
     '/policies/:policyId/settings',
     '/policies/:policyId/items/new',
     '/policies/:policyId/items/:itemId/edit',
--- a/src/helpers/routes.ts
+++ b/src/helpers/routes.ts
@@ -45,3 +45,5 @@ export const householdSettingsNewDependent = (policyId: string) => `/policies/${
 
 export const TERMS_OF_SERVICE = '/terms'
 export const PRIVACY_POLICY = '/privacy'
+
+export const routeIsNotPolicySettings = (path: string): boolean => path !== '/policies/:policyId/settings'
--- a/src/helpers/routes.ts
+++ b/src/helpers/routes.ts
@@ -45,16 +45,3 @@ export const householdSettingsNewDependent = (policyId: string) => `/policies/${
 
 export const TERMS_OF_SERVICE = '/terms'
 export const PRIVACY_POLICY = '/privacy'
-
-export const pathIsforCustomerOnly = (path: string): boolean =>
-  [
-    '/policies/:policyId/settings',
-    '/policies/:policyId/items/new',
-    '/policies/:policyId/items/:itemId/edit',
-    '/policies/:policyId/items/:itemId/remove-coverage',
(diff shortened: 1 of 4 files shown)


src/helpers/routes.ts
```diff
@@ -45,3 +45,16 @@ export const householdSettingsNewDependent = (policyId: string) => `/policies/${
 
 export const TERMS_OF_SERVICE = '/terms'
 export const PRIVACY_POLICY = '/privacy'
+
+export const pathIsforAdmins = (path: string): boolean =>
```
RussellChamp, line 49: hrm, i don't really like this pattern. it means that each newly added user route will have to be added here, right? that seems ripe to get unexpectedly foot-gunned

src/pages/_layout.svelte
```diff
@@ -82,13 +88,13 @@ const goToCustomerView = (event: CustomEvent) => {
   }
 }
 const goToAdminView = (event: CustomEvent) => {
-  if ($params.policyId) {
+  if ($params.policyId && routes.pathIsforAdmins($route.path)) {
     if ($params.claimId) {
-      gotoPath(event.detail.policyId, { claimId: $params.claimId })
+      gotoPath($selectedPolicyId, { claimId: $params.claimId })
```
RussellChamp, line 93: it might just be my undying love for ternaries but I think this would look cleaner as 
```
const params = 
    $params.claimId ? { claimId: $params.claimId } :
    $params.itemId ? { itemId: $params.itemId } :
    {};
gotoPath($selectedPolicyId, params)
```

src/helpers/routes.ts
RussellChamp: if you want to use this anyway, i would suggest removing the `!` (it's harder to reason in your head about not conditionals and not-not conditionals) and align the case to `pathIsForUsers/Admins`

src/pages/policies/[policyId].svelte
```diff
@@ -53,24 +54,24 @@ th {
       </tr>
       <tr>
         <th>Account</th>
-        <td>{policy.account}</td>
+        <td>{policy.account || '-'}</td>
```
Comment (RussellChamp) [inline src/pages/policies/[policyId].svelte:57]: while this code totally works as is, you can also do some CSS magic with the `:empty` pseudo-class selector. https://css-tricks.com/almanac/selectors/e/empty/ in this case, the following CSS rule would add a dash in _any_ empty 'td'.
```
td:empty:after {
  content: '-'
}
```
if you wanted it more specifically targeted, you could add a class on the elements and target that instead.

RussellChamp (changes requested):
left some comments

src/helpers/routes.ts
hobbitronics: not every user route will need to be added as most/some are ok for Admins roles.

src/pages/_layout.svelte
hobbitronics: I'm not sure if Matt will like it, but I like that we only write gotoPath once.

forevermatt (approved):

src/pages/_layout.svelte
forevermatt: You know me well: I'm not fond of multi-line logic that's not contained in a `{ }` code block. That said, @RussellChamp's suggestion is actually quite clean and readable, which is really what I'm after. 👍 

(Eek... I like Russell's code... but what ended up in this PR is noticeably less readable. Livable, but the extra newlines at every `?` and `:` seem to make it harder to read. _Blames the autoformatter_ 😉 )

hobbitronics: yeah, the autoformatter did leave its mark.

hobbitronics: I unformatted, but that will only last till someone saves this file

PR #325, opened 2021-11-17
Fix null policy bug (on first login)

forevermatt:
### Fixed
- Don't use `$user.policy_id` in determining default policy 
- Wait until we have a policy for the user before trying to view it


hobbitronics (approved):
It will nice to have this gremlin gone.

PR #326, opened 2021-11-17
Stop using Google Places API

forevermatt:
### Changed
- Stop using the Google Places Autocomplete API, to prepare for simpler country-select design

---

When all we're really interested in is the country, using a list of countries from the database works better:

- It avoids needing to adding the Google Maps/Places API to our privacy policy.
- It paves the way for future expectations of limiting certain coverage based on country.
- It avoids the problem of that Google API not allowing us to filter to just countries anyway.


hobbitronics (approved):
I unnaprove this... er approve this pull regress.

hobbitronics (commented):
Just wondering if you are leaving the env in there intentionally.

forevermatt:
> Just wondering if you are leaving the env in there intentionally.

Nope 🤦

hobbitronics (approved):
re-approved

PR #327, opened 2021-11-17
use account for teams on checkout and use entity code instead of name

hobbitronics:
- Checkout was showing ui for teams and household policies


RussellChamp (approved):
lgtm

PR #328, opened 2021-11-17
add description to incident types

hobbitronics:


src/components/RadioOptions.svelte
```diff
@@ -18,10 +18,18 @@ label {
   display: block;
   margin: 1rem 0;
 }
+.description {
+  display: block;
+  margin: 0.3em 0 0 2em;
```
RussellChamp, line 23: we should probably try to not mix em and px at least in the same component rules.

RussellChamp (approved):
good enough but left a comment.

src/components/RadioOptions.svelte
hobbitronics: Should be a little less hacky now.

PR #329, opened 2021-11-18
add person item to accountable persons

RussellChamp:
* Add a SelectAccountablePerson component based on a Select component
* Add additional data validation to DependentForm around unique dependent names and dependent birth year
* Add ability to specify size of DependentForm (so it can be w-100 on the modal and w-50 elsewhere)
* Add a Modal component extending the Dialog.Alert component
  * Additional features include the ability to programmatically close the modal, specify a title icon, and remove the default action buttons
  * These changes could potentially be merged into ui-components once functionality is finalized

### DIFF
--- a/src/components/SelectAccountablePerson.svelte
+++ b/src/components/SelectAccountablePerson.svelte
@@ -1,0 +1,106 @@
+<script lang="ts">
+import DependentForm from './forms/DependentForm.svelte'
+import Modal from './mdc/Modal.svelte'
+import { AccountablePersonOptions, getDependentOptions, getPolicyMemberOptions } from 'data/accountablePersons'
+import { addDependent, dependentsByPolicyId, PolicyDependent } from 'data/dependents'
+import { policies, Policy, PolicyType } from 'data/policies'
+import { membersByPolicyId } from 'data/policy-members'
+import { Select } from '@silintl/ui-components'
+import { createEventDispatcher } from 'svelte'
+
+export let policyId: string
+export let selectedID: string
+
+let dispatch = createEventDispatcher<{ change: AccountablePersonOptions }>()
+
+let modalOpen = false
+let showSelectBox = true
+
+$: policy = $policies.find((policy) => policy.id === policyId) || ({} as Policy)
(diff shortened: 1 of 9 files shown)


src/components/mdc/Modal.svelte
```diff
@@ -0,0 +1,92 @@
...
+  },
+]
+
+const dispatch = createEventDispatcher()
+
+let element = {}
+let mdcDialog = {}
+
+$: open && mdcDialog.open && mdcDialog.open()
+$: !open && mdcDialog.close && mdcDialog.close('closed')
```
hobbitronics, line 31: I was under the impression mdc closed itself. I thought that's what the MDCDialog:closed listener on line 36 was doing. Can you help me understand what this line does?

src/components/SelectAccountablePerson.svelte
```diff
@@ -0,0 +1,116 @@
...
+    modalOpen = true
+  } else {
+    dispatch('change', event.detail)
+  }
+}
+
+const onModalFormSubmit = async (event: CustomEvent) => {
+  if (event.type === 'submit') {
+    const depData = event.detail
+    // TODO: Figure out why adding items to this select box causes it to throw MDC errors
```
hobbitronics, line 62: there is a `populated` event Matt added to `Select` that lets you know when its populated and safe to initialize the selectedID.

hobbitronics: hmm, looks like you're already using that, so I don't know what these error are about.

hobbitronics (approved):
Looks good. I pulled the code and didn't get any errors yet.

src/components/mdc/Modal.svelte
RussellChamp: what wasn't happening before that this line adds is the ability to close the modal when the parent sets the `open` value to false. before it would ignore that and just stay open

src/components/SelectAccountablePerson.svelte
RussellChamp: i was able to work around the issue I had but think it's best to be explicit about potentially hidden issues

PR #331, opened 2021-11-18
add NoHouseholdId modal

hobbitronics:
- Customers without household Id could fill out a form but would be rejected on submitting it.
- add the modal to the new view since I cant make the `Drawer` button trigger it (urls only)
- add to views that have a "Add Item" button (to prevent too much url jumping to the form and back)

### DIFF
--- a/src/components/NoHouseholdIdModal.svelte
+++ b/src/components/NoHouseholdIdModal.svelte
@@ -2,13 +2,13 @@
 import { Dialog } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
 
-export let open = true
+export let open = false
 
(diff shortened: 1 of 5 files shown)


Schparky:
Haven't reviewed the content, so this is just based on the title. Would it be better to disable the "Add Item" button so they couldn't get to the add item form at all?

RussellChamp (changes requested):
shouldn't this modal and redirect occur _before_ the user has done a bunch of data entry?

hobbitronics:
> shouldn't this modal and redirect occur _before_ the user has done a bunch of data entry?

Yes, it does occur before the customer has a chance to fill out the form. I am thinking of just adding the household ID form to the modal instead of redirecting them to settings though.

PR #332, opened 2021-11-18
Use countries list from API

forevermatt:
### Added
- Add a `countries` svelte store, populated asynchronously when first used

### Changed
- Change "Personal Settings" page to be a searchable-select of a list of countries

---

We'll probably want to extract this to a `<CountrySelector ... />` or something like that if we end up using this in multiple places (such as on the DependentForm). There were just enough details involved that I wouldn't want them replicated throughout our codebase.

### DIFF
--- a/src/data/countries.ts
+++ b/src/data/countries.ts
@@ -1,0 +1,17 @@
+import { GET } from '.'
+import { readable } from 'svelte/store'
+
+export type Country = {
+  name: string
+}
+
+let hasLoaded = false
+
+export const countries = readable<Country[]>([], function start(set) {
+  if (!hasLoaded) {
+    GET<Country[]>('config/countries').then((values: Country[]) => {
+      set(values)
+      hasLoaded = true
+    })
+  }
+})
--- a/src/pages/settings/personal.svelte
+++ b/src/pages/settings/personal.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
 import user, { attachUserPhoto, updateUser } from '../../authn/user'
-import { Breadcrumb, FileDropArea, RadioOptions } from 'components'
+import { Breadcrumb, FileDropArea, RadioOptions, SearchableSelect } from 'components'
 import { upload } from 'data'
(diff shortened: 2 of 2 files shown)


src/pages/settings/personal.svelte
```diff
@@ -18,11 +19,14 @@ let uploading = false
 let notification_email = $user.email_override ? NOTIFICATION_OPTION_CUSTOM : NOTIFICATION_OPTION_DEFAULT
 let email_override = $user.email_override || ''
 let country = $user.country || ''
+let countryOptions = {}
 let croppie: Croppie
 let croppieContainer: HTMLDivElement
 let breadcrumbLinks = [{ name: 'Personal Settings', url: SETTINGS_PERSONAL }]
 metatags.title = formatPageTitle('Personal Settings')
 
+$: $countries.forEach((country: Country) => (countryOptions[country.name] = country.name))
```
hobbitronics, line 28: I am curious what the benefit of using `countryOptions` rather than `$countries` is.

hobbitronics: nevermind, I see now that `$countries` is an array.

hobbitronics (approved):
Nice work.

src/data/countries.ts
```diff
@@ -0,0 +1,17 @@
...
+import { readable } from 'svelte/store'
+
+export type Country = {
+  name: string
+}
+
+let hasLoaded = false
+
+export const countries = readable<Country[]>([], function start(set) {
+  if (!hasLoaded) {
```
RussellChamp, line 11: i don't believe this 'hasLoaded' check is necessary. if my understanding of readable is correct, this part should only trigger once

RussellChamp (changes requested):

src/data/countries.ts
forevermatt: In my testing, it was triggered every time I went to the page that uses it, so I think it is actually beneficial.

PR #333, opened 2021-11-19
remove auto-width from app drawer

RussellChamp:
* resolves the issue where the app drawer resizes with a long team name
* instead it will now show as below (default behavior for a Button)
 
image


hobbitronics (approved):
Looks good

forevermatt (approved):

PR #334, opened 2021-11-19
add household id form to modal

hobbitronics:
- Add a householdId form to a modal so users can add it before filling out `ItemForm`

### DIFF
--- a/src/components/NoHouseholdIdModal.svelte
+++ b/src/components/NoHouseholdIdModal.svelte
@@ -9,7 +9,7 @@ const policyData: UpdatePolicyRequestBody = {}
 
 let title = 'Missing information'
 let householdId = ''
-let hasSubmitted: boolean = false
+let hasClosed: boolean = false
 
 const buttons: Dialog.AlertButton[] = [
   { label: 'Go Back', action: 'cancel', class: 'mdc-dialog__button' },
@@ -25,14 +25,14 @@ const handleDialog = (event: CustomEvent) => {
     if (isIdValid(householdId)) {
       policyData.household_id = householdId
       dispatch('closed', { choice, data: policyData })
-      hasSubmitted = true
+      hasClosed = true
     } else {
       setNotice('Please enter a valid Household ID')
     }
   } else if (choice === 'cancel') {
     dispatch('closed', { choice })
-    // the second event when this closes happens before the policy is updated so I need to prevent it from firing if a valid id is selected
-  } else if (!hasSubmitted) {
+    hasClosed = true
+    // prevents emiiting a second 'closed' event
     dispatch('closed', { choice })
   }
 }
--- a/src/components/NoHouseholdIdModal.svelte
+++ b/src/components/NoHouseholdIdModal.svelte
@@ -2,13 +2,13 @@
 import { Dialog } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
 
-export let open = true
+export let open = false
 
 let message = 'A household ID is required before getting coverage. Add it in Policy Settings.'
 let title = 'Missing information'
 
 const buttons: Dialog.AlertButton[] = [
-  { label: 'cancel', action: 'cancel', class: 'mdc-dialog__button' },
+  { label: 'Go Back', action: 'cancel', class: 'mdc-dialog__button' },
   { label: 'Policy Settings', action: 'gotoSettings', class: 'mdc-button--raised' },
 ]
 
--- a/src/components/NoHouseholdIdModal.svelte
+++ b/src/components/NoHouseholdIdModal.svelte
@@ -38,7 +38,7 @@ const handleDialog = (event: CustomEvent) => {
   }
 }
 
-const isIdValid = (sanitizedId: string): number | boolean =>
+const isIdValid = (sanitizedId: string): boolean =>
   (sanitizedId.length > 0) && sanitizedId.split('').every((digit) => /[0-9]/.test(digit))
 </script>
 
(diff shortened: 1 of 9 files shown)


src/pages/policies/[policyId]/items/new.svelte
```diff
@@ -24,6 +26,8 @@ $: metatags.title = formatPageTitle('Items > New')
 
 $: policyId && loadItems(policyId)
 
+$: $selectedPolicy.type === PolicyType.Household && !$selectedPolicy.household_id && (open = true)
```
Comment (hobbitronics) [inline src/pages/policies/[policyId]/items/new.svelte:29]: opens the modal if a householdId doesn't exist

src/pages/policies/[policyId]/items/new.svelte
```diff
@@ -54,12 +58,24 @@ const onDelete = async (event: CustomEvent<string>) => {
...
+
+const onClosed = async (event: CustomEvent<any>) => {
+  const choice = event.detail.choice
+  if (choice === 'submit') {
+    await updatePolicy(policyId, event.detail.data)
+    setNotice('Your household ID has been saved')
+  } else {
+    history.back()
+  }
+  open = false
```
Comment (hobbitronics) [inline src/pages/policies/[policyId]/items/new.svelte:70]: line 70 is not really necessary as the modal won't open again, but you never know...

src/components/NoHouseholdIdModal.svelte
```diff
@@ -0,0 +1,54 @@
...
+      policyData.household_id = householdId
+      dispatch('closed', { choice, data: policyData })
+      hasSubmitted = true
+    } else {
+      setNotice('Please enter a valid Household ID')
+    }
+  } else if (choice === 'cancel') {
+    dispatch('closed', { choice })
+    // the second event when this closes happens before the policy is updated so I need to prevent it from firing if a valid id is selected
+  } else if (!hasSubmitted) {
```
hobbitronics, line 35: this last block happens when the Dialog.Alert fires a second time (first is choice) or the user clicks off of the modal (in which case we want them to go back to the previous page).

src/components/NoHouseholdIdModal.svelte
```diff
@@ -0,0 +1,54 @@
...
+import { Dialog, setNotice, TextField } from '@silintl/ui-components'
+import { createEventDispatcher } from 'svelte'
+
+export let open = false
+
+const policyData: UpdatePolicyRequestBody = {}
+
+let title = 'Missing information'
+let householdId = ''
+let hasSubmitted: boolean = false
```
hobbitronics, line 12: rename to hasClosed and use to prevent a second back when user clicks cancel.

src/components/NoHouseholdIdModal.svelte
```diff
@@ -0,0 +1,55 @@
...
+  } else if (choice === 'cancel') {
+    dispatch('closed', { choice })
+    hasClosed = true
+    // prevents emiiting a second 'closed' event
+  } else if (!hasClosed) {
+    dispatch('closed', { choice })
+  }
+}
+
+const isIdValid = (sanitizedId: string): number | boolean =>
```
forevermatt, line 41: Why is `number` included in this return type? It seems like we want the returned value (i.e. the answer to the question "is this ID valid?") to be just a boolean.

src/components/NoHouseholdIdModal.svelte
```diff
@@ -0,0 +1,55 @@
...
+  sanitizedId.length && sanitizedId.split('').every((digit) => /[0-9]/.test(digit))
```
forevermatt, line 42: This might accomplish the "boolean only" goal:
```suggestion
  (sanitizedId.length > 0) && sanitizedId.split('').every((digit) => /[0-9]/.test(digit))
```

forevermatt (commented):
One suggestion, but otherwise looks good.

src/components/NoHouseholdIdModal.svelte
hobbitronics: yeah, I was kind of confused by the ts warming since it doesn't seem like it should return -1. We can change it to just boolean but there might be an extra warning.

RussellChamp: the return type was determined to be `number | boolean` because that's exactly what _could_ be returned by the below expression. In the case of an empty string being passed, the expression will resolve as `"".length && "".split()...` => `0 && false` => `0` (which is a numeric value).
To resolve this, you can either change the first expression to `sanitizedId.length > 0` or you can coerce it into a boolean expression with `!!sanitizedId.length`

RussellChamp (changes requested):
lgtm other than updating the `isIdValid` method

src/components/NoHouseholdIdModal.svelte
forevermatt: Sorry, I should have included this with my previous suggested code change:
```suggestion
const isIdValid = (sanitizedId: string): boolean =>
```

src/components/NoHouseholdIdModal.svelte
```diff
@@ -0,0 +1,55 @@
...
+    dispatch('closed', { choice })
+    hasClosed = true
+    // prevents emiiting a second 'closed' event
+  } else if (!hasClosed) {
+    dispatch('closed', { choice })
+  }
+}
+
+const isIdValid = (sanitizedId: string): boolean =>
+  (sanitizedId.length > 0) && sanitizedId.split('').every((digit) => /[0-9]/.test(digit))
```
forevermatt, line 42: Not at all required, but I wonder if we could have simply done something like this:
```suggestion
const isIdValid = (id: string): boolean => /^[0-9]+$/.test(id)
```

forevermatt (approved):

src/components/NoHouseholdIdModal.svelte
hobbitronics: yeah, as long as it works the same i like it better.

PR #336, opened 2021-11-22
add CountrySelect and use in DependentForm and personal settings

hobbitronics:
- add SearchableSelect with countries list from api to DependentForm

### DIFF
--- a/src/components/CountrySelector.svelte
+++ b/src/components/CountrySelector.svelte
@@ -5,8 +5,7 @@ import { setNotice } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
 
 export let title = 'Country'
-
-let country = ''
+export let country = ''
 
 let countryOptions: any = {}
 
--- a/src/components/CountrySelector.svelte
+++ b/src/components/CountrySelector.svelte
@@ -1,0 +1,30 @@
+<script type="ts">
+import SearchableSelect from './SearchableSelect.svelte'
+import { countries, Country } from 'data/countries'
+import { setNotice } from '@silintl/ui-components'
+import { createEventDispatcher } from 'svelte'
+
+export let title = 'Country'
+
+let country = ''
+
+let countryOptions: any = {}
+
+$: $countries.forEach((country: Country) => (countryOptions[country.name] = country.name))
+
+const dispatch = createEventDispatcher()
+
+const isCountryValid = (countryName: string) => $countries.some((country) => country.name === countryName)
+
+const updateCountry = (event: CustomEvent) => {
+  if (isCountryValid(event.detail)) {
+    country = event.detail
+    dispatch('chosen', country)
+  } else {
+    setNotice('Please select a country from the list')
+  }
+}
+</script>
+
+<span class="header">{title}<span class="required">*</span></span>
+<SearchableSelect choice={country} options={countryOptions} placeholder="Enter country" on:check={updateCountry} />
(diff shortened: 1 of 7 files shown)


forevermatt (changes requested):
Let's extract the country-selection code to a new "CountrySelect" component, to avoid duplicating the logic here that we added elsewhere.

src/components/CountrySelector.svelte
```diff
@@ -0,0 +1,29 @@
...
+import { countries, Country } from 'data/countries'
+import { setNotice } from '@silintl/ui-components'
+import { createEventDispatcher } from 'svelte'
+
+export let title = 'Country'
+export let country = ''
+
+let countryOptions: any = {}
+
+$: $countries.forEach((country: Country) => (countryOptions[country.name] = country.name))
```
RussellChamp, line 12: depending on how flexible and generic you want this component to be, i'd suggest allowing passing in the list of countries to the component instead of specifying them here. for the most flexible components, they probably shouldn't ever be referencing state stores

RussellChamp (approved):
Optional suggestion for the CountrySelector

src/components/CountrySelector.svelte
hobbitronics: I thought about doing that but this reduced the duplicated code and isn't used anywhere else for now. I suppose we could make a countryOptions store derived from $countries. What are your thoughts on that?

src/components/CountrySelector.svelte
```diff
@@ -0,0 +1,29 @@
...
+  if (isCountryValid(event.detail)) {
+    country = event.detail
+    dispatch('chosen', country)
+  } else {
+    setNotice('Please select a country from the list')
+  }
+}
+</script>
+
+<span class="header">{title}<span class="required">*</span></span>
```
forevermatt, line 28: Let's leave the header out of this, since what we call this depends on where we're using this. This component itself is just the country-selector itself.

src/pages/settings/personal.svelte
```diff
@@ -140,8 +130,7 @@ p {
   {/if}
 
   <p>
-    <span class="header">Primary Location<span class="required">*</span></span>
```
forevermatt, line 143: As above, let's leave this header here on the page, since what we call it and whether it's required are outside the scope of the CountrySelect's purpose.

src/components/CountrySelector.svelte
forevermatt: I agree that it's often better to pass to components whatever they'll need, rather than having them reach out and get the data themselves.

However, in this case it seems fitting to have the CountrySelect access the `$countries` store directly, since that's a core part of this component's purpose, everywhere we use this component we'll be passing it that same data, and we don't expect that data to change on any regular basis.

forevermatt (changes requested):
Only change requested is to pull out the header for the CountrySelect, since that may differ for each page that uses this.

src/components/forms/DependentForm.svelte
```diff
@@ -107,7 +107,8 @@ const onChosen = (event: CustomEvent) => (formData.country = event.detail)
       </p>
     {/if}
     <p>
-      <CountrySelector on:chosen={onChosen} title={'Dependent Location'} />
+      <span class="header">Dependent Location<span class="required">*</span></span>
```
forevermatt, line 110: It looks like this file doesn't define that `required` CSS class, but otherwise this looks good. 👍

forevermatt (approved):

src/components/forms/DependentForm.svelte
hobbitronics: good catch. Moved it to global so they can both use it.

hobbitronics: and removed other instanced of 
.required {
  color: var(--mdc-theme-status-error);
}

PR #339, opened 2021-11-29
set all text to max-width of 80ch

hobbitronics:
- set all text to max of 80ch (this worked on text inputs too)
- del unused classes

### DIFF
--- a/src/components/global.css
+++ b/src/components/global.css
@@ -9,8 +9,8 @@ body {
   width: 100%;
 }
 
-.auto-width {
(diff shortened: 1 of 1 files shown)


RussellChamp (approved):
as long as it creates a newline and not truncates i guess this is ok. there may be some unexpected new-lining, though

merchako:
This pull might have broken pages with text centering.
Screen Shot 2021-11-30 at 10 05 00

PR #343, opened 2021-11-30
use upload instead of submit for receipt

hobbitronics:
Screen Shot 2021-11-30 at 11 40 29 AM

### DIFF
--- a/src/pages/policies/[policyId]/claims/[claimId].svelte
+++ b/src/pages/policies/[policyId]/claims/[claimId].svelte
@@ -48,7 +48,7 @@ import { formatPageTitle } from 'helpers/pageTitle'
 import { assertHas } from '../../../../validation/assertions'
 import { onMount } from 'svelte'
 import { goto, metatags } from '@roxi/routify'
-import { Page } from '@silintl/ui-components'
+import { Button, Page, setNotice } from '@silintl/ui-components'
 import { formatDistanceToNow } from 'date-fns'
 
 export let claimId: string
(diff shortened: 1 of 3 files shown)


src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -99,6 +99,8 @@ $: needsReplaceReceipt = needsReceipt && payoutOption === 'Replacement'
 $: filePurpose = getFilePurpose(claimItem, needsReceipt) as ClaimFilePurpose
 $: noFilesUploaded = !isFileUploadedByPurpose(filePurpose, claimFiles)
 $: uploadLabel = getUploadLabel(claimItem, needsReceipt, receiptType) as string
+$: uploadLabelForButton = getUploadLabel(claimItem, needsReceipt, receiptType, false) as string
```
Comment (RussellChamp) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:102]: the `string` type is already inferred from the return type of `getUploadLabel` and does not need `as string` unless you are trying to cudgel the code over the head and claim that the function returns a different value that it claims to return

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -141,7 +143,10 @@ const onDenyClaim = async (event: CustomEvent<string>) => {
   await denyClaim(claimId, message)
 }
 
-const onSubmit = async () => await submitClaim(claimId)
+const onSubmit = async () => {
+  await submitClaim(claimId)
+  setNotice('Added replacement cost and receipt')
```
Comment (RussellChamp) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:148]: `onSubmit` is called both by ClaimActions and by the new button you added. Are you sure this behaves as desired (eg when we submit from the ClaimActions)?

RussellChamp (changes requested):

RussellChamp (approved):
lgtm

PR #344, opened 2021-11-30
Better line-height, and add real Terms of Service

forevermatt:
### Added
- Add real Terms of Service

### Fixed
- Set global line-height for paragraph tags to a better default
- Make "/terms" a public route

### DIFF
--- a/src/components/App.svelte
+++ b/src/components/App.svelte
@@ -25,7 +25,7 @@ const queryHandler = {
   stringify,
 }
 
-const publicRoutes = ['/invite/:uuid', '/privacy']
+const publicRoutes = ['/invite/:uuid', '/privacy', '/terms']
 
 const authenticateUser = async () => {
   loadUser().catch((error: CustomError) => {
--- a/src/components/global.css
+++ b/src/components/global.css
@@ -10,6 +10,7 @@ body {
 }
 
 p {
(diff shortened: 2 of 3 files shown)


hobbitronics (approved):
The following makes me wonder how any claims on electronics besides theft or electrical surges would be covered. Batteries ballooning, dying, hdd/sdds failing could all be wear and tear. I would guess in practice failed hardware claims probably do get approved. Does dropping a phone from your count as negligence? Code looks fine but, the terms would make me hesitant to sign up if this were a market service like Apple Care, Square Trade, etc.
`<li><p><b>Wear and tear</b> or gradual deterioration</p></li>
    <li>
      <p>
        <b>Items lost due to negligence,</b> such items left in a vehicle, store, or falling out of a pocket or backpack.
      </p>
...

 <p><b>Bricked electronics</b></p>
      <p>
        A &ldquo;bricked&rdquo; item is an electronic device such as a mobile device, game console, or router that, due
        to corrupted firmware, user error, or a hardware problem, can no longer function.
      </p>`

merchako:
@hobbitronics , I could see why you would have hesitations. Want to take a stab at fixing them? The terms lean a bit in favor of Cover just because they're ripe for abuse if they lean in favor of the customer, and getting it just right is really hard.

PR #347, opened 2021-12-01
Fix display of items whose coverage will end soon

forevermatt:
### Added
- In the ItemsTable, indicate when coverage will end (if applicable)

### Changed
- Don't show ItemCard for items whose coverage will end soon

### DIFF
--- a/src/components/CardsGrid.svelte
+++ b/src/components/CardsGrid.svelte
@@ -14,6 +14,10 @@ export let isAdmin: boolean = false
 
 $: cardItems = recentChanges.length ? parseRecentChanges(recentChanges) : parseClaimsAndPolicyItems(claims, policyItems)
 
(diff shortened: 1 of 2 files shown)


src/components/CardsGrid.svelte
```diff
@@ -46,7 +50,12 @@ const parseRecentChanges = (changes: RecentChange[]): CardItem[] => {
 
 const parseClaimsAndPolicyItems = (claims: Claim[], policyItems: PolicyItem[]): CardItem[] => {
   // Add all the policy items and claim items together then filter and sort them
-  let cards: CardItem[] = policyItems.map((item) => ({ data: item })) || []
+  let cards: CardItem[] = []
```
RussellChamp, line 53: could probably be done with a chained filter/map

RussellChamp (approved):
lgtm with a comment

PR #350, opened 2021-12-01
Update Terms of Service to incorporate changes from Dec. 1, 2021

forevermatt:
### Changed
- Update Terms of Service to incorporate changes from Dec. 1, 2021

---

We don't yet need to notify users of this change since we have not yet gone live with the production version of this service.


RussellChamp (approved):
sure i guess

PR #352, opened 2021-12-02
fixed Approve button label & changed type to enum

hobbitronics:
on Claim view the Steward was seeing the wrong Approve button label due to a typo.

### DIFF
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -28,7 +28,7 @@ let message = ''
 let status: ClaimStatus
 $: status = claim.status
 
-$: approveButtonLabel = receiptType === 'Replacement' ? 'replace' : 'repair'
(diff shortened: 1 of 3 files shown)


src/components/ClaimActions.svelte
```diff
@@ -28,7 +28,7 @@ let message = ''
 let status: ClaimStatus
 $: status = claim.status
 
-$: approveButtonLabel = receiptType === 'Replacement' ? 'replace' : 'repair'
+$: approveButtonLabel = receiptType === 'replacement' ? 'replace' : 'repair'
```
forevermatt, line 31: Should we also change the receipt type values to an enum?

forevermatt (approved):

src/components/ClaimActions.svelte
hobbitronics: probably should. It is distinct from the PayoutOption which is the confusion over why this bug happened.

PR #354, opened 2021-12-02
center pages using <p> for main content

hobbitronics:
- after adding p { max-width: 80ch}, pages with `<p>` for main content were not centered
- added m-0-auto to global.css and these pages

### DIFF
--- a/src/components/global.css
+++ b/src/components/global.css
@@ -14,6 +14,11 @@ p {
   max-width: 80ch;
 }
 
+/* TODO move this to ui-components */
+.m-0-auto {
+  margin: 0 auto !important;
+}
+
 .max-content-width {
   width: max-content;
 }
--- a/src/pages/_fallback.svelte
+++ b/src/pages/_fallback.svelte
@@ -17,7 +17,7 @@ p {
 </style>
 
 <Page>
-  <p>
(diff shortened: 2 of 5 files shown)


src/components/global.css
```diff
@@ -14,6 +14,11 @@ p {
   max-width: 80ch;
 }
 
+/* TODO move this to ui-components */
+.m-0-auto {
```
forevermatt, line 18: What if instead we simply add `margin-right: auto;` and `margin-left: auto;` to our `p` rule, above?

hobbitronics: do we really want that in all uses of `<p>` though?

forevermatt: Good question... I don't know. I'm mostly wondering if there's some way to avoid having to add a custom class to get our pages to center properly.

Would there be a way to center the contents of our `<Page>` tags instead? We can proceed with what you have here in the meantime, though.

forevermatt (approved):

src/components/global.css
hobbitronics: we might be able to target an MDC class in '<Page>' with the automargin. I will have a look to see if that's a solution.

PR #355, opened 2021-12-02
add enums

hobbitronics:
- used ItemCoverageStatus, PayoutOption and ClaimStatus instead of strings.
- fixed a few ts errors


RussellChamp:
oh yeah baybee. these are the PRs I live for

src/components/ClaimActions.svelte
```diff
@@ -5,12 +5,14 @@ import {
...
   editableStatuses,
   PayoutOption,
+  statusesAvaitingAdmin,
   statusesAwaitingSignator,
   statusesAwaitingSteward,
 } from 'data/claims'
 import { roleSelection } from 'data/role-policy-selection'
 import Description from './Description.svelte'
 import { throwError } from '../error'
+import { ReceiptType } from '../pages/policies/[policyId]/claims/[claimId].svelte'
```
RussellChamp, line 15: it's probably a bad idea in general to have things being exported from our 'page' components (which really should just be consumers of stuff.) We might accidentally introduce circular dependencies that way. if something like 'ReceiptType' is shared across components it should probably be in 'components' somewhere (either in a component there or in an 'enums' folder or something else)

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -1,3 +1,10 @@
+<script context="module" lang="ts">
```
Comment (RussellChamp) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:1]: i'd suggest moving this out of a 'pages' component

RussellChamp (changes requested):
looks really good. we already set most of those datatypes on the response data we get from BE endpoints already. i'd just recommend we move that one type out of the 'page' component.

RussellChamp (approved):

PR #356, opened 2021-12-02
fix some type errors in MoneyInput

hobbitronics:
- I know the input `type="number"` but it causes a type error and this works fine


RussellChamp (approved):
better than before, i guess. but i really don't like the `let obj = {} as SomethingElse` pattern we are using. :/

PR #357, opened 2021-12-03
allow radio options to be selected using label

RussellChamp:
the problem was that multiple radio options were given the same id


hobbitronics (approved):
lgtm

PR #360, opened 2021-12-03
fixed SelectAccountablePerson initializing empty

hobbitronics:
- The problem was not the selectedID being set too early but the options loading too late for the Select component.
- When hard refreshed from the edit view for loadPolicyDependants and loadPolicyMembers were delivering the goods, but when hard refreshed from home or items views then navigating to itemId/edit they were not able to deliver them on time. Maybe its because they are being called from onMount, but ultimately I blame MDC and hate the Select component.
- its an inelegant solution but it works for now.

### DIFF
--- a/src/pages/policies/[policyId]/items/[itemId]/edit.svelte
+++ b/src/pages/policies/[policyId]/items/[itemId]/edit.svelte
@@ -2,6 +2,8 @@
 import Checkout from 'Checkout.svelte'
 import { Breadcrumb, ItemBanner, ItemForm } from 'components'
 import { loading } from 'components/progress'
+import { loadDependents } from 'data/dependents'
+import { loadMembersOfPolicy } from 'data/policy-members'
 import {
   deleteItem,
   ItemCoverageStatus,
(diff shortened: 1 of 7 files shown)


src/pages/policies/[policyId]/items/[itemId]/edit.svelte
```diff
@@ -26,8 +24,6 @@ export let policyId = $selectedPolicyId
 let isCheckingOut: boolean = false
 
 onMount(() => {
```
Comment (hobbitronics) [inline src/pages/policies/[policyId]/items/[itemId]/edit.svelte:26]:
```suggestion
onMount(() => {
//dependents and members are loading from _layout.svelte
```

RussellChamp (changes requested):
i'm not a fan of this work around. it takes the logic from a single component and spreads it around other places. i was looking at this the other day, too.
I think a more elegant work around would be to update some code in the SelectAccountablePerson component. we already conditionally show the 'Select' component. We could add some code that says "if we have specified a 'selectedID', then set 'showSelectBox' to false until 'accountablePersons' is populated and included that ID in it". i believe this would resolve the issue and result in the select box not being present until it is ready to be displayed.
The best-best way to resolve this would be to have route guards on page components but i'm not sure how svelte handles that.

RussellChamp (approved):
cool. i think this looks even better from before

PR #362, opened 2021-12-06
don't allow customer to change household ID

hobbitronics:
- remove householdId form noHouseholdIdModal
- add message to contact cover@sil.org for householdId changes
- hide householdId field on 'policy/settings' view for customer
- add link to 'policy/settings' on policy view
- add a const and route for cover.sil.or
ezgif-2-a1d99ef91226
g

image

### DIFF
--- a/src/components/NoHouseholdIdModal.svelte
+++ b/src/components/NoHouseholdIdModal.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
+import { COVER_EMAIL } from './const'
 import type { UpdatePolicyRequestBody } from 'data/policies'
-import { POLICY_NEW_TEAM } from 'helpers/routes'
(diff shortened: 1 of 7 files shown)


merchako (approved):
We should add breadcrumbs when we have time, especially for these pages that are 2 and 3 levels deep into the navigational hierarchy.
https://trello.com/c/losqxvVD/368-add-breadcrumbs-for-steward-editing-a-policys-settings

hobbitronics:
> We should add breadcrumbs when we have time, especially for these pages that are 2 and 3 levels deep into the navigational hierarchy. https://trello.com/c/losqxvVD/368-add-breadcrumbs-for-steward-editing-a-policys-settings

added them to this page for the Admins.

PR #365, opened 2021-12-07
show "submit changes" when no file required

hobbitronics:
- when changes are requested don't show "Upload.. " on button
- adjust message after submiting


RussellChamp (approved):
lgtm. always something fidgety

PR #366, opened 2021-12-07
change claimStates.Revew3.title

hobbitronics:
- update title to "Awaiting review 3/3" https://trello.com/c/fCN5ZgVv/333-clarify-review-process

### DIFF
--- a/src/data/states.ts
+++ b/src/data/states.ts
@@ -62,7 +62,7 @@ export const claimStates: { [stateName: string]: State } = {
   DraftSecondary: warning,
   Review1: pendingClaim,
   Review2: pendingClaim,
-  Review3: pendingClaim,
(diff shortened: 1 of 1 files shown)


RussellChamp (approved):
nice. why did i think we were using messages returned by the BE?

hobbitronics:
> nice. why did i think we were using messages returned by the BE?

You were probably thinking of claim.status_reason.

merchako:
Do we have different state callout text coded for customers vs stewards, or are they both the same?

hobbitronics:
> Do we have different state callout text coded for customers vs stewards, or are they both the same?

yes they are different. There is a state getter function that is supposed to retrieve claim states for admins vs customers.

PR #367, opened 2021-12-07
hide secondary claimCard banner from admin

hobbitronics:
- Admin shouldn't see this banner
daryl


RussellChamp (approved):
lgtm as long as there are never any cases where we want to show an admin a secondary banner.

PR #368, opened 2021-12-07
sort policyItems so Inactive is not at the top

hobbitronics:
image

### DIFF
--- a/src/pages/policies/[policyId].svelte
+++ b/src/pages/policies/[policyId].svelte
@@ -23,7 +23,9 @@ $: policy = $selectedPolicy
 $: members = policy.members || []
 
 $: policyId && loadItems(policyId)
-$: items = $selectedPolicyItems
+$: items = $selectedPolicyItems.sort((a, b) =>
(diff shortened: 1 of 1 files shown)


src/pages/policies/[policyId].svelte
```diff
@@ -23,7 +23,9 @@ $: policy = $selectedPolicy
 $: members = policy.members || []
 
 $: policyId && loadItems(policyId)
-$: items = $selectedPolicyItems
+$: items = $selectedPolicyItems.sort((a, b) =>
+  a.coverage_status === b.coverage_status ? 0 : a.coverage_status > b.coverage_status ? 1 : -1
```
Comment (RussellChamp) [inline src/pages/policies/[policyId].svelte:27]: it's not immediately obvious what the intent is here. either leave an inline comment saying what you're trying to do or refactor the logic into a well-named helper function

RussellChamp (changes requested):
lgtm with one small change

RussellChamp (approved):
good enuf

PR #370, opened 2021-12-08
email override not initializing properly on reload

hobbitronics:
- fix custom email bug


RussellChamp (approved):
lgtm but i think country also didn't load right

PR #371, opened 2021-12-08
use claim.total_payout

hobbitronics:
- use property supplied by api and delete ui logic


Schparky:
Would it make sense to make the label changes in this PR as well? (Remove "if approved" and "maximum" when not appropriate.)

Schparky (approved):

PR #375, opened 2021-12-09
new-claim show save for later and disable buttons

hobbitronics:
- show "save for later" alongside "continue"
- disable buttons when missing fields
- allow saving a draft claim without all the fields

### DIFF
--- a/src/business-rules/claim-payout-amount.ts
+++ b/src/business-rules/claim-payout-amount.ts
@@ -4,7 +4,7 @@ import { PayoutOption, ClaimItem, ClaimStatus, ClaimFilePurpose } from 'data/cla
 export const DEDUCTIBLE = 0.05
 export const LOSS_REASON_EVACUATION = 'Evacuation'
 
-export const isFairMarketValueNeeded = (isRepairable?: boolean | null, payoutOption?: string): boolean => {
+export const isFairMarketValueNeeded = (isRepairable?: boolean, payoutOption?: string): boolean => {
   return isRepairable || payoutOption === PayoutOption.FMV
 }
 
--- a/src/business-rules/claim-payout-amount.ts
+++ b/src/business-rules/claim-payout-amount.ts
@@ -4,7 +4,7 @@ import { PayoutOption, ClaimItem, ClaimStatus, ClaimFilePurpose } from 'data/cla
 export const DEDUCTIBLE = 0.05
 export const LOSS_REASON_EVACUATION = 'Evacuation'
 
-export const isFairMarketValueNeeded = (isRepairable?: boolean, payoutOption?: string): boolean => {
+export const isFairMarketValueNeeded = (isRepairable?: boolean | null, payoutOption?: string): boolean => {
   return isRepairable || payoutOption === PayoutOption.FMV
 }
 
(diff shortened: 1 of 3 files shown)


src/business-rules/claim-payout-amount.ts
```diff
@@ -4,7 +4,7 @@ import { PayoutOption, ClaimItem, ClaimStatus, ClaimFilePurpose } from 'data/cla
 export const DEDUCTIBLE = 0.05
 export const LOSS_REASON_EVACUATION = 'Evacuation'
 
-export const isFairMarketValueNeeded = (isRepairable?: boolean, payoutOption?: string): boolean => {
+export const isFairMarketValueNeeded = (isRepairable?: boolean | null, payoutOption?: string): boolean => {
```
Schparky, line 7: Maybe if you did this, you could leave the `| null` off.
```
$: shouldAskForFMV = isRepairable !== null && isFairMarketValueNeeded(isRepairable, payoutOption)
```

src/components/forms/ClaimForm.svelte
```diff
@@ -95,6 +95,13 @@ $: !isRepairable && unSetRepairEstimate()
 $: needsEvidence = !unrepairableOrTooExpensive || payoutOption === PayoutOption.FMV
 $: needsPayoutOption = !(isRepairable || isEvacuation) || repairCostIsTooHigh
 $: canContinueToEvidence = (!!repairEstimateUSD && !!fairMarketValueUSD) || (!!fairMarketValueUSD && !isRepairable)
+$: saveButtonIsDisable = !situationDescription || !lossReason
```
Schparky, line 98: Did you leave off the 'd' on purpose? (saveButtonIsDisabled)

src/components/forms/ClaimForm.svelte
```diff
@@ -278,14 +284,11 @@ const unSetReplaceEstimate = () => {
...
-          <Button on:click={onSaveForLater} raised>Continue</Button>
-        {:else}
-          <Button on:click={onSaveForLater} outlined>Save For Later</Button>
-        {/if}
+        <Button on:click={onSaveForLater} disabled={saveButtonIsDisable} outlined>Save For Later</Button>
+        <Button on:click={onSaveForLater} disabled={!canContinueToEvidence} raised>Continue</Button>
       {:else}
-        <Button on:click={onSaveForLater} outlined>Save For Later</Button>
-        <Button on:click={onSubmitClaim} raised>Submit Claim</Button>
+        <Button on:click={onSaveForLater} disabled={saveButtonIsDisable} outlined>Save For Later</Button>
```
Schparky, line 290: This is the same now, so it could be moved outside the `if needsEvidence`

Schparky (approved):

src/components/forms/ClaimForm.svelte
hobbitronics: oh, nope.

merchako:
No need for an empty radio input. Set the default option to "Repairable". 
Why not empty? Article: https://www.nngroup.com/articles/radio-buttons-default-selection/

We can keep that as a radio button for now. I'm inclined to switch to a singleton checkbox or toggle, but I'm doing some research.

PR #376, opened 2021-12-09
remove session resume feature

hobbitronics:
- reverts https://github.com/silinternational/cover-ui/pull/309


Schparky (approved):
In the code that redirects to login, can you add the current url as a return-to?

hobbitronics:
> In the code that redirects to login, can you add the current url as a return-to?

It seems you are returned to the url after logging in. Maybe code this was breaking that.

Schparky:
> > In the code that redirects to login, can you add the current url as a return-to?
> 
> It seems you are returned to the url after logging in. Maybe code this was breaking that.

Yes, that is possible, but if you really want to return the user back to the page they were on, the API will have to be informed of that URL. This is not a Priority 1 thing, though.

PR #377, opened 2021-12-10
get rid of nulls in claimform except for one

hobbitronics:
- search and destroy the nulls


RussellChamp (approved):
better than before i guess

Schparky:
Based on @merchako 's recent comment on Slack, we might have to rip out all of these changes

PR #378, opened 2021-12-10
repairableSelection default to 'repairable

hobbitronics:
- set claimForm radio button default

### DIFF
--- a/src/components/forms/ClaimForm.svelte
+++ b/src/components/forms/ClaimForm.svelte
@@ -180,9 +180,12 @@ const setInitialValues = (claim: Claim, claimItem: ClaimItem) => {
   }
   lossReason = claim.incident_type || lossReason
   situationDescription = claim.incident_description || situationDescription
-  if (typeof claimItem.is_repairable === 'boolean') {
-    repairableSelection = claimItem.is_repairable ? 'repairable' : 'not_repairable'
-  }
+  repairableSelection =
+    typeof claimItem.is_repairable === 'boolean'
+      ? claimItem.is_repairable
+        ? 'repairable'
+        : 'not_repairable'
+      : 'repairable'
 
   payoutOption = claimItem.payout_option || payoutOption
 
--- a/src/components/forms/ClaimForm.svelte
+++ b/src/components/forms/ClaimForm.svelte
@@ -181,11 +181,9 @@ const setInitialValues = (claim: Claim, claimItem: ClaimItem) => {
   lossReason = claim.incident_type || lossReason
   situationDescription = claim.incident_description || situationDescription
   repairableSelection =
-    typeof claimItem.is_repairable === 'boolean'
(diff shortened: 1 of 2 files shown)


Schparky:
If this is what I think it is, we should revert the API changes that allow `is_repairable` to be `null`

src/components/forms/ClaimForm.svelte
```diff
@@ -180,9 +180,12 @@ const setInitialValues = (claim: Claim, claimItem: ClaimItem) => {
...
   situationDescription = claim.incident_description || situationDescription
-  if (typeof claimItem.is_repairable === 'boolean') {
-    repairableSelection = claimItem.is_repairable ? 'repairable' : 'not_repairable'
-  }
+  repairableSelection =
+    typeof claimItem.is_repairable === 'boolean'
+      ? claimItem.is_repairable
+        ? 'repairable'
+        : 'not_repairable'
+      : 'repairable'
```
RussellChamp, line 188: i lover ternaries but nesting them can quickly become unmaintainable and hard to read. this could be refactored
```suggestion
  repairableSelection =
    typeof claimItem.is_repairable !== 'boolean' ||  claimItem.is_repairable
        ? 'repairable'
        : 'not_repairable'
```

RussellChamp (changes requested):
left suggestion

hobbitronics:
> If this is what I think it is, we should revert the API changes that allow `is_repairable` to be `null`

After we did that change Alex showed me this https://www.nngroup.com/articles/radio-buttons-default-selection/

Schparky:
> > If this is what I think it is, we should revert the API changes that allow `is_repairable` to be `null`
> 
> After we did that change Alex showed me this https://www.nngroup.com/articles/radio-buttons-default-selection/

Right, which means you will never send `is_repairable` as `null`, right?

hobbitronics:
> > > If this is what I think it is, we should revert the API changes that allow `is_repairable` to be `null`
> > 
> > 
> > After we did that change Alex showed me this https://www.nngroup.com/articles/radio-buttons-default-selection/
> 
> Right, which means you will never send `is_repairable` as `null`, right?

Correct. The design is to make it a checkbox in the future so that will never be `null` either.

src/data/claims.ts
```diff
@@ -362,7 +362,7 @@ export async function submitClaim(claimId: string): Promise<void> {
 export async function updateClaimItem(claimId: string, claimItemId: string, claimItemData: any): Promise<void> {
   const parsedData: UpdateClaimItemRequestBody = {
     fmv: convertToCents(claimItemData.fairMarketValueUSD),
-    is_repairable: claimItemData.isRepairable,
+    is_repairable: claimItemData.isRepairable ?? null,
```
hobbitronics, line 365:
```suggestion
    is_repairable: claimItemData.isRepairable,
```

PR #379, opened 2021-12-10
Fix claims page

forevermatt:
### Fixed
- Add (missing) table of claims to the claims page

---

Side note: We still seem to have an off-by-one error in the output of both `formatDate()` and `formatFriendlyDate()`, and I haven't had time (and probably won't have time) to fix that myself in the near future, if one of you two would be available to handle that.

### DIFF
--- a/src/components/ClaimsTable.svelte
+++ b/src/components/ClaimsTable.svelte
@@ -1,0 +1,60 @@
+<script lang="ts">
+import { Claim } from '../data/claims'
+import { formatFriendlyDate } from 'helpers/date'
+import { formatMoney } from 'helpers/money'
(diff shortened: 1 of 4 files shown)


Schparky (approved):

hobbitronics (approved):
Looks good.

merchako (commented):
The idea sounds good to me, but sometimes I have trouble critiquing interfaces by reading TypeScript.

RussellChamp:
> The idea sounds good to me, but sometimes I have trouble critiquing interfaces by reading TypeScript.

@merchako hasn't lived in the matrix for long enough, i guess.

PR #383, opened 2021-12-10
hide menu items while waiting for a policy to be created

Schparky:
Perhaps a bit crude, but better than letting the user wander around broken links


hobbitronics (approved):
could we just add `|| !policyId` to 'userIsAnonymous' and rename that var?

Schparky:
> could we just add `|| !policyId` to 'userIsAnonymous' and rename that var?

Not sure, but I think it's a bit different. I imagine this will be improved soon with a better UX.

PR #384, opened 2021-12-13
Footer placement

hobbitronics:
- center footer by adding right-margin equivalent to Drawer width when drawer is static(above mobile sizes)
- update ui-components to fix the Drawer/Page height issue to stop forcing the footer off the bottom of the page (pending release)
- 
image


RussellChamp (approved):
ah, cool. if it looks right then lgtm

Schparky (approved):

PR #389, opened 2021-12-14
center and resize logo

hobbitronics:
image


RussellChamp (approved):
lgtm

merchako (commented):
lgtm

merchako (commented):
Y'all, I just discovered what the Cloudflare pages does. It's so great!

Could we do 8rem instead of 10rem?

PR #390, opened 2021-12-14
force login on 401

hobbitronics:
- fix: users able to navigate around and view data in memory after logging out on another tab

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -1,7 +1,7 @@
 import { getToken } from '../authn/token'
 import { start, stop } from 'components/progress'
 import { throwError } from '../error'
-import t from '../i18n'
+import { login } from '../authn'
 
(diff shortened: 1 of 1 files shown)


Schparky:
Does this address @forevermatt 's concern about the public pages?

mtompset (commented):
All 401's force a login. 401's were the only thing using t(). Looks code equivalent otherwise. Wait for an approval by someone more familiar with cover. :)

hobbitronics:
> Does this address @forevermatt 's concern about the public pages?

I have tested out `terms` and `privacy policy` and do not get any `401`s. I am unaware of any other public pages or other 401's (like a Customer trying to call an Admin endpoint gives a 404)

Schparky (approved):

PR #391, opened 2021-12-14
add maxlength to text inputs

hobbitronics:
- added maxlength const
- set maxlength in all text/number inputs
- TextArea will now show how many characters are left TextField does not do this
- Added required prop to MoneyInput (did not use it yet in any instances)
- further changes will need to be made in ui-components

### DIFF
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 import { UserAppRole } from '../authn/user'
-import { MAX_INPUT_LENGTH as maxlength } from 'components/const'
+import { MAX_TEXT_AREA_LENGTH as maxlength } from 'components/const'
 import {
   Claim,
   ClaimStatus,
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { UserAppRole } from '../authn/user'
+import { MAX_INPUT_LENGTH as maxlength } from 'components/const'
 import {
   Claim,
   ClaimStatus,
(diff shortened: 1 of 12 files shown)


src/components/MoneyInput.svelte
```diff
@@ -7,9 +7,10 @@ import { onMount } from 'svelte'
 export let label = ''
 export let value = ''
 export let placeholder = ''
-export let maxlength: number = 524288 /* default */
+export let maxlength: number = 2048 /* default */
```
hobbitronics, line 10: I guess this should be 255 as well.
```suggestion
export let maxlength: number = 255/* default */
```

src/components/SearchableSelect.svelte
```diff
@@ -60,6 +61,7 @@ const onChange = () => options[internalChoice] && dispatch('chosen', options[int
 
 <label class="custom-field" style="--field-padding: {padding};">
   <input
+    {maxlength}
```
hobbitronics, line 64: not really necessary since the value is only used to lookup the value of `options`

src/components/forms/ClaimForm.svelte
```diff
@@ -181,9 +182,7 @@ const setInitialValues = (claim: Claim, claimItem: ClaimItem) => {
   lossReason = claim.incident_type || lossReason
   situationDescription = claim.incident_description || situationDescription
   repairableSelection =
-    typeof claimItem.is_repairable !== 'boolean' ||  claimItem.is_repairable
-        ? 'repairable'
-        : 'not_repairable'
+    typeof claimItem.is_repairable !== 'boolean' || claimItem.is_repairable ? 'repairable' : 'not_repairable'
```
hobbitronics, line 185: prettier did this

src/components/ClaimActions.svelte
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { UserAppRole } from '../authn/user'
+import { MAX_INPUT_LENGTH as maxlength } from 'components/const'
```
Schparky, line 3: I think this should be MAX_TEXT_AREA_LENGTH. Is it the message a steward sends on a claim review?

hobbitronics: yes, that is correct. I missed it because its a long TextInput

Schparky (approved):

PR #397, opened 2022-01-03
hide inactive items on policy detail page and more...

Schparky:
### Fixed
- Fix logic for coverage & premium totals. It was including all non-`Inactive` items, rather than only `Approved` items.
### Changed
- Hide `Inactive` items on the policy detail page.
### Added
- Add `itemIsInactive` helper and use coverage status helpers on the Policy Items page.


hobbitronics (approved):
Looks good.

PR #401, opened 2022-01-03
fix circular dependency on authn/user

Schparky:
Storage code borrowed from silinternational/wecarry-ui

### DIFF
--- a/src/data/storage.ts
+++ b/src/data/storage.ts
@@ -1,8 +1,10 @@
 
-export function onClear(fn: clearFn): void {
+// onClear pushes a function that will be executed one and only one time, when clearApp is called
+export function onClear(fn: ClearFn): void {
   customClearFns.push(fn)
 }
 
+// clearApp executes any and all functions pushed using onClear since the last time clearApp was called
 export function clearApp(): void {
   while (customClearFns.length) {
     const fn = customClearFns.pop()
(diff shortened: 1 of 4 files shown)


src/data/storage.ts
```diff
@@ -0,0 +1,19 @@
...
+    const fn = customClearFns.pop()
+    if (typeof fn === 'function') {
+      fn()
+    }
+  }
+}
+
+const customClearFns: Array<clearFn> = []
+
+interface clearFn {
```
RussellChamp, line 17: by convention, interfaces (much like classes and types) should start with a capital letter

src/data/storage.ts
```diff
@@ -0,0 +1,19 @@
...
+export function clearApp(): void {
+  while (customClearFns.length) {
+    const fn = customClearFns.pop()
+    if (typeof fn === 'function') {
+      fn()
+    }
+  }
+}
+
+const customClearFns: Array<clearFn> = []
```
RussellChamp, line 15: while both syntax are valid, we have standardized on using `T[]` for array typing

src/data/storage.ts
```diff
@@ -0,0 +1,19 @@
+
+export function onClear(fn: clearFn): void {
```
RussellChamp, line 2: if the intent is that the `clearFn`s are only to be executed _once_ and then removed from the array (as it appears to do in `clearApp` it should be noted in an inline comment on one of these lines as the behavior could easily be misunderstood by "the next guy" who need to make a change to the code.

RussellChamp (changes requested):
looks mostly good

RussellChamp (commented):
lgtm

hobbitronics (approved):
Nice

PR #403, opened 2022-01-04
move user.ts from authn to data

Schparky:


RussellChamp (approved):
looks like to checks out. you didn't even try to pull a Russell and hide extra refactors among the renames.

Schparky:
> looks like to checks out. you didn't even try to pull a Russell and hide extra refactors among the renames.

Ha! Nope, this is all the work of jetbrains

PR #405, opened 2022-01-04
resolve "Rollup 'sourcemap'" warning

Schparky:
"Rollup 'sourcemap' option must be set to generate source maps."

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -19,7 +19,7 @@ export default {
 	output: {
 		file: 'dist/bundle.js',
 		format: 'iife',
-		sourcemap: production,
+		sourcemap: !production,
 	},
 	plugins: [
 		svelte({


Schparky:
See https://stackoverflow.com/questions/63128597/how-to-get-rid-of-the-rollup-plugin-typescript-rollup-sourcemap-option-must

hobbitronics (approved):
not sure what this does exactly, but trust it doesn't blow things up.

Schparky:
If I understand correctly, it generates a source map only if NOT on production. Seems it should have been this way to start with. But there's a decent chance I'm altogether wrong!

Schparky:
I just confirmed. With this change, `bundle.map` is generated in the `dist` directory when I build locally. After reverting the change, the map file is no longer there.

PR #410, opened 2022-01-05
show different claim states to Signator for Review

hobbitronics:
Signator:
image
Steward:
image

### DIFF
--- a/src/components/banners/ClaimBanner.svelte
+++ b/src/components/banners/ClaimBanner.svelte
@@ -2,11 +2,11 @@
 import StatusBanner from '../banners/StatusBanner.svelte'
 import { getClaimState, SecondaryClaimStatus, State } from 'data/states'
 import type { ClaimStatus, ReceiptType } from 'data/claims'
-import type { UserAppRole } from 'data/user'
+import { UserAppRole } from 'data/user'
 
 export let claimStatus: ClaimStatus | SecondaryClaimStatus
-export let receiptType: ReceiptType | undefined = undefined
-export let roleSelection: UserAppRole | undefined = undefined
+export let receiptType: ReceiptType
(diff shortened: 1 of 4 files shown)


src/components/banners/ClaimBanner.svelte
```diff
@@ -1,13 +1,14 @@
...
+import StatusBanner from '../banners/StatusBanner.svelte'
 import { getClaimState, SecondaryClaimStatus, State } from 'data/states'
 import type { ClaimStatus, ReceiptType } from 'data/claims'
-import StatusBanner from '../banners/StatusBanner.svelte'
+import type { UserAppRole } from 'data/user'
 
 export let claimStatus: ClaimStatus | SecondaryClaimStatus
 export let receiptType: ReceiptType
-export let isAdmin: boolean = false
+export let roleSelection: UserAppRole
```
Schparky, line 9: There are other usages of ClaimBanner that probably need to be updated with this new property.

Schparky (changes requested):
added one change request

src/components/banners/ClaimBanner.svelte
hobbitronics: I think I will just default to claimStates instead of requiring `roleSelection`

Schparky: Does the roleSelection prop need a default? I can't remember if a warning or error is thrown when a component doesn't provide a default value and none is given by its consumer. I seem to remember a console warning.

RussellChamp (approved):
looks pretty good to me if you resolve schram's issues

hobbitronics:
> added one change request

No build/console warnings now.

Schparky (approved):

PR #411, opened 2022-01-06
Release

Schparky:
### Added
- Added buttons to show more claims or more items on Policy view
- Added pagination on admin policy page
### Fixed
- Fixed some of the circular dependencies
- Cleared a warning regarding an unused property on FilePreview
- Cleared a warning regarding an unused CSS selector on the privacy page
### Changed
- Removed `customer/home` page and `policies/{id}/home` page
- Don't include a sourcemap file for production, but do so if NOT in production
- Show different claim states to Signator than Steward
- made ItemsTable sortable and used in Policy view
- updated ui-components
---

**CAUTION**: https://github.com/silinternational/cover-api/pull/374 must be merged before this PR


src/components/banners/ClaimBanner.svelte
```diff
@@ -1,13 +1,14 @@
 <script lang="ts">
+import StatusBanner from '../banners/StatusBanner.svelte'
 import { getClaimState, SecondaryClaimStatus, State } from 'data/states'
 import type { ClaimStatus, ReceiptType } from 'data/claims'
-import StatusBanner from '../banners/StatusBanner.svelte'
+import { UserAppRole } from 'data/user'
 
 export let claimStatus: ClaimStatus | SecondaryClaimStatus
 export let receiptType: ReceiptType
```
hobbitronics, line 8: I decided to leave the warning on ReceiptType so we can address it properly in a sep. PR.

hobbitronics (approved):
Looks good, just a bug to deal with in what seems to be the Paginator where you get redirected to admin/policies after changing limit, but not a blocker. Also there seems to be a little jumpiness an view changes related to this as well.

PR #412, opened 2022-01-06
improve role menu

RussellChamp:
Copied the Menu component from ui-components in order to start massaging it into a better state.
this half-completes the work in https://trello.com/c/tyIZVBsz/402-update-show-policy-nav-dropdown but i'm not enough of a CSS wizard to get it to work any time soon
Todo:
* Get the "Show policy" display
* Add a drop-down notch


src/components/RoleAndPolicyMenu.svelte
```diff
@@ -97,6 +112,27 @@ const getButtonText = (userAppRoleSelection: UserAppRole, policyIdSelection: str
   return 'Household'
 }
 
+const getButtonIcon = (userAppRoleSelection: UserAppRole, policyIdSelection: string, myPolicies: Policy[]) => {
```
hobbitronics, line 115: could you leave out `myPolicies` and just use `$selectedPolicy` from `policies.ts`?

src/components/mdc/SuperMenu.svelte
```diff
@@ -0,0 +1,90 @@
+<!-- https://github.com/material-components/material-components-web/tree/master/packages/mdc-menu -->
+<script lang="ts">
+import { MDCMenu } from '@material/menu'
+import { goto } from '@roxi/routify'
+import { onMount } from 'svelte'
+import type { MenuItem } from './types'
+
+// TODO:
```
hobbitronics, line 8: forget something?

hobbitronics (approved):
Question, is this breaking with the current Menu or just an extension? If an extension I would normally move the change to ui-components, but if there is still more work to do rather than publishing partially done work it can live here for a bit. It will have to make its bed and do its own dishes though.

src/components/RoleAndPolicyMenu.svelte
Schparky: Either way, it would be good to have `policy` as a parameter. The calling code can use either `$selectedPolicy` or the `find` code below, as appropriate.

Schparky (approved):
cool stuff

src/components/RoleAndPolicyMenu.svelte
RussellChamp: i think it makes reading the reactive statemenets a little easier

src/components/mdc/SuperMenu.svelte
RussellChamp: merge changes back into ui-components when happy with the changes

PR #415, opened 2022-01-07
Update items view

hobbitronics:
- make Items view filterable
- remove card
- combine active and inactive tables
- add cardLimit prop to CardsGrid
- use editableCoverageStatuses to show 'Edit' in menu

image

### DIFF
--- a/src/pages/policies/[policyId].svelte
+++ b/src/pages/policies/[policyId].svelte
@@ -11,7 +11,7 @@ import { formatMoney } from 'helpers/money'
 import { formatPageTitle } from 'helpers/pageTitle'
 import { customerClaimDetails, itemDetails, items as itemsRoute, settingsPolicy } from 'helpers/routes'
 import { goto, metatags } from '@roxi/routify'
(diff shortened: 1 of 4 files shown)


src/pages/policies/[policyId].svelte
```diff
@@ -95,8 +94,9 @@ th {
   <Row cols={'12'}>
     <CardsGrid
       isAdmin={isAdmin($roleSelection)}
-      claims={claimsForGrid}
-      policyItems={items.slice(0, 3)}
+      claims={recentClaims}
+      policyItems={items}
```
Comment (Schparky) [inline src/pages/policies/[policyId].svelte:98]: Shouldn't claims and policies be limited in how many cards can show?

Schparky (approved):
one question

src/pages/policies/[policyId].svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId].svelte:98]: that is what `cardLimit` is doing if I understand your question.

Comment (Schparky) [reply src/pages/policies/[policyId].svelte:98]: Oh, sorry. You're right

PR #418, opened 2022-01-10
fix data not reloading on policy change

hobbitronics:
# Fix
- data not reloading when switching policies on claims/items views
# Remove
- removed some unused code
- loading data in onMount no longer necessary

### DIFF
--- a/src/pages/policies/[policyId]/claims.svelte
+++ b/src/pages/policies/[policyId]/claims.svelte
@@ -13,7 +13,7 @@ import { Page } from '@silintl/ui-components'
 export let policyId: string
 
 $: policy = $selectedPolicy
-$: policy && loadClaimsByPolicyId(policyId)
+$: policyId && loadClaimsByPolicyId(policyId)
 $: isAdmin = checkIsAdmin($roleSelection)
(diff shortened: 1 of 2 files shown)


src/pages/policies/[policyId]/claims.svelte
```diff
@@ -2,18 +2,18 @@
...
 import { formatPageTitle } from 'helpers/pageTitle'
 import { items } from 'helpers/routes'
 import { goto, metatags } from '@roxi/routify'
 import { Page } from '@silintl/ui-components'
-import { onMount } from 'svelte'
 
 export let policyId: string
+
 $: policy = $selectedPolicy
+$: policy && loadClaimsByPolicyId(policyId)
```
Comment (Schparky) [inline src/pages/policies/[policyId]/claims.svelte:16]: Does this need to be dependent on `policyId`?

Comment (hobbitronics) [reply src/pages/policies/[policyId]/claims.svelte:16]: yes, thanks

Comment (hobbitronics) [reply src/pages/policies/[policyId]/claims.svelte:16]: It still works, but I meant to use policyId

Schparky (approved):

PR #421, opened 2022-01-11
filter out non actionable cards on admin/home

hobbitronics:
image
image

### DIFF
--- a/src/data/claims.ts
+++ b/src/data/claims.ts
@@ -82,6 +82,10 @@ export function isClaimItem(obj: any): obj is ClaimItem {
   return (obj as ClaimItem)?.claim_id !== undefined
 }
 
+export function isClaim(obj: any): obj is Claim {
+  return (obj as Claim)?.claim_items?.length !== undefined
+}
+
 export type Claim = {
   claim_files: ClaimFile[]
   claim_items: ClaimItem[]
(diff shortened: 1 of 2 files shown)


src/data/claims.ts
```diff
@@ -82,6 +82,10 @@ export function isClaimItem(obj: any): obj is ClaimItem {
   return (obj as ClaimItem)?.claim_id !== undefined
 }
 
+export function isClaim(obj: any): obj is Claim {
```
Schparky, line 85: How does this syntax work? Specifically, what is the `obj is Claim` part?

hobbitronics: I think it is just saying the function returns a boolean re: the type of the object is said object. I just copied Russel's function though so I don't have a deep understanding.

Schparky: I found an answer on SO. Interesting feature

src/pages/admin/home.svelte
```diff
@@ -1,29 +1,39 @@
...
 
+$: filteredChanges = $roleSelection && $recentChanges.filter(recentChangesFilter)
+
 onMount(() => {
   // TODO: recent activity does not include old items (those over 1 week) that are still unresolved.
   loadRecentActivity()
 })
 
+const recentChangesFilter = (change: RecentChange): boolean => {
+  const item: Claim | PolicyItem = isRecentClaim(change) ? change.Claim : change.Item
```
Schparky, line 22: If I understand correctly, this variable (`item`) should be named something more general. Maybe `obj` or `claimOrItem`.

Schparky (approved):
Just a minor edit for clarity, but looks good otherwise.

src/pages/admin/home.svelte
hobbitronics: agreed

PR #423, opened 2022-01-13
update ui-components and run npm update

hobbitronics:
# Changed
- updated ui-components to 3.10.6 (removes a svelte a11y warning)
- ran npm update


Schparky (approved):

hobbitronics:
for some reason one of the dependencies now requires tslib (which causes additional warnings when installed) to be a peer dependency so it is requiring us to install it. I have not sorted out which one it is yet.

PR #428, opened 2022-01-20
add showApp store to hide app

hobbitronics:
Check for public routes to show the app
Check if user logged in to hide the app


hobbitronics:
Will admins always have a policy or should we check something else on the user?

Schparky:
> Will admins always have a policy or should we check something else on the user?

All users should have a policy since one is created when a user is created. Though there is a ~10 second window before that first policy is created.

Schparky (approved):

PR #429, opened 2022-01-21
New itemform design and fix a bug

hobbitronics:
# Changed
- ItemForm uses new design mockup
- Notes (Description) not required anymore

# Fixed
- When adding a dependent then saving for later, the new dependent wasn't selected when going back to edit
image

### DIFF
--- a/src/components/Description.svelte
+++ b/src/components/Description.svelte
@@ -2,7 +2,6 @@
 .description {
   color: rgba(0, 0, 0, 0.5);
   display: block;
-  padding: 0.5rem 1rem;
 }
 </style>
 
--- a/src/components/Description.svelte
+++ b/src/components/Description.svelte
@@ -1,8 +1,1 @@
-<style>
-.description {
-  color: rgba(0, 0, 0, 0.5);
-  display: block;
-}
-</style>
-
-<span class="description"><slot /></span>
+<span class="d-block mdc-theme--neutral"><slot /></span>
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -5,12 +5,13 @@ import Description from '../Description.svelte'
 import MakeAndModelModal from 'MakeAndModelModal.svelte'
 import ItemDeleteModal from '../ItemDeleteModal.svelte'
 import SelectAccountablePerson from '../SelectAccountablePerson.svelte'
-import { MAX_INPUT_LENGTH as maxlength, MAX_TEXT_AREA_LENGTH } from 'components/const'
+import { MAX_TEXT_AREA_LENGTH } from 'components/const'
 import type { AccountablePersonOptions } from 'data/accountablePersons'
 import { ItemCoverageStatus, PolicyItem } from 'data/items'
 import { categories, loadCategories, initialized as catItemsInitialized } from 'data/itemCategories'
+import TextFieldWithLabelAndDescription from '../TextFieldWithLabelAndDescription.svelte'
 import { assertHas } from '../../validation/assertions'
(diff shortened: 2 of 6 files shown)


src/components/Description.svelte
```diff
@@ -1,9 +1 @@
-<style>
-.description {
-  color: rgba(0, 0, 0, 0.5);
-  display: block;
-  padding: 0.5rem 1rem;
-}
-</style>
-
-<span class="description"><slot /></span>
+<span class="d-block mdc-theme--neutral"><slot /></span>
```
Schparky, line 1: Is it still worth having this as a component?

src/components/forms/ItemForm.svelte
```diff
@@ -164,43 +163,53 @@ const setInitialValues = (user: User, item: PolicyItem) => {
...
-      rows="4"
-    />
-    <Description>For personal use.</Description>
+    <span class="label">Model</span>
+    <TextField {maxlength} bind:value={model} />
+    <Description>For example, “iPhone 10 Max 64 GB” or “A1921”</Description>
   </p>
   <p>
-    <TextField {maxlength} label="Unique identifier" bind:value={uniqueIdentifier} />
+    <span class="label"> Unique identifier </span>
```
Schparky, line 197: I don't know, but it seems like bad practice to separate the label from the input like this. What do you think?

src/components/forms/ItemForm.svelte
```diff
@@ -164,43 +163,53 @@ const setInitialValues = (user: User, item: PolicyItem) => {
...
+    <TextField {maxlength} bind:value={uniqueIdentifier} />
     <Description>Optional. Serial number, IMEI, service tag, VIN</Description>
```
Schparky, line 199: Since most of the fields have a description, could/should it be wrapped in a component? Particularly so, if the label is also outside the TextField.

Schparky (approved):
Looks good, just a couple questions

src/components/forms/ItemForm.svelte
hobbitronics: I can't find it anymore but @merchako had referenced on article on placeholders and floating labels and how they are less intuitive than a regular label. They said material had prioritized form over function, so while this isn't the way Google does forms it ends up being more straightforward for the user.

Schparky: I wasn't commenting on the layout/rendering so much as the markup. I take it you can't get the desired output while still using the `label` attribute. So I guess that points to my other comment. Wrap the `span` and the `TextField` in one component. Maybe even put the `span` inside the `TextField`.

src/components/Description.svelte
hobbitronics: maybe not, but it does make the code look a bit cleaner and give some semantic meaning.

src/components/forms/ItemForm.svelte
hobbitronics: are you suggesting getting rid of the `Description' component?

src/components/Description.svelte
Schparky: I'm fine with keeping it then.

src/components/forms/ItemForm.svelte
Schparky: No, I'm suggesting to include it in `TextField` or create another component (`TextFieldWithDescription` ?) if that doesn't work.

src/components/forms/ItemForm.svelte
hobbitronics: I would like to keep the span separate from the Textfield component unless we know that's the only kind of label we will be using. I can however just wrap them together in a new components along with Description.

Schparky (approved):
looks good, thanks

PR #431, opened 2022-01-26
add policy renewals page for Steward

Schparky:
Performance is lacking when processing data on the scale of the production database, so work is yet to be done here.

image

### DIFF
--- a/src/helpers/sort.ts
+++ b/src/helpers/sort.ts
@@ -1,6 +1,3 @@
-import type { Claim } from 'data/claims'
-import type { PolicyItem } from 'data/items'
-
 export const compare = (first: string, second: string, ascending: boolean): number => {
   if (ascending) {
     if (first < second) return -1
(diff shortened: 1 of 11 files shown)


src/helpers/sort.ts
```diff
@@ -1,6 +1,3 @@
-import type { Claim } from 'data/claims'
-import type { PolicyItem } from 'data/items'
```
hobbitronics, line 2: I am curious what prompted replacing these with any down below.

src/pages/admin/renewals.svelte
```diff
@@ -0,0 +1,51 @@
...
+  }
+</script>
+
+<style>
+
+</style>
+
+<Page>
+  <div class="flex justify-between align-items-center">
+    <Button raised on:click={() => onUpdate()}>Update</Button>
```
hobbitronics, line 46:
```suggestion
    <Button raised on:click={onUpdate}>Update</Button>
```
Does this work?

src/pages/admin/renewals.svelte
```diff
@@ -0,0 +1,51 @@
...
+    <Button disabled={!reportFile?.url} on:click={() => onDownload()}>Download</Button>
```
hobbitronics, line 47:
```suggestion
    <Button disabled={!reportFile?.url} on:click={onDownload}>Download</Button>
```

same as above

hobbitronics (approved):
looks good, just a few questions.

src/helpers/sort.ts
Schparky: I didn't want to add another type to the list. Plus, it seems to make sense that a helper like this shouldn't be limited to a list of types.

src/pages/admin/renewals.svelte
Schparky: same as above

PR #437, opened 2022-02-01
create a new "ledger reports" page 

Schparky:
image

image

### DIFF
--- a/src/components/FileLink.svelte
+++ b/src/components/FileLink.svelte
@@ -1,0 +1,20 @@
+<script type="ts">
+  import type { CoverFile } from 'data/file'
+  import { onMount } from 'svelte'
+
+  export let file: CoverFile
+
+  onMount(() => {
+    setTimeout(onTimeout, new Date(file.url_expiration) - new Date())
+  })
+
+  let expired = new Date(file.url_expiration) < new Date()
+  const onTimeout = () => {
+    expired = true
+  }
+
+  $: url = !expired && file.url || null
+  $: filename = file?.name || '-'
+</script>
+
+<a href={url}>{filename}</a>
--- a/src/data/ledger.ts
+++ b/src/data/ledger.ts
@@ -61,10 +61,10 @@ export async function getPolicyRenewals(): Promise<LedgerReport> {
   return createLedgerReport(LedgerReportType.annual, new Date().toISOString().split('T')[0])
 }
 
-export async function createLedgerReport(reportType: LedgerReportType, date: string): Promise<LedgerReport> {
+export async function createLedgerReport(type: LedgerReportType, date: string): Promise<LedgerReport> {
   const params: CreateLedgerReportInput = {
-    date: date,
-    type: reportType,
+    date,
+    type,
   }
   return await CREATE('ledger-reports', params)
 }
--- a/src/data/ledger.ts
+++ b/src/data/ledger.ts
@@ -1,12 +1,31 @@
 import type { PolicyType } from 'data/policies'
 import type { CoverFile } from 'data/file'
-import { GET, CREATE } from './index'
+import { CREATE, GET, UPDATE } from './index'
 
 export type LedgerReport = {
+  id: string
   file: CoverFile
-  ledger_entries: LedgerEntry[]
+  type: string
+  date: string
+  created_at: string
+  updated_at: string
 }
 
+export type CreateLedgerReportInput = {
+  // Report date, e.g. return the ledger entries prior to the given date. Details vary by the report type.
+  date: string
+  type: LedgerReportType
+}
+
+// Ledger report types:
+// + `monthly` - Return all ledger entries not yet reconciled, up to the beginning of the given date.
+// + `annual` - Return the policy renewal entries for the year of the given date.
+export enum LedgerReportType {
+  annual = 'Annual',
+  monthly = 'Monthly',
+}
+
+// LedgerEntry is not currently in use, but will hold it for future use
 export type LedgerEntry = {
   id: string
   policy_id: string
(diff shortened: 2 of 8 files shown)


src/components/FileLink.svelte
```diff
@@ -0,0 +1,20 @@
...
+  onMount(() => {
+    setTimeout(onTimeout, new Date(file.url_expiration) - new Date())
+  })
+
+  let expired = new Date(file.url_expiration) < new Date()
+  const onTimeout = () => {
+    expired = true
+  }
+
+  $: url = !expired && file.url || null
```
hobbitronics, line 16: I thought we were trying to keep `null out of our front end code when possible`, would moving it to the data layer be better?

src/data/ledger.ts
```diff
@@ -29,14 +48,27 @@ export type LedgerEntry = {
   updated_at: string /*Date*/
 }
 
+export async function getLedgerReports(): Promise<LedgerReport[]> {
+  return await GET('ledger-reports')
+}
```
hobbitronics, line 53: what do you think about adding a LedgerReports store and updating it in the function rather than returning them? That's how we've being handling most things in the data layer.

src/data/ledger.ts
```diff
@@ -29,14 +48,27 @@ export type LedgerEntry = {
...
-  return await GET('ledger?report-type=annual')
+  await processPolicyRenewals()
+  return createLedgerReport(LedgerReportType.annual, new Date().toISOString().split('T')[0])
 }
 
-export async function downloadPolicyRenewals(): Promise<any> {
-  return await GET('ledger?report-type=annual')
+export async function createLedgerReport(reportType: LedgerReportType, date: string): Promise<LedgerReport> {
+  const params: CreateLedgerReportInput = {
+    date: date,
```
hobbitronics, line 66:
```suggestion
    date,
```
Maybe you don't like the way this looks, either way is fine.

src/pages/admin/_components/CreateReportModal.svelte
```diff
@@ -0,0 +1,78 @@
+<script lang="ts">
+  import {Button, Form, Select} from '@silintl/ui-components'
+  import {CreateLedgerReportInput, LedgerReportType} from 'data/ledger'
+  import {createEventDispatcher} from 'svelte'
+  import DateInput from '../../../components/DateInput.svelte'
+  import Modal from '../../../components/mdc/Modal.svelte'
```
hobbitronics, line 6: sort imports

src/pages/admin/_components/CreateReportModal.svelte
```diff
@@ -0,0 +1,78 @@
...
+
+  export let modalOpen = false
+
+  const dispatch = createEventDispatcher()
```
hobbitronics, line 10: up to you, but you can find examples in the code of how to type what is dispatched here. for example `const dispatch = createEventDispatcher<{ 'submit': void }>()`

src/pages/admin/_components/CreateReportModal.svelte
```diff
@@ -0,0 +1,78 @@
...
+>
+  {#if modalOpen}
+    <div class="w-100">
+      <Form on:submit={onSubmit}>
+        <p>
+          <span class="mdc-bold-font">Report Type</span>
+          <Select label="Input" options={reportTypes} selectedID='Monthly' on:change={onSelectType} />
+        </p>
+        <p>
+          <span class="mdc-bold-font">Report Date</span>
```
hobbitronics, line 66: I remember Alex commenting on one of our forms that `<label>` is semantically more correct, but I had an issue when trying to use a `<div>` inside the `<p>` elements so maybe we are misusing `<p>`?

src/pages/admin/_components/CreateReportModal.svelte
```diff
@@ -0,0 +1,78 @@
...
+</script>
+
+<style>
+  .float-right {
+    float: right;
+  }
+
+  .form-button {
+    margin: 0.5rem;
+  }
```
hobbitronics, line 48: can these classes be combined?

hobbitronics (approved):
Nice work. Assuming you sort the imports and consider my comments/suggestions looks great.

src/components/FileLink.svelte
Schparky: Is `undefined` better? This implementation requires one or the other in order to not include the `href` attribute. Maybe there's a different way altogether, but I didn't find it.

src/data/ledger.ts
Schparky: I didn't feel it was worth the extra complexity. What is the benefit of doing that?

src/pages/admin/_components/CreateReportModal.svelte
Schparky: Thanks for the tip.

src/pages/admin/_components/CreateReportModal.svelte
Schparky: Cool

src/pages/admin/_components/CreateReportModal.svelte
Schparky: For me, it was because the form element is inside a component. I was too lazy to find a way through that.

src/components/FileLink.svelte
hobbitronics: I think sticking to undefined is better according to Russel. When you have to manage both null and undefined it creates more complex logic.

src/data/ledger.ts
hobbitronics: It would only be beneficial if you are using the store in multiple places which it probably is not.

PR #441, opened 2022-02-02
Release

Schparky:
includes:
#437 - new Ledger Reports page `/admin/reports` replaces /admin/renewals
#438 - redirect on `/admin/policies` to add qs params without adding extra page to browser history
#439 - don't auto-capitalize breadcrumbs
#440 - "NOT PRODUCTION" indicator on staging/dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "material-components-web": "^10.0.0",
     "npm-run-all": "^4.1.5",
     "postcss": "^8.3.11",
-    "prettier": "^2.4.1",
+    "prettier": "2.5.1",
     "prettier-plugin-svelte": "^2.4.0",
     "qs": "^6.10.1",
     "rimraf": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "scripts": {
     "prebuild:dev": "npm run clean:build",
     "build:dev": "rollup -c -w",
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "scripts": {
     "prebuild:dev": "npm run clean:build",
     "build:dev": "rollup -c -w",
-    "prebuild:prod": "echo GA_MEASUREMENT_ID=${GA_MEASUREMENT_ID} >> .env; echo FRESHDESK_ID=${FRESHDESK_ID} >> .env; echo API_HOST=${API_HOST} >> .env; npm run clean:build",
+    "prebuild:prod": "echo CF_PAGES_BRANCH=${CF_PAGES_BRANCH} >> .env; echo GA_MEASUREMENT_ID=${GA_MEASUREMENT_ID} >> .env; echo FRESHDESK_ID=${FRESHDESK_ID} >> .env; echo API_HOST=${API_HOST} >> .env; npm run clean:build",
```
hobbitronics, line 6: this is a built in CF variable. See https://developers.cloudflare.com/pages/platform/build-configuration#environment-variables
I don't think we need to set it ourselves, it just looks at the branch name.

hobbitronics (changes requested):
Unless I am mistaked we don't need to set the branch env var.

package.json
Schparky: It didn't work until I added that. I suspect it's because the `process` polyfill doesn't actually look at the environment, but just loads the .env file.

hobbitronics (approved):

PR #453, opened 2022-02-16
Add Watermark component

hobbitronics:
- Add Watermark.svelte
- replace text with Watermark
 
image


Schparky (approved):
Cool stuff! Now we can go totally retro and have garish backgrounds on every page. Customers will love it. 
image

PR #455, opened 2022-02-18
Update UI components

hobbitronics:
- Update UI components to 4.0.0
- update interfaces


src/components/AppDrawer.svelte
```diff
@@ -26,6 +26,12 @@ onMount(() => {
 })
 
 $: drawerWidth = `${drawerEl?.offsetWidth || 0}px`
+$: currentUrl = window.location.pathname
```
Schparky, line 29: Would it be better to use routify's helper? https://v1.routify.dev/docs/helpers#url

Schparky (approved):

src/components/AppDrawer.svelte
hobbitronics: yes, and the reactive statement isn't updating anyways, that's why `$beforeUrlChange` is needed

PR #459, opened 2022-02-25
allow steward/signator to change coverage amount

Schparky:


src/validation/assertions.ts
```diff
@@ -18,6 +18,12 @@ export function assertIsLessThan(value: any, high: any, errorMessage: string): v
   }
 }
 
+export function assertIsLessOrEqual(value: any, high: any, errorMessage: string): void {
```
hobbitronics, line 21:
```suggestion
export function assertIsLessOrEqual(value: number | string, high: number | string, errorMessage: string): void {
```

Assuming the parameters are only possibly numbers the `| string` part won't be necessary, but Javascript...

hobbitronics (approved):
Looks good, just one suggestion.

PR #464, opened 2022-03-09
add invites to policy settings

hobbitronics:
image

### DIFF
--- a/src/data/policies.ts
+++ b/src/data/policies.ts
@@ -87,12 +87,14 @@ const updatePolicyStore = (changedPolicy: Policy) => {
     if (i === -1) {
       policies.push(changedPolicy)
     } else {
-      // the policies returned from the get policies list never include claims or invites on them
+      // the policies returned from the get policies list never include claims, dependents, members or invites on them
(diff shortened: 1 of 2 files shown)


src/data/policies.ts
```diff
@@ -80,11 +87,12 @@ const updatePolicyStore = (changedPolicy: Policy) => {
     if (i === -1) {
       policies.push(changedPolicy)
     } else {
-      // the policies returned from the get policies list never include claims on them
+      // the policies returned from the get policies list never include claims or invites on them
```
Schparky, line 90: Don't know if it's worth mentioning, but dependents are also omitted.

Schparky (approved):

src/data/policies.ts
hobbitronics: and what about member?

hobbitronics: Added the code to keep both

Schparky: > and what about members?

Members come with PoliciesList.

hobbitronics: thanks

PR #465, opened 2022-03-09
Delete/Revoke a claim

hobbitronics:
## Add
- RevokeModal
- delete claim button

image
image

### DIFF
--- a/src/components/ClaimActions.svelte
+++ b/src/components/ClaimActions.svelte
@@ -121,10 +121,10 @@ const onDeny = () => dispatch('deny', message)
   {/if}
 {/if}
 {#if isMemberOfPolicy}
-  {#if isEditable}
-    <Button raised on:click={on('edit')}>edit claim</Button>
-    <Button outlined on:click={on('delete')}><span class="material-icons">delete</span></Button>
-  {/if}
+  <Button raised disabled={!isEditable} on:click={on('edit')}>edit claim</Button>
+  <Button outlined disabled={!claim.is_removable} on:click={on('delete')}>
+    <span class="material-icons">delete</span>
+  </Button>
 
   {#if showSubmit}
     <Button raised disabled={noFilesUploaded} on:click={on('submit')}>Submit claim</Button>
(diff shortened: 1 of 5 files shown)


src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -152,6 +156,23 @@ const onSubmit = async () => {
   await submitClaim(claimId)
   setNotice(didNeedFile ? `Added replacement cost and ${oldUploadLabel}` : 'Submitted changes')
 }
+const onClosed = async (e: CustomEvent): Promise<void> => {
```
Comment (Schparky) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:159]: Given the size of this page, should `onClosed` be more descriptive? It's less than obvious what is being closed.

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -353,12 +380,20 @@ const isFileUploadedByPurpose = (purpose: ClaimFilePurpose, files: ClaimFile[]):
...
+      />
 
       {#if showUploadButton}
         <Button raised disabled={noFilesUploaded} on:click={onSubmit}>{uploadLabelForButton}</Button>
       {/if}
       <br />
     </Row>
   {/if}
+
+  <RevokeModal {claim} open={revokeModalOpen} on:closed={onClosed} />
```
Comment (Schparky) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:398]: This should be disabled if `is_removable` is `false`

Schparky:
> * evokeModal
> * delete claim button

I often have to do a double-take on these when you omit the "Add". I generally repeat the "Add" or "Remove" on each bullet.

src/components/ClaimActions.svelte
```diff
@@ -122,7 +122,8 @@ const onDeny = () => dispatch('deny', message)
 {/if}
 {#if isMemberOfPolicy}
   {#if isEditable}
-    <Button on:click={on('edit')} outlined>Edit claim</Button>
+    <Button raised on:click={on('edit')}>edit claim</Button>
+    <Button outlined on:click={on('delete')}><span class="material-icons">delete</span></Button>
```
Schparky, line 126: Should this button be disabled as well if `is_removable` is false?

Schparky (approved):

src/components/ClaimActions.svelte
hobbitronics: perhaps and maybe it it doesn't need to be inside the control block then.

PR #473, opened 2022-03-16
Fix invites not updating

hobbitronics:
- load policy after inviting
- check for the email already existing in invites, member and user


src/pages/policies/[policyId]/settings.svelte
```diff
@@ -171,12 +171,15 @@ const onRemoveModal = (event: CustomEvent<string>) => {
...
+  [...invites, ...policyMembers, $user].every((el) => el.email !== email)
 const onSubmitModal = async (event: CustomEvent<DependentFormData>) => {
   let { id, name, relationship, country, childBirthYear, permissions, email, message } = event.detail
 
   if (permissions === 'can-edit') {
-    // TODO: Figure out if we've already sent an invite or not
-    await invitePolicyMember(policyId, name, email, message)
+    hasNotBeenInvited(email)
+      ? await invitePolicyMember(policyId, name, email, message)
+      : setNotice('This person is a member of this policy or has already been invited')
```
Comment (Baggerone) [inline src/pages/policies/[policyId]/settings.svelte:182]: I don't know how you want to word this, but technically (from the BE's point of view) it would be more accurate to say 'This person is already a member of a household policy or ...'

src/data/policy-members.ts
```diff
@@ -1,6 +1,7 @@
 import { GET, CREATE as POST } from './index'
-import { derived, writable } from 'svelte/store'
+import { loadPolicy } from './policies'
```
Schparky, line 2: This looks like a new circular dependency. If at all possible, let's try to avoid this.

src/pages/policies/[policyId]/settings.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/settings.svelte:182]: This settings page is for both household and team policies. I could make the message dynamic based on the type of policy though.

src/data/policy-members.ts
hobbitronics: I'm not actually getting a warning from this for some reason. ```[build:dev ] > build:dev
[build:dev ] > rollup -c -w
[build:dev ] 
[build:dev ] [Routify] Watching src/pages
[build:dev ] rollup v2.59.0
[build:dev ] bundles src/main.ts → dist/bundle.js...
[build:dev ] [svelte-preprocess] Deprecation notice: using the "type" attribute is no longer recommended and will be removed in the next major version. Please use the "lang" attribute instead.
[build:dev ] (node:20203) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at /Users/mike/projects/cover-ui/node_modules/tslib/package.json.
[build:dev ] Update this package.json to use a subpath pattern like "./*".
[build:dev ] (Use `node --trace-deprecation ...` to show where the warning was created)
[build:dev ] LiveReload enabled
[build:dev ] (!) Plugin dotenv: @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
[build:dev ] (!) Circular dependency
[build:dev ] src/data/user.ts -> src/data/policies.ts -> src/data/user.ts
[build:dev ] created dist/bundle.js in 10.6s```

Schparky: Yeah, I checked it out and was surprised to see it wasn't identified. It's still a circular dependency, even though the tool didn't spot it.

src/pages/policies/[policyId]/settings.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/settings.svelte:183]: maybe pure types don't count

src/data/policy-members.ts
hobbitronics: maybe types don't count

src/pages/policies/[policyId]/settings.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/settings.svelte]: The error message for a household policy already having the email is straight from the BE. I tried to catch it and display a different message but it just ends up showing both.

src/data/policy-members.ts
Schparky: That makes sense. Thanks for digging into it.

Schparky (approved):

PR #475, opened 2022-03-17
add admin breadcrumbs to policy

hobbitronics:
It seems I forked this branch off the previous one, so it might make more sense of it after merging the previous branch. Or the last [commit](https://github.com/silinternational/cover-ui/pull/475/commits/47477c9b84118263221a0eb39130b1a855492003) would be the changes after merging.

image


hobbitronics:
> It's often helpful to create PRs like this as a merge into the previous feature branch so it doesn't clutter the diff with unrelated changes. Then when the previous branch is merged (and deleted), Github automatically updates the target branch.

Good tip. Thanks

PR #480, opened 2022-03-19
Add strikes to policies

hobbitronics:
- add basic ability to add and view strikes on a policy
- show and explain strikes to customer (basics for now)

Admin view
image

customer view
image

Hide the form
image

image

PR #482, opened 2022-03-22
Release - truncate long policy names, add/view strikes, delete dependents, filter reports, cleanup code/warnings, RemoveDependentModal, align banners

hobbitronics:
## Includes
- <https://github.com/silinternational/cover-ui/pull/485>
- <https://github.com/silinternational/cover-ui/pull/484>
- <https://github.com/silinternational/cover-ui/pull/483>
- <https://github.com/silinternational/cover-ui/pull/481>
- <https://github.com/silinternational/cover-ui/pull/480>
- <https://github.com/silinternational/cover-ui/pull/479>
- <https://github.com/silinternational/cover-ui/pull/478>
- <https://github.com/silinternational/cover-ui/pull/477>


src/components/Strikes.svelte
```diff
@@ -0,0 +1,64 @@
...
+  createPolicyStrike(policy.id, strikeDescription)
+  strikeDescription = ''
+}
+</script>
+
+{#if strikes.length || userIsAdmin}
+  <h4>Strikes</h4>
+  {#if !userIsAdmin}
+    <p>
+      <!-- TODO: Add strike rules to terms when finalized -->
```
Schparky, line 31: Should this TODO be resolved before release?

Schparky (commented):
Might want to update the terms page with information about strikes. It could probably wait a bit, but I'm not sure what would serve as a reminder if this release PR is merged.

src/data/items.ts
```diff
@@ -272,6 +272,10 @@ export async function deleteItem(policyId: string, itemId: string): Promise<any>
 
 export const itemBelongsToPolicy = (policyId: string, item: PolicyItem): boolean => item.policy_id === policyId
 
+export const isDependentOnItemsByPolicyId = (dependentId: string, policyId: string): boolean => {
+  return get(itemsByPolicyId)[policyId]?.every((item) => item.accountable_person?.id === dependentId)
```
forevermatt, line 276: Wouldn't `some()` (not `every()`) be the appropriate function call here?

Worded another way, do you want to know whether the dependent is the accountable person for **any** of the items on the policy, or whether the dependent is the accountable person for **aall** of the items on the policy?

I had the impression that the question was whether any of them were tied to the dependent (since you can't delete the dependent while there are any items associated with them).

src/pages/admin/reports.svelte
```diff
@@ -39,7 +39,8 @@ async function createReport(e: CustomEvent) {
       <Datatable.Header.Item>File</Datatable.Header.Item>
     </Datatable.Header>
     <Datatable.Data>
-      {#each ledgerReports as report (report.id)}
+      <!-- TODO: remove this filter when BE filters out policy reports -->
+      {#each ledgerReports.filter((r) => !r.file?.name.includes('policy')) as report (report.id)}
```
forevermatt, line 43: It may be worth creating a simple filter method named appropriately to clarify what the purpose of this filter is. It's not obvious to me _what_ the filter is doing, though I can see _how_ it's doing it.

Would this be it?
```
const isAdminReport = (report: LedgerReport) => !report.file?.name.includes('policy')
```

hobbitronics: yes, but since this is a temporary fix would it be sufficient to describe its purpose in the comment? Mainly the benefit of this would be to keep the removal down to just one line.

forevermatt: I think it will still be better to extract a simple method named appropriately. That way the code stays readable, whether that particular method is only need for a day or a year. (One of my code-smell mental filters: If you need a comment, could you instead improve the code so that it's so readable no comment is necessary?)

My 2 cents.

Schparky (approved):

forevermatt (approved):

PR #488, opened 2022-03-30
Remove labels claim form and fix bug

hobbitronics:
- Stop using built in TextInput labels
- fix ClaimId view bug for FMV claims where repairable is selected but estimate is too high and unable to submit

image

image

### DIFF
--- a/src/business-rules/claim-payout-amount.ts
+++ b/src/business-rules/claim-payout-amount.ts
@@ -45,10 +45,9 @@ export const isUnrepairableOrTooExpensive = (
 }
 
 export const isEvidenceNeeded = (claimItem: ClaimItem, claimStatus: ClaimStatus): boolean => {
-  const willNeedEvidence = claimItem.fmv > 0 || claimItem.repair_estimate > 0
-  const repairCostIsNotTooHigh = !isRepairCostTooHigh(claimItem.repair_estimate, claimItem.fmv)
+  const willNeedEvidence = [PayoutOption.FMV, PayoutOption.Repair].includes(claimItem.payout_option)
   const canProvideEvidenceNow = [ClaimStatus.Draft].includes(claimStatus)
-  return willNeedEvidence && repairCostIsNotTooHigh && canProvideEvidenceNow
+  return willNeedEvidence && canProvideEvidenceNow
 }
 
 export const getFilePurpose = (claimItem: ClaimItem, needsReceipt: boolean): ClaimFilePurpose | '' => {
--- a/src/components/forms/ClaimForm.svelte
+++ b/src/components/forms/ClaimForm.svelte
@@ -230,7 +230,7 @@ const unSetReplaceEstimate = () => {
     </p>
     <p>
       <span class="header">What happened?</span>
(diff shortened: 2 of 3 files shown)


src/business-rules/claim-payout-amount.ts
```diff
@@ -45,10 +45,9 @@ export const isUnrepairableOrTooExpensive = (
 }
 
 export const isEvidenceNeeded = (claimItem: ClaimItem, claimStatus: ClaimStatus): boolean => {
-  const willNeedEvidence = claimItem.fmv > 0 || claimItem.repair_estimate > 0
-  const repairCostIsNotTooHigh = !isRepairCostTooHigh(claimItem.repair_estimate, claimItem.fmv)
+  const willNeedEvidence = [PayoutOption.FMV, PayoutOption.Repair].includes(claimItem.payout_option)
```
forevermatt, line 48: So from the code change it looks like the repair estimate being more or less than 70% of the fair market value no longer affects whether evidence is needed. Is that true across the board, or were we just hitting an edge case that needed additional logic?

hobbitronics: I don't think it matters to either of the 2 places (claimCard and [claimId]) this function are used. If the repair cost were too high we wouldn't have let them continue as a repair in the claimForm.

hobbitronics: I can't remember why we had it in there but maybe it was trying to do logic that is now in the form?

forevermatt: I think we had it in there because that was part of the definition of when evidence was needed. If that's changed, I'm fine with us removing it as this PR does. If that definition hasn't changed, I'd be hesitant to proceed with this. It's possible we need a different function, or a more specified function (e.g. `isEvidenceNeededInSuchAndSuchACase()`).

forevermatt: I don't think we ever need evidence for the FMV payout option (if I'm reading our Claim Payout Amount workflow properly... I'll Slack you the link so you have it handy).

I think this bug might merit setting up some unit tests, to help us get the logic right in this `claim-payout-amount.ts` file's functions.

Do you want to revert this file's changes in this PR, so that we can get the MoneyInput fixes pushed through and come back to this function's fixes separately?

forevermatt (commented):
I suggest splitting these changes to two separate PRs (see individual comments).

src/business-rules/claim-payout-amount.ts
merchako: Darryl writes, 
> FMV Payout is a financial reimbursement.  So we do need a "receipt". Typically we would ask for a link to a website selling an item of the same "like kind and quality" and recommend eBay sold or Amazon used.  FMV is not arbitrary, so something in print is helpful and we generally take what the member provides.

Darryl Johnson and the mockup have it right. That [Claim Payout Amount](https://www.figma.com/file/5EELttFC6weg0HF4IfxBkF/Cover%3A-Payout-%26-Breadcrumbs?node-id=5%3A222) diagram is only for calculating the payout. It is not a UX flow.

PR #489, opened 2022-03-30
prevent customer from manually viewing admin page

hobbitronics:
- If customer types in an admin url show them something else
image


Schparky (approved):
Good fix, thanks for catching this.

PR #494, opened 2022-03-31
Add ledgerReports store and renew report url

hobbitronics:
- add on:expired to FileLink
- add ledgerReports store and UpdateLedgerReports
- get new report when url expires with on:expired
- get Policy Reports from Policy object
image

### DIFF
--- a/src/components/CustomerReport.svelte
+++ b/src/components/CustomerReport.svelte
@@ -12,16 +12,6 @@ let modalOpen = false
 
 $: $policyLedgerReports = policy.ledger_reports || []
 
-async function getReportAndUpdateReports(reportId: string) {
-  const report = await getLedgerReportById(reportId)
-  const index = $policyLedgerReports.findIndex((r) => r.id === report.id)
-  if (index !== -1) {
-    $policyLedgerReports[index] = report
-  } else {
-    $policyLedgerReports = [...$policyLedgerReports, report]
-  }
-}
-
 async function createReport(e: CustomEvent) {
   const reportType: LedgerReportType = e.detail.type
   const { month, year } = e.detail.date
@@ -52,7 +42,7 @@ async function createReport(e: CustomEvent) {
           <Datatable.Data.Row.Item>{formatFriendlyDate(report.date)}</Datatable.Data.Row.Item>
           <Datatable.Data.Row.Item>{formatDateAndTime(report.created_at)}</Datatable.Data.Row.Item>
           <Datatable.Data.Row.Item>
-            <FileLink on:expired={() => getReportAndUpdateReports(report.id)} file={report.file} />
+            <FileLink on:expired={() => getLedgerReportById(report.id, true)} file={report.file} />
           </Datatable.Data.Row.Item>
         </Datatable.Data.Row>
       {/each}
--- a/src/components/CustomerReport.svelte
+++ b/src/components/CustomerReport.svelte
@@ -1,7 +1,13 @@
 <script lang="ts">
 import CreateCustomerReportModal from './CreateCustomerReportModal.svelte'
 import type { Policy } from 'data/policies'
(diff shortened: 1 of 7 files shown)


src/components/CustomerReport.svelte
```diff
@@ -1,26 +1,25 @@
...
-  const allReports = await getLedgerReports()
-  return allReports.filter((report) => report.file?.created_by_id === $user.id && report.file?.name.includes('policy'))
+async function getReportAndUpdateReports(reportId: string) {
+  const report = await getLedgerReportById(reportId)
+  const index = $policyLedgerReports.findIndex((r) => r.id === report.id)
+  if (index !== -1) {
+    $policyLedgerReports[index] = report
+  } else {
+    $policyLedgerReports = [...$policyLedgerReports, report]
+  }
```
forevermatt, line 22: Something about doing this here feels odd, as though the logic for retrieving a report and adding-or-updating it in our `$policyLedgerReports` store belongs in a data file, not in a UI component.

What would you think of moving this into a datafile (and using a function like our other "update(StoreName)Store" to internally handle the add-or-update logic)?

That could potentially simplify this CustomerReport component's data call to something like `loadReport(reportId)`, which would behind the scenes update the `$policyLedgerReports` store and thereby indirectly update this component's content.

However, I also don't quite understand why `$policyLedgerReports` is being overwritten in this component, so there is quite possibly something about the intended use of this component that I'm just not quite grasping yet. Feel free to counter my suggestions with an explanation of why this existing approach is appropriate.

src/data/policies.ts
```diff
@@ -101,6 +103,8 @@ const updatePolicyStore = (changedPolicy: Policy) => {
         claims: changedPolicy.claims == null ? policies[i].claims : changedPolicy.claims,
         invites: changedPolicy.invites == null ? policies[i].invites : changedPolicy.invites,
         dependents: changedPolicy.dependents == null ? policies[i].dependents : changedPolicy.dependents,
+        ledger_reports:
```
forevermatt, line 106: The comment a few lines above (on 99) may also need to be updated, if the ledge_reports are also never included by that API endpoint.

src/pages/admin/reports/[reportId].svelte
```diff
@@ -28,6 +26,10 @@ const handleDialog = async (event: string) => {
     report = await reconcileLedgerReport(reportId)
   }
 }
+
+async function getReport() {
+  report = await getLedgerReportById(reportId)
```
Comment (forevermatt) [inline src/pages/admin/reports/[reportId].svelte:31]: Do we need to do something here to protect against calling this with an empty string, or will this page only be reached if a non-empty string is provided in the URL for the report id?

src/components/CustomerReport.svelte
hobbitronics: The problem is that `getLedgerReportById` is used by both customers and admins. If we update the stores when we call this endpoint we will also need to check if it is an admin report or customer report (no property which tells us so would have to be a parameter) and update the appropriate store.

I could move this function to the data file and call it here which would be cleaner or I could add a parameter like `isCustomerReport` to `getLedgerReportById` and then update the store in the function if true. What do you prefer?

To answer your question about policyLedgerReports, when we loaded `policies`, the `ledger_reports` property is not hydrated, but I have fixed that so it preserves those that already exist.

src/pages/admin/reports/[reportId].svelte
Comment (hobbitronics) [reply src/pages/admin/reports/[reportId].svelte:31]: I don't think there is a way to view it with an empty string so should be safe.

src/components/CustomerReport.svelte
hobbitronics: Thanks for the feedback @forevermatt. After thinking this through I came up with a painless solution to keep all that code in the datalayer.

forevermatt (approved):

PR #502, opened 2022-04-28
Preserve history on checkout

hobbitronics:
- Make checkout its own page to preserve history
- add autosaving to ItemForm with debounce
- new -> edit on ItemForm autosave so when user click 🔙  they go to edit with data preserved

some notes:
when creating a new item once category and name are filled out and 4 sec of inactivity the user is taking to edit, but page scroll and focus are not preserved in ItemForm. Is there a way around this?

I failed to capture clicking the back button in the frame and that the form does indeed start on new.
preserve-history

### DIFF
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -110,7 +110,8 @@ const validateOnSave = (formData: any) => {
 const validate = (formData: any) => {
   validateOnSave(formData)
   assertHas(formData.marketValueUSD, 'Please specify the market value')
-  assertIsLessOrEqual(formData.marketValueUSD * 100, item.coverage_amount, 'Coverage amount cannot be increased')
+  item.coverage_status !== ItemCoverageStatus.Draft &&
+    assertIsLessOrEqual(formData.marketValueUSD * 100, item.coverage_amount, 'Coverage amount cannot be increased')
   return true
 }
 
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -10,9 +10,10 @@ import type { AccountablePersonOptions } from 'data/accountablePersons'
 import { ItemCoverageStatus, PolicyItem } from 'data/items'
 import { categories, loadCategories, initialized as catItemsInitialized } from 'data/itemCategories'
 import TextFieldWithLabel from '../TextFieldWithLabel.svelte'
+import { assertHas, assertIsLessOrEqual } from '../../validation/assertions'
+import { debounce } from 'lodash-es'
 import { Button, Form, MoneyInput, Select, TextArea } from '@silintl/ui-components'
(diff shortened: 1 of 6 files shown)


src/components/forms/ItemForm.svelte
```diff
@@ -52,6 +53,14 @@ $: !$catItemsInitialized && loadCategories()
 $: itemIsDraft = item.coverage_status === ItemCoverageStatus.Draft
 $: marketValueIsDisabled = !!item.id && !itemIsDraft && !isAdmin
 $: applyBtnLabel = !item.coverage_status || itemIsDraft ? 'review and checkout' : 'save changes'
+$: make,
+  model,
+  itemDescription,
+  uniqueIdentifier,
+  marketValueUSD,
+  accountablePersonId && categoryId && name && debouncedSave()
```
forevermatt, line 61: Effectively, I think all but the last line of this are simply `let`s with no initialization. For readability's sake, what would you think of moving them out of the reactive statement (`$:`) since I think they are irrelevant to it?

forevermatt (commented):
If we want to send the user to the item-edit page after they have provided a name and category, we could also make that a separate, simple item-name-and-category form and have them click "Next" to get to the rest of the form, if we want to. That would avoid an unexpected page-scroll jump. However, what you have here seems fine for the moment.

forevermatt (approved):
I'm going ahead and approving so progress on this isn't blocked, but if you're open to making the small changes I suggested, feel free to do so before merging.

src/components/forms/ItemForm.svelte
hobbitronics: I believe the reactive statement is watching for changes to them and then runs the last line. I will double check that.

hobbitronics:
> If we want to send the user to the item-edit page after they have provided a name and category, we could also make that a separate, simple item-name-and-category form and have them click "Next" to get to the rest of the form, if we want to. That would avoid an unexpected page-scroll jump. However, what you have here seems fine for the moment.

That's a more straightforward solution, but let's see what @merchako thinks as it changes the design.

src/components/forms/ItemForm.svelte
hobbitronics: https://github.com/sveltejs/svelte/issues/4159
If you need more complete control, another common idiom is to use a sequence expression:

$: dependency_1, dependency_2, dependency_3, thing_to_run();

hobbitronics: I used this for autosaving when any of those values changes which will also prevent going from new to edit while the user is typing one of these fields.

hobbitronics:
I commented out the code and added a TODO that directs to ItemEdit from New until we get the BE Item requirements sorted out. I still the think the autosaving and having a separate page for Checkout are worthwile changes.

PR #504, opened 2022-05-02
update to ui-components 6

hobbitronics:
- update ui-components and sass
- install tslib

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@rollup/plugin-node-resolve": "^13.0.6",
     "@rollup/plugin-typescript": "^8.3.0",
     "@roxi/routify": "^2.18.3",
-    "@silintl/ui-components": "^5.0.1",
+    "@silintl/ui-components": "^6.0.0",
     "@tsconfig/svelte": "^2.0.1",
     "@types/croppie": "^2.6.1",
     "@types/gtag.js": "^0.0.8",
@@ -38,7 +38,7 @@
     "eslint": "^7.32.0",
     "eslint-plugin-svelte3": "^3.2.1",
     "lodash-es": "^4.17.21",
-    "material-components-web": "^10.0.0",
+    "material-components-web": "^13.0.0",
     "npm-run-all": "^4.1.5",
     "postcss": "^8.3.11",
     "prettier": "2.5.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "@rollup/plugin-node-resolve": "^13.0.6",
     "@rollup/plugin-typescript": "^8.3.0",
     "@roxi/routify": "^2.18.3",
-    "@silintl/ui-components": "^5.0.1",
+    "@silintl/ui-components": "^6.0.0",
```
forevermatt, line 27: Does this change mean that we will now need to fix various aspects of our markup (e.g. `class="mdc-list flex"` to `class="mdc-deprecated-listflex"`)?

hobbitronics: I think you mean if we are styling using the mdc class names for styling or building new components. I am just confused by the lack of a space before flex. Is that a typo?

forevermatt: No, that was copied from [a change included in the ui-components 6.0.0 PR](https://github.com/silinternational/ui-components/pull/141/files#diff-c1cf8407ca53b75f6fd3c95697856f858b9bfd848733441720f2e3ab1af6e2bcR127), as an example of the CSS class name changes.

forevermatt: But yes, in response to the first part of your comment: This is if we are using the mdc class names for styling anywhere in cover-ui.

forevermatt (approved):

package.json
hobbitronics: > No, that was copied from [a change included in the ui-components 6.0.0 PR](https://github.com/silinternational/ui-components/pull/141/files#diff-c1cf8407ca53b75f6fd3c95697856f858b9bfd848733441720f2e3ab1af6e2bcR127), as an example of the CSS class name changes.

I think that was a find and replace error. I will want to fix that.

hobbitronics: > Does this change mean that we will now need to fix various aspects of our markup (e.g. `class="mdc-list flex"` to `class="mdc-deprecated-listflex"`)?

We are using the mdc-list class in _typography.scss so thanks for raising this question. I will need to update that.

forevermatt (approved):

PR #505, opened 2022-05-03
add RemoveMemberModal

hobbitronics:
- added a basic modal for removing a member (similar to dependents modal)
- changed isDependentOnItemsByPolicyId to howManyItemsAccountablePersonIsOn
- fixed some imports
- added a delete button for policy-members in settings (excluding self)
image
remove-member

### DIFF
--- a/src/components/RemoveMemberModal.svelte
+++ b/src/components/RemoveMemberModal.svelte
@@ -24,7 +24,7 @@ const buttonsWithItems: Dialog.AlertButton[] = [
 $: buttons = !!numberOfItemsDependentIsOn ? buttonsWithItems : buttonsWithoutItems
 $: message = !!numberOfItemsDependentIsOn
   ? `Please remove this person from all items before removing.`
-  : `Permanently remove this person?`
+  : `Permanently remove this person? This cannot be undone.`
 
 const dispatch = createEventDispatcher<{ remove: string; cancel: string; gotoItems: string; closed: string }>()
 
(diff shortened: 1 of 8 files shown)


src/components/RemoveMemberModal.svelte
```diff
@@ -0,0 +1,41 @@
...
+  { label: 'cancel', action: 'cancel', class: 'mdc-dialog__button' },
+  { label: 'Remove', action: 'remove', class: 'error-button' },
+]
+const buttonsWithItems: Dialog.AlertButton[] = [
+  { label: 'cancel', action: 'cancel', class: 'mdc-dialog__button' },
+  { label: 'Items', action: 'gotoItems', class: 'error-button' },
+]
+$: buttons = !!numberOfItemsDependentIsOn ? buttonsWithItems : buttonsWithoutItems
+$: message = !!numberOfItemsDependentIsOn
+  ? `Please remove this person from all items before removing.`
```
Schparky, line 26: I don't see any code that blocks removal in the case where the person has items. Granted, I think the backend will just force the removal of the person from their items, but if that's the case then why bother telling the user to do it themselves? Maybe it could be something like "Permanently remove this person? If you proceed, this person will be removed from (n) covered items". @merchako do you have thoughts on this?

Schparky (approved):
Didn't find any blockers, but I did add a comment.

src/components/RemoveMemberModal.svelte
hobbitronics: We give them a different button if they have items. see line 16 and 20. They will only have the option of cancelling or going to the items view.

PR #506, opened 2022-05-05
reassign items when deleting an accountable person

hobbitronics:
- add option to reassign items when removing a policy member or dependent
Screenshot from 2022-05-05 13-46-44

### DIFF
--- a/src/components/forms/DependentForm.svelte
+++ b/src/components/forms/DependentForm.svelte
@@ -12,12 +12,13 @@ export type DependentFormData = {
 </script>
 
 <script lang="ts">
-import RadioOptions from '../RadioOptions.svelte'
-import CountrySelector from '../CountrySelector.svelte'
-import RemoveDependentModal from '../RemoveDependentModal.svelte'
+import { assignItems } from 'data/items'
 import { MAX_INPUT_LENGTH as maxlength, MAX_TEXT_AREA_LENGTH } from 'components/const'
+import CountrySelector from '../CountrySelector.svelte'
 import type { PolicyDependent } from 'data/dependents'
 import { ITEMS } from 'helpers/routes'
+import RadioOptions from '../RadioOptions.svelte'
+import RemoveDependentModal from '../RemoveDependentModal.svelte'
 import { assertEmailAddress, assertHas, assertIsLessThan, assertUnique } from '../../validation/assertions'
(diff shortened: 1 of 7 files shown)


src/components/forms/DependentForm.svelte
```diff
@@ -213,5 +218,6 @@ const onChosen = (event: CustomEvent) => (formData.country = event.detail)
     on:gotoItems={() => $goto(ITEMS)}
     on:cancel={() => (removeModalIsOpen = false)}
     on:closed={() => (removeModalIsOpen = false)}
+    on:assign(onAssign)
```
forevermatt, line 221: This looks like a function call, not like the typical `on:event={eventHandler}` syntax. Was this intentional? If so, what does it do?

hobbitronics: no, just a typo. It's interesting I didn't get an error.

src/data/items.ts
```diff
@@ -301,3 +305,21 @@ export const itemIsApproved = (item: PolicyItem): boolean => {
 export const itemIsInactive = (item: PolicyItem): boolean => {
   return item.coverage_status === ItemCoverageStatus.Inactive
 }
+
+export const assignItems = (newMemberId: string, policyId: string, selectedPolicyMemberId: string): void => {
+  const items = getItemsAccountablePersonIsOn(selectedPolicyMemberId, policyId)
+  items.forEach((item) => {
```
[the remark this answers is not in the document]
hobbitronics: since updateItem doesn't return anything I was just going to rely on our normal error handling. updateItem will throw an error if and itemId is not set.

forevermatt (approved):

src/data/items.ts
```diff
@@ -335,14 +335,32 @@ export const itemIsInactive = (item: PolicyItem): boolean => {
...
+
+export const parseItemForAddItem = (item: PolicyItem): NewItemFormData => {
   return {
     accountablePersonId: item.accountable_person.id,
     categoryId: item.category.id,
     country: item.country || item.accountable_person.country,
     marketValueUSD: item.coverage_amount / 100,
     coverageStartDate: new Date().toISOString().slice(0, 10),
-    description: item.description,
+    itemDescription: item.description,
```
hobbitronics, line 363: ts caught another 2 bugs here. When I merged develop I added a return type to this and the function above.

PR #507, opened 2022-05-12
preserve new item data with qs

hobbitronics:
- added a qs for ItemId in ItemsNew
- fixed a bug (on my machine anyways) where Checkout wasn't importing properly
- fixed a bug where policyId was undefined

To summarize now can can go to ItemsNew, fill it out (minimum of name and category for it to create an item), navigate to Checkout or anywhere else, click 🔙   and arrive back at `ItemsNew` with the data still intact.

### DIFF
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -7,7 +7,7 @@ import ItemDeleteModal from '../ItemDeleteModal.svelte'
 import SelectAccountablePerson from '../SelectAccountablePerson.svelte'
 import { MAX_TEXT_AREA_LENGTH } from 'components/const'
 import type { AccountablePersonOptions } from 'data/accountablePersons'
-import { ItemCoverageStatus, PolicyItem } from 'data/items'
+import { ItemCoverageStatus, NewItemFormData, PolicyItem, SaveFormData } from 'data/items'
 import { categories, loadCategories, initialized as catItemsInitialized } from 'data/itemCategories'
 import TextFieldWithLabel from '../TextFieldWithLabel.svelte'
 import { assertHas, assertIsLessOrEqual } from '../../validation/assertions'
@@ -19,12 +19,12 @@ export let item = {} as PolicyItem
 export let policyId: string
 
 let applyBtnLabel = ''
-let formData = {} as any
+let formData = {} as NewItemFormData
 let open = false
 let makeModelIsOpen = false
 let selectedAccountablePersonId: string
 
-const dispatch = createEventDispatcher<{ submit: any; 'save-for-later': any; delete: any }>()
+const dispatch = createEventDispatcher<{ submit: NewItemFormData; 'save-for-later': SaveFormData; delete: any }>()
 
 // Set default values.
 let accountablePersonId = ''
@@ -70,7 +70,7 @@ const onSelectCategory = (event: any) => {
   categoryId = event.detail?.id
 }
 
-const getFormData = () => {
+const getFormData = (): NewItemFormData => {
   return {
     accountablePersonId,
     categoryId,
--- a/src/components/forms/ItemForm.svelte
+++ b/src/components/forms/ItemForm.svelte
@@ -7,7 +7,7 @@ import ItemDeleteModal from '../ItemDeleteModal.svelte'
 import SelectAccountablePerson from '../SelectAccountablePerson.svelte'
 import { MAX_TEXT_AREA_LENGTH } from 'components/const'
 import type { AccountablePersonOptions } from 'data/accountablePersons'
-import { ItemCoverageStatus, NewItemFormData, PolicyItem, SaveFormData } from 'data/items'
+import { ItemCoverageStatus, NewItemFormData, PolicyItem, UpdateItemFormData } from 'data/items'
 import { categories, loadCategories, initialized as catItemsInitialized } from 'data/itemCategories'
 import TextFieldWithLabel from '../TextFieldWithLabel.svelte'
 import { assertHas, assertIsLessOrEqual } from '../../validation/assertions'
@@ -24,7 +24,7 @@ let open = false
 let makeModelIsOpen = false
 let selectedAccountablePersonId: string
 
-const dispatch = createEventDispatcher<{ submit: NewItemFormData; 'save-for-later': SaveFormData; delete: any }>()
+const dispatch = createEventDispatcher<{ submit: NewItemFormData; 'save-for-later': UpdateItemFormData; delete: any }>()
 
 // Set default values.
 let accountablePersonId = ''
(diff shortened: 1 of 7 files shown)


src/pages/policies/[policyId]/items/new.svelte
```diff
@@ -18,35 +19,36 @@ let open = false
...
-  if (event.detail.isAutoSaving) {
-    //Todo once autosaving can happen on an empty form send the user immediately to edit.
-    // $goto(itemEdit(policyId, item.id))
-  } else {
+  if (!event.detail.isAutoSaving) {
     $goto(HOME)
   }
 }
 
+const saveOrAddItem = async (event: CustomEvent) =>
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items/new.svelte:49]: It might be worth changing this parameter to be the specific piece(s) of data needed, rather than just passing the full event. As-is, I don't know what `event.detail` holds, but switching to named parameters on this function would help clarify that.

src/helpers/routes.ts
```diff
@@ -22,6 +22,7 @@ export const FAQ = '/faq'
 export const ITEMS = '/items'
 export const items = (policyId: string) => `/policies/${policyId}/items`
 export const itemsNew = (policyId: string) => `/policies/${policyId}/items/new`
+export const itemsNewId = (policyId: string, query: string) => `/policies/${policyId}/items/new?${query}`
```
forevermatt, line 25: What would you think of making the `qs.stringify()` call here, rather than in all of places that call this? Just a thought.

forevermatt: It might also be worth naming this something more about the query string, rather than id, since this code doesn't even know whether an id was passed. Perhaps `itemsNewQs`? Your decision.

src/pages/policies/[policyId]/items/new.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/items/new.svelte]: I added some ts interfaces to help clarify what the data is.

src/helpers/routes.ts
hobbitronics: Good idea

src/pages/policies/[policyId]/items/new.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/items/new.svelte]: I added some ts interfaces to help clarify what the data is.

src/helpers/routes.ts
hobbitronics: Good idea

src/components/forms/ItemForm.svelte
```diff
@@ -19,12 +19,12 @@ export let item = {} as PolicyItem
...
 
 let applyBtnLabel = ''
-let formData = {} as any
+let formData = {} as NewItemFormData
 let open = false
 let makeModelIsOpen = false
 let selectedAccountablePersonId: string
 
-const dispatch = createEventDispatcher<{ submit: any; 'save-for-later': any; delete: any }>()
+const dispatch = createEventDispatcher<{ submit: NewItemFormData; 'save-for-later': UpdateItemFormData; delete: any }>()
```
forevermatt, line 27: Do we only use the `submit` event for new items? It seems like we would for editing an item, as well.

src/components/forms/ItemForm.svelte
```diff
@@ -70,7 +70,7 @@ const onSelectCategory = (event: any) => {
   categoryId = event.detail?.id
 }
 
-const getFormData = () => {
+const getFormData = (): NewItemFormData => {
```
forevermatt, line 73: Similarly here: Couldn't this also be data for updating an item, not just for adding a new item?

src/data/items.ts
```diff
@@ -217,7 +241,7 @@ export async function submitItem(itemId: string): Promise<void> {
  * @param {Object} itemData
  * @return {Object}
  */
-export async function updateItem(policyId: string, itemId: string, itemData: any): Promise<void> {
+export async function updateItem(policyId: string, itemId: string, itemData: ItemFormData): Promise<void> {
```
forevermatt, line 244: Should this `ItemFormData` be `UpdateItemFormData`? Based on the names, I expected `UpdateItemFormData` to be intended for use when updating an item (i.e. in this method call).

src/pages/policies/[policyId]/items/new.svelte
```diff
@@ -18,35 +26,40 @@ let open = false
...
-$: policyId && loadItems(policyId)
-
+$: item.id && $redirect(itemsNewQs(policyId, { itemId: item.id }))
 $: $selectedPolicy.type === PolicyType.Household && !$selectedPolicy.household_id && (open = true)
 
+$: $params.itemId && (item = $selectedPolicyItems.find((i) => i.id === $params.itemId) || {})
 $: breadcrumbLinks = [
   { name: 'Items', url: itemsRoute(policyId) },
-  { name: 'New', url: itemsNew(policyId) },
+  { name: 'New', url: item.id ? itemsNewQs(policyId, { itemId: item.id }) : itemsNew(policyId) },
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items/new.svelte:40]: If you wanted, you could simplify this down as follows (since Item ID is the only piece of data it can include), though you'd also need to update the route helper accordingly:
```suggestion
  { name: 'New', url: item.id ? itemsNewQs(policyId, item.id) : itemsNew(policyId) },
```
However, that's effectively undoing my earlier recommendation (and changing some other things), so feel free to use your judgment. It just seemed like we ended up an an unnecessarily-complicated syntax for calling this route helper function.

src/pages/policies/[policyId]/items/new.svelte
```diff
@@ -18,35 +26,40 @@ let open = false
...
-  } else {
+  if (!event.detail.isAutoSaving) {
     $goto(HOME)
   }
 }
 
+const saveOrAddItem = async (itemData: UpdateItemFormData | NewItemFormData) =>
+  item.id
+    ? await updateItem(policyId, item.id, itemData)
+    : (item = await addItem(policyId, itemData as NewItemFormData))
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items/new.svelte:61]: I had a little trouble reading this inline-function-with-ternary-operator. How do you feel about this alternate syntax, breaking it out with a normal `if` and `return`? I think the `else` didn't need to return anything (so I didn't do so here), but feel free to correct that if I misread it.

```suggestion
const saveOrAddItem = async (itemData: UpdateItemFormData | NewItemFormData) => {
  if (item.id) {
    return updateItem(policyId, item.id, itemData)
  } else {
    item = await addItem(policyId, itemData as NewItemFormData)
  }
}
```

src/components/forms/ItemForm.svelte
hobbitronics: yes, this is true, but it causes a ts error to  include UpdateItemFormData, because in itemForm we always have a covereageStartDate. Sometimes it seems you just can't win against type script.

src/data/items.ts
hobbitronics: yes, it could actually be either since the extra property on UpdateItemFormData isn't used by this function.

forevermatt (approved):

PR #509, opened 2022-05-15
add an info Card to category selector in ItemForm

hobbitronics:
image


forevermatt (approved):
Approved, with one question

PR #514, opened 2022-06-10
monthly policy report for team and household

hobbitronics:
latest:
image

old:
image

image

### DIFF
--- a/src/components/CustomerReport.svelte
+++ b/src/components/CustomerReport.svelte
@@ -3,6 +3,7 @@ import CreateCustomerReportModal from './CreateCustomerReportModal.svelte'
 import type { Policy } from 'data/policies'
 import { createPolicyLedgerReport, getLedgerReportById, LedgerReportType, policyLedgerReports } from 'data/ledger'
 import { formatDateAndTime, formatFriendlyDate } from 'helpers/dates'
+import { policyReportDetails } from 'helpers/routes'
 import FileLink from './FileLink.svelte'
 import { Button, Datatable, setNotice } from '@silintl/ui-components'
 
@@ -34,6 +35,7 @@ async function createReport(e: CustomEvent) {
       <Datatable.Header.Item>Report Date</Datatable.Header.Item>
       <Datatable.Header.Item>Created At</Datatable.Header.Item>
       <Datatable.Header.Item>File</Datatable.Header.Item>
+      <Datatable.Header.Item>Printable Receipt</Datatable.Header.Item>
     </Datatable.Header>
     <Datatable.Data>
       {#each $policyLedgerReports as report (report.id)}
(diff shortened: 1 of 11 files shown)


src/pages/policies/[policyId]/report.svelte
```diff
@@ -0,0 +1,175 @@
...
+      </thead>
+      <tbody>
+        {#each entries as entry}
+          <tr>
+            <td>{entry.item_name || ''}</td>
+            <td>{entry.status_before || '-'}</td>
+            <td>-></td>
+            <td>{entry.status_after}</td>
+            <td>{entry.type}</td>
+            <td>{formatMoney(entry.value)}</td>
```
Comment (jason-jackson) [inline src/pages/policies/[policyId]/report.svelte:156]: Should this be right-aligned so decimals line up (and same with total below)? Also, I think `$-1.00` looks odd. `-$1.00` or `($1.00)` are the more standardized ways of showing negative numbers.

Baggerone:
Maybe hide last_changed if it is 1/1/1 or just hide it completely until we know what to do with it with the back end?
Or leave is a hyphen rather than a bad date.

src/pages/policies/[policyId]/report.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/report.svelte]: Good catch, thanks

hobbitronics:
for some reason some of the <td>'s are line breaking (eg. ->) when printing now unless you print in landscape. I haven't found a solution yet without shortening the text.

src/components/CustomerReport.svelte
```diff
@@ -44,6 +46,16 @@ async function createReport(e: CustomEvent) {
           <Datatable.Data.Row.Item>
             <FileLink on:expired={() => getLedgerReportById(report.id, true)} file={report.file} />
           </Datatable.Data.Row.Item>
+          {#if report.type === LedgerReportType.monthly}
+            <Datatable.Data.Row.Item>
+              <a
+                target="_blank"
+                href={policyReportDetails(policy.id, report.date.split('-')[1], report.date.split('-')[0])}
```
forevermatt, line 53: I wonder if there's some simple way to clarify what's going on here, for readability.

It's easy enough to see **how** the code is getting the desired data (get the report's date, split that string on hyphens, take the 2nd piece, then do that all again and take the 1st piece), and we can figure out from that **what** the desired data is, but it feels like it takes more mental effort than is necessary.

What about something like this? (Just a suggestion; feel free to use, modify, or reject)
```
href={policyReportDetails(policy.id, getMonth(report.date), getYear(report.date))}
```
... and then add those little in-file helper methods above:
```
const getMonth = (iso8601Date: string) => iso8601Date.split('-')[1]
const getYear = (iso8601Date: string) => iso8601Date.split('-')[0]
```

src/helpers/money.ts
```diff
@@ -9,5 +9,7 @@ export const formatMoney = (cents: number): string => {
   if (!cents || !Number.isFinite(+cents)) {
     return ''
   }
-  return '$' + Number(cents / 100).toFixed(2)
+  const convertToDollars = (cents: number) => '$' + Number(cents / 100).toFixed(2)
+
+  return cents > 0 ? convertToDollars(cents) : `(${convertToDollars(cents * -1)})`
```
forevermatt, line 14: I'm not sure, but I think 0 itself wouldn't have the parentheses, so this `>` might need to be a `>=`.
```suggestion
  return cents >= 0 ? convertToDollars(cents) : `(${convertToDollars(cents * -1)})`
```

src/pages/policies/[policyId]/report.svelte
```diff
@@ -0,0 +1,180 @@
...
+
+  {#if $loading && isLoadingById(`policies/${policyId}/ledger-reports?month=${month}&year=${year}`)}
+    Loading report...
+  {:else if entries.length > 0}
+    <h3>Table</h3>
+    <table>
+      <thead>
+        <tr>
+          {#each columns as column}
+            <td
```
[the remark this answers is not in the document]
Comment (hobbitronics) [reply src/pages/policies/[policyId]/report.svelte:143]: yes, <th> is what we want. Thanks

hobbitronics:
Thanks for the feedback @forevermatt and @jason-jackson

PR #518, opened 2022-06-28
optimize batch item delete label

hobbitronics:
- show "delete" on button when only Drafts are selected
- otherwise continue showing "End Coverage"
image

### DIFF
--- a/src/components/Datatable/ItemsTable.svelte
+++ b/src/components/Datatable/ItemsTable.svelte
@@ -80,6 +80,8 @@ $: selectedItemNames = checkedItems.map((item) => item.name)
 $: sortedItemsArray = currentColumn.numeric
   ? sortByNum(currentColumn.path, items, ascending)
   : sortByString(currentColumn.path, items, ascending)
+$: allCheckedItemsAreDraft =
+  checkedItems.length > 0 && checkedItems.every((item) => item.coverage_status === ItemCoverageStatus.Draft)
 
 const dispatch = createEventDispatcher()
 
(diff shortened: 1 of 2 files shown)


src/components/Datatable/ItemsTable.svelte
```diff
@@ -80,6 +80,8 @@ $: selectedItemNames = checkedItems.map((item) => item.name)
 $: sortedItemsArray = currentColumn.numeric
   ? sortByNum(currentColumn.path, items, ascending)
   : sortByString(currentColumn.path, items, ascending)
+$: allCheckedItemsAreDraft =
+  checkedItems.length > 0 && checkedItems.every((item) => item.coverage_status === ItemCoverageStatus.Draft)
```
jason-jackson, line 84: Not sure if the length check is needed, it might be better to check if any items are not Draft instead. That way, if all items in table are in draft status, the button will show as "delete", otherwise it will be shown as "end coverage". Something like:

```suggestion
  items.some((item) => item.coverage_status !== ItemCoverageStatus.Draft) &&
  checkedItems.every((item) => item.coverage_status === ItemCoverageStatus.Draft)
```

jason-jackson (approved):

src/components/Datatable/ItemsTable.svelte
hobbitronics: The code works but still shows "End Coverage" when I only have a Draft Item. Funny that running [].every(callback) or [].some(callback) returns true, I tested it out in a sandbox. think I'll leave it as is since It will probably require yet another condition to have the benefit of showing "Delete" before selecting the lone Draft item.

PR #522, opened 2022-07-06
refactor for readability using style directive

hobbitronics:
- saw the style directive while looking up something else and thought it would be more readable

### DIFF
--- a/src/components/banners/StatusBanner.svelte
+++ b/src/components/banners/StatusBanner.svelte
@@ -6,7 +6,7 @@ export let state: any
 export let receiptType = '' as ReceiptType
 
 $: bgColor = state.bgColor || ''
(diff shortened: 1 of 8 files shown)


src/components/banners/StatusBanner.svelte
```diff
@@ -6,7 +6,7 @@ export let state: any
 export let receiptType = '' as ReceiptType
 
 $: bgColor = state.bgColor || ''
```
jason-jackson, line 8: Would it be worthwhile changing this to:

```$: background = `var(${state.bgColor})` || ''```

so you can just do `{background}` in Banner?

jason-jackson (approved):

hobbitronics:
Thanks for the feedback @jason-jackson!

PR #524, opened 2022-07-13
Disable Batch Delete when ended items selected and assert no open claims

hobbitronics:
- Disable Batch Delete when ended items selected
- Assert no open claims on items being deleted
- make item.status column red in ItemsTable to draw attention that coverage cancellation has already happened
  
image

### DIFF
--- a/src/components/Datatable/ItemsTable.svelte
+++ b/src/components/Datatable/ItemsTable.svelte
@@ -1,9 +1,11 @@
 <script lang="ts">
 import BatchItemClone from '../BatchItemClone.svelte'
 import BatchItemDelete from '../BatchItemDelete.svelte'
+import { ClaimItem, incompleteClaimItemStatuses, selectedPolicyClaims } from 'data/claims'
 import { getItemState } from 'data/states'
 import { AccountablePerson, editableCoverageStatuses, ItemCoverageStatus, PolicyItem } from 'data/items'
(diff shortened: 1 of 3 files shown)


src/components/Datatable/ItemsTable.svelte
```diff
@@ -235,7 +263,7 @@ const registerNewCheckbox = () => {
         <Datatable.Data.Row.Item>{item.name || ''}</Datatable.Data.Row.Item>
         <Datatable.Data.Row.Item class={getStatusClass(item.coverage_status)}>
           {#if item.coverage_status === ItemCoverageStatus.Approved && item.coverage_end_date}
-            Covered through {formatFriendlyDate(item.coverage_end_date)}
+            <div class="red">Covered through {formatFriendlyDate(item.coverage_end_date)}</div>
```
Schparky, line 266: I'm not sure this helps much. But I suppose it's not a problem either.

Schparky:
How difficult would it be to add a hover popup to explain why the Delete button is disabled?

Schparky (approved):
Thanks for the quick response.

hobbitronics:
> How difficult would it be to add a hover popup to explain why the Delete button is disabled?

Not too hard. I considered that but thought how would a user know to hover over it to get the information. I am favoring a little info icon like we've used elsewhere.

PR #528, opened 2022-07-25
added open graph meta tagas

hobbitronics:
- To help our shared links on facebook, etc added open graph meta

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -6,7 +6,7 @@
 	<meta property="og:title" content="Cover" />
   <meta property="og:type" content="website" />
   <meta property="og:image" content="https://cover.sil.org/logo.svg" />
-  <meta property="og:url" content="https://www.cover.sil.org" />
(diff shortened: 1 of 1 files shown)


dist/index.html
```diff
@@ -3,7 +3,10 @@
 <head>
 	<meta charset='utf-8'>
 	<meta name='viewport' content='width=device-width,initial-scale=1'>
-
+	<meta property="og:title" content="Cover" />
+  <meta property="og:type" content="website" />
+  <meta property="og:image" content="https://cover.sil.org/logo.svg" />
+  <meta property="og:url" content="https://www.cover.sil.org" />
```
jason-jackson, line 9: Is this correct? When I go here, I get the server cannot be found.

hobbitronics: I'm not sure where I copied an pasted that from, but you're correct. Its not.

jason-jackson (approved):

PR #531, opened 2022-07-26
update deps

hobbitronics:
- update non breaking deps
- remove FileDropArea and import from ui-components


forevermatt (approved):
Well, _that_ was an easy review 🙂

PR #534, opened 2022-10-25
Entity codes views

hobbitronics:
- add /admin/entities
- add /admin/{entityId}
- in /policies/new and /policies/{policyId}/settings concatenate for the menu
- use new endpoints for EntityCodes


src/pages/_layout.svelte
```diff
@@ -44,6 +44,12 @@ $: menuItems = [
     label: 'Reports',
     hide: !inAdminRole || userIsAnonymous,
   },
+  {
+    url: routes.ENTITIES,
+    icon: 'domain',
+    label: 'Entities',
+    hide: !inAdminRole || userIsAnonymous,
```
Schparky, line 51: This could probably use a helper with a name like `userNotAdmin()` since there are three of them.

Schparky (approved):
nice work!

src/pages/_layout.svelte
hobbitronics: I used a reactive statement.

PR #536, opened 2022-10-26
Release 1.1.0 -- manage entity codes

Schparky:
### Added
- New admin pages for managing entity codes
### Changed
- When displaying entities in the selection list on policy forms, the name is prefixed with the code to make it searchable by either. Previously, this prefixing was done in the backend code.


src/pages/policies/[policyId]/settings.svelte
```diff
@@ -61,7 +70,7 @@ $: if (policyId) {
 }
 
 $: $entityCodes.forEach((code) => {
-  entityOptions[code.name] = code.code
+  entityOptions[`${code.name} - ${code.code}`] = code.code
```
Comment (hobbitronics) [inline src/pages/policies/[policyId]/settings.svelte:73]: In the comments you said prefixed with the code. Does it matter that this is suffixed?

hobbitronics (approved):
Just one question about whether we meant to prefix with the code instead of suffix.

src/pages/policies/[policyId]/settings.svelte
Comment (Schparky) [reply src/pages/policies/[policyId]/settings.svelte:73]: Hmm, sorry about not making that clear. I think we should keep it as it was, just to not cause any confusion. Would you mind switching that around real quick?

Comment (hobbitronics) [inline src/pages/policies/[policyId]/settings.svelte:73]:
```suggestion
  entityOptions[`${code.code} - ${code.name}`] = code.code
```

src/pages/policies/new.svelte
```diff
@@ -21,7 +21,7 @@ onMount(() => $entityCodes.length || loadEntityCodes())
 
 $: metatags.title = formatPageTitle('New Team Policy')
 $: $entityCodes.forEach((code) => {
-  entityOptions[code.name] = code.code
+  entityOptions[`${code.name} - ${code.code}`] = code.code
```
hobbitronics, line 24:
```suggestion
  entityOptions[`${code.code} - ${code.name}`] = code.code
```

src/pages/policies/[policyId]/settings.svelte
Comment (Schparky) [reply src/pages/policies/[policyId]/settings.svelte:73]: Thanks

PR #538, opened 2022-11-14
Add copy button and serial number to ItemsTable 

hobbitronics:
- add CopyTableButton
- add serial number and button to show/hide it on ItemsTable
copytable
Added make and model
image
They are hidden by default
image

### DIFF
--- a/src/components/Datatable/ItemsTable.svelte
+++ b/src/components/Datatable/ItemsTable.svelte
@@ -14,7 +14,7 @@ import ItemDeleteModal from '../ItemDeleteModal.svelte'
 import type { Column } from './types'
 import { capitalize, random } from 'lodash-es'
 import { createEventDispatcher } from 'svelte'
-import { Button, Datatable, Menu, MenuItem } from '@silintl/ui-components'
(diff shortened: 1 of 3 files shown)


Schparky:
What would it look like to change the "show/hide serial #" to "show all columns" and "hide make/model/serial number columns". I don't really know what data people want, but it might help to look forward a bit.

hobbitronics:
> What would it look like to change the "show/hide serial #" to "show all columns" and "hide make/model/serial number columns". I don't really know what data people want, but it might help to look forward a bit.

That would be pretty easy. We can do all together or separate them out individually and maybe use checkboxes instead of the button.

src/components/Datatable/ItemsTable.svelte
```diff
@@ -65,6 +87,7 @@ const columns: Column[] = [
     sortable: true,
   },
 ]
+const itemsTableId = `items-table-${random(1000000, 9999999).toString()}`
```
Schparky, line 90: would `generateRandomID` work here?

Schparky (approved):

Schparky:
nice!

PR #540, opened 2022-11-16
Release 1.2.0

hobbitronics:
- Add SN, Make and Model to ItemsTable with a checkbox to toggle them
- Add a button to copy HTML from ItemsTable so it can be pasted into spreadsheets
- Bump minimatch version
- Bump loader-utils version


src/components/Datatable/CopyTableButton.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+      function handleCopy(event: any) {
+        event.clipboardData.setData('text/html', html)
+        event.clipboardData.setData('text/plain', html)
+
+        event.preventDefault()
+      }
+    } finally {
+      setNotice('Copied to clipboard')
+    }
```
forevermatt, line 41: A lot of this logic seems like it could be extracted to some kind of `clipboard.ts` helper file. This `CopyTableButton` needs to try to send things to the user's clipboard, but how that is done seems like implementation details that don't need to be here in this file (fwiw).

src/components/Datatable/CopyTableButton.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
+import { Button, setNotice } from '@silintl/ui-components'
+
+export let tableId: string
+
+async function copy() {
+  const tableContentsClone = document.getElementsByClassName(tableId)[0].cloneNode(true)
```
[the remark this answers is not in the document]
forevermatt: As another option, instead of passing in the table id/class, you could simply pass in the Node that you want to copy (i.e. use some kind of `bind:this={tableElement}` in the page or component using this button, then pass that in as a `tableElement` parameter to this CopyTableButton component).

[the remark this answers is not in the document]
hobbitronics: unfortunately that doesn't work as I can't clone or append `tableElement` to `hiddenElement`.

src/components/Datatable/CopyTableButton.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+  const elementByClassNameToRemove = [
+    '.home-table-more-vert',
+    '.mdc-data-table__sort-icon-button',
+    '.item-menu',
+    '.mdc-data-table__cell--checkbox',
+  ]
+  elementByClassNameToRemove.forEach((className) => {
+    const elements = hiddenElement?.querySelectorAll(className)
+    elements.forEach((el) => el.remove())
```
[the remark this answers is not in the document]
hobbitronics: I don't think this is necessary as `querySelectorAll` will return an empty NodeList if no selectors are found. I think it is even safe to get rid of ? in line 18.

src/components/Datatable/ItemsTable.svelte
```diff
@@ -206,6 +231,11 @@ const assertItemsHaveNoOpenClaims = (items: PolicyItem[]): void => {
     items.forEach((item) => checkClaimItemsForItemAndOpenClaim(claim.claim_items, item))
   })
 }
+
+const toggleShowSnMakeAndModel = () => {
+  ;[1, 2, 3].forEach((i) => (columns[i].hidden = !columns[i].hidden))
```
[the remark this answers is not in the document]
hobbitronics: I think making `const columnIndicesToToggle = [1, 2, 3]` up with columns is an improvement.

src/components/Datatable/CopyTableButton.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+  elementByClassNameToRemove.forEach((className) => {
+    const elements = hiddenElement.querySelectorAll(className)
+    elements.forEach((el) => el.remove())
+  })
+
+  const html = hiddenElement.innerHTML
+
+  if (html) {
+    await copyToClipboard(html)
```
forevermatt, line 27: Very nice extraction 👍

src/components/Datatable/ItemsTable.svelte
```diff
@@ -11,21 +12,44 @@ import { itemDetails, itemEdit } from 'helpers/routes'
...
 import { capitalize } from 'lodash-es'
+import { createEventDispatcher } from 'svelte'
+import { Checkbox, Datatable, Menu, MenuItem } from '@silintl/ui-components'
+import { generateRandomID } from '@silintl/ui-components/random'
 
 export let items = [] as PolicyItem[]
 export let policyId: string
 export let title: string = ''
 
+const columnIndicesToToggle = [1, 2, 3]
```
forevermatt, line 24: Good idea. I like how that groups these so close to the list of columns. 👍

src/components/Datatable/CopyTableButton.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+import { copyToClipboard } from '../../helpers/clipboard'
+import { Button, setNotice } from '@silintl/ui-components'
+
+export let tableIdClass: string
```
forevermatt, line 5: Unfortunately, I think renaming this from `tableId` (used as a class) to `tableIdClass` merely continues the same problem: An HTML `id` and an HTML `class` mean different things, and we even use different JS methods to look up an element by id vs by class.

I only see places where this is being used as a class, never as an id. Is that accurate? If so, would renaming it to something like `uniqueTableClass` accomplish the goal? It looks like this is only used to let one component (this `CopyTableButton`) reach out and find a DOM Element (the Datatable being copied) that is defined in a different component.

Sorry to be holding up a release, but this seems like an issue (while small) that will make the code harder for us to come back to later and properly understand.

I think renaming `tableIdClass` to `uniqueTableClass` will resolve this issue, and I'd be glad to approve it at that point. It clarifies that it is used as a CSS class, while also making it more obvious why we're using `generateRandomID()` to generate it.

hobbitronics: Yes, spot on.

forevermatt (approved):

PR #547, opened 2023-01-23
add EntityModal

hobbitronics:
image

### DIFF
--- a/src/components/TextFieldWithLabel.svelte
+++ b/src/components/TextFieldWithLabel.svelte
@@ -23,4 +23,4 @@ export let value: string = ''
     {/if}
   </span>
 {/if}
-<TextField {required} {description} {maxlength} bind:value />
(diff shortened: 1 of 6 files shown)


Schparky (commented):
Why does the model look so narrow in the screenshot?

hobbitronics:
> Why does the model look so narrow in the screenshot?

Since we are not using the default buttons the normal layout isn't being used. I could try using the buttons and getting rid of the form buttons or I could try hacking the the layout a bit.

Schparky:
> > Why does the model look so narrow in the screenshot?
> 
> ~~Since we are not using the default buttons the normal layout isn't being used. I could try using the buttons and getting rid of the form buttons or I could try hacking the the layout a bit. ~~ Actually it's not the buttons, but that we don't have anything wider than that in the modal. How's this though. image

Looks better. Just wanted to be sure there is room for the Name.

Schparky (approved):

PR #551, opened 2023-01-25
fix bad url bug

hobbitronics:
- when switching from Customer to Admin or opposite from Items or Claims views, would get a bad url

image

### DIFF
--- a/src/pages/_layout.svelte
+++ b/src/pages/_layout.svelte
@@ -121,22 +121,20 @@ const isCustomerOnOwnPolicy = (policyId: string) => policyId === $selectedPolicy
 const gotoPath = (policyId: string, claimOrItemIdObj = {}) => $goto($route.path, { policyId, ...claimOrItemIdObj })
 
 const goToCustomerView = (event: CustomEvent) => {
+  policyId = event.detail
   if (!urlIsClaimOrItem && $params.policyId) {
-    policyId = event.detail
     if ($route.path.includes('items')) {
       $goto(routes.items(policyId))
-      return
     } else if ($route.path.includes('claims')) {
       $goto(routes.customerClaims(policyId))
-      return
     } else {
       gotoPath(policyId)
     }
-  } else if (urlIsClaimOrItem && isCustomerOnOwnPolicy(event.detail)) {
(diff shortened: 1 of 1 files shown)


src/pages/_layout.svelte
```diff
@@ -122,7 +122,16 @@ const gotoPath = (policyId: string, claimOrItemIdObj = {}) => $goto($route.path,
 
 const goToCustomerView = (event: CustomEvent) => {
   if (!urlIsClaimOrItem && $params.policyId) {
-    gotoPath(event.detail)
+    policyId = event.detail
+    if ($route.path.includes('items')) {
+      $goto(routes.items(policyId))
+      return
```
Schparky, line 128: The `return`s are not needed since you have an else for every if

src/pages/_layout.svelte
```diff
@@ -132,11 +141,18 @@ const goToCustomerView = (event: CustomEvent) => {
...
+      parameters.claimId = $params.claimId
+    } else if ($params.itemId) {
+      parameters.itemId = $params.itemId
+    } else if ($route.path.includes('items')) {
+      $goto(routes.items($params.policyId))
+      return
+    } else if ($route.path.includes('claims')) {
+      $goto(routes.customerClaims($params.policyId))
+      return
+    }
```
Schparky, line 155: If you add an `else` you can get rid of two `return`s

Schparky (approved):
It's pretty complex. If you can think of ways to reduce complexity, it would help code quality. Unfortunately, I'm not coming up with anything to do about it right now.

src/pages/_layout.svelte
hobbitronics: I think for `goToAdminView` I don't need any of this logic because the admin can view any url it can just remain on that view.

hobbitronics:
> It's pretty complex. If you can think of ways to reduce complexity, it would help code quality. Unfortunately, I'm not coming up with anything to do about it right now.

The problem is that `$goto` isn't handling the routes with paths after `$params` properly. It is leaving out a `/` otherwise it would have worked as it was. There is an update to routify 3 now, but not sure that would fix it.

Schparky (approved):
Much better!

PR #553, opened 2023-02-15
hide the RoleAndPolicyMenu when anonymous

hobbitronics:
- hide the RoleAndPolicyMenu

### DIFF
--- a/src/components/AppDrawer.svelte
+++ b/src/components/AppDrawer.svelte
@@ -14,6 +14,7 @@ import { onMount } from 'svelte'
 export let menuItems: any[]
 export let myPolicies: Policy[]
 export let role: UserAppRole
+export let userIsAnonymous: boolean
 
 const isNotProduction = process.env.CF_PAGES_BRANCH !== 'main'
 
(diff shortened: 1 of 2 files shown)


mtompset (commented):
Looks good. Though, I can't say I ran it.

src/components/AppDrawer.svelte
```diff
@@ -57,14 +58,16 @@ const logoClickHandler = () => $goto(ROOT)
   title="Covered"
   class="drawer border-white {$showApp ? 'opacity1' : 'opacity0'}"
 >
-  <span class="pointer" on:click={logoClickHandler} slot="header">
+  <span class="pointer" on:click={logoClickHandler} on:keyup slot="header">
```
Schparky, line 61: This change seems unrelated. I also see nothing in the layout that handles the keyup event. Can you help me understand the purpose of this change?

Schparky (commented):
I have a question for my own understanding that I'd like answered before I approve the PR.

src/components/AppDrawer.svelte
hobbitronics: non button's that have on:click but no key event cause a warning, but since this doesn't get focus when tabbing it doesn't get used. Perhaps we can use an ignore comment or maybe dig into whether this should be focused by tabbing.

Schparky: Either one (ignore or actual fix) would be fine with me. Or even a TODO comment might be fine.

Schparky (approved):

forevermatt (approved):

src/components/AppDrawer.svelte
```diff
@@ -57,14 +58,17 @@ const logoClickHandler = () => $goto(ROOT)
   title="Covered"
   class="drawer border-white {$showApp ? 'opacity1' : 'opacity0'}"
 >
+  <!-- svelte-ignore a11y-click-events-have-key-events -->
   <span class="pointer" on:click={logoClickHandler} slot="header">
     <img class="logo" src="/logo.svg" alt="Cover" />
   </span>
```
forevermatt, line 64: I'm okay with proceeding as-is now, but it does bring up a good question of why we're listening for click events on the `span` rather than wrapping the logo image in an `a` link tag. Might be worth revising later.

hobbitronics: I'll open another PR for that issue. Thanks @forevermatt

PR #556, opened 2023-02-21
add save, save changes and discard button to personal settings

hobbitronics:
- Added buttons to personal settings
- custom_email gets saved when either "save" or "save changes" is clicked rather than on:blur
see https://trello.com/c/znDg06sW/245-add-save-changes-button-to-the-bottom-of-settings-pages-tell-user-about-autosave-onclick

image

### DIFF
--- a/src/pages/settings/personal.svelte
+++ b/src/pages/settings/personal.svelte
@@ -27,10 +27,28 @@ metatags.title = formatPageTitle('Personal Settings')
 
 $: country = $user.country || ''
 $: $user.id && setEmail()
+$: customEmailHasChanged = email_override && email_override !== $user.email_override
 $: notificationOptions = [
   { label: 'Default email: ' + $user.email, value: NOTIFICATION_OPTION_DEFAULT },
   { label: 'Custom email', value: NOTIFICATION_OPTION_CUSTOM },
 ]
+
+function onSave() {
+  if (customEmailHasChanged) {
+    updateCustomEmail()
+  } else {
+    setNotice('Changes are saved after each change')
+  }
+}
+
+function onDiscard() {
+  if (customEmailHasChanged) {
+    email_override = $user.email_override
+    setNotice('Your changes have been discarded')
+  } else {
+    setNotice('No changes to discard')
+  }
+}
 
 const setEmail = () => {
   if ($user.email_override !== $user.email) {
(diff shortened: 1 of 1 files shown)


Schparky:
This doesn't match my understanding of the Trello card. If I'm reading it correctly, we should keep the on:blur but add a notification telling the user that their information was saved. Then add the two buttons at the bottom for those that feel the need to click a button. Does that work?

hobbitronics:
> This doesn't match my understanding of the Trello card. If I'm reading it correctly, we should keep the on:blur but add a notification telling the user that their information was saved. Then add the two buttons at the bottom for those that feel the need to click a button. Does that work?

We could keep the on:blur for saving the custom email, but then there is nothing that the "Discard Changes" button does. I think it would only confuse users further. I also thought the on:blur was a little awkward as the user might possibly click/tab away from the input before they are done typing.

Schparky:
> > This doesn't match my understanding of the Trello card. If I'm reading it correctly, we should keep the on:blur but add a notification telling the user that their information was saved. Then add the two buttons at the bottom for those that feel the need to click a button. Does that work?
> 
> We could keep the on:blur for saving the custom email, but then there is nothing that the "Discard Changes" button does. I think it would only confuse users further. I also thought the on:blur was a little awkward as the user might possibly click/tab away from the input before they are done typing.

My personal opinion is that autosave is a nicer UX. Not sure how most people think about it, though. I think the important thing is to be consistent. Unfortunately, I can't remember if we use on:blur in other places in the Cover UI.

hobbitronics:
> > > This doesn't match my understanding of the Trello card. If I'm reading it correctly, we should keep the on:blur but add a notification telling the user that their information was saved. Then add the two buttons at the bottom for those that feel the need to click a button. Does that work?
> > 
> > 
> > We could keep the on:blur for saving the custom email, but then there is nothing that the "Discard Changes" button does. I think it would only confuse users further. I also thought the on:blur was a little awkward as the user might possibly click/tab away from the input before they are done typing.
> 
> My personal opinion is that autosave is a nicer UX. Not sure how most people think about it, though. I think the important thing is to be consistent. Unfortunately, I can't remember if we use on:blur in other places in the Cover UI.

We do in the Policy settings but the larger forms such as ItemForm do not since they need to create the item with a minimum set of information. I am happy to keep this on:blur but don't see the "Discard" button as misleading.

Schparky:
@hobbitronics I'll go ahead and approve it since either way is fine with me. @merchako if you have an opinion on this, please weigh in before this goes to production.

Schparky (approved):

PR #559, opened 2023-02-22
fix info modal for claim form

hobbitronics:
## fix
- the modal needed the internal button in some but not all location

### DIFF
--- a/src/components/InfoModal.svelte
+++ b/src/components/InfoModal.svelte
@@ -3,7 +3,7 @@ import { Dialog, IconButton } from '@silintl/ui-components'
 
 export let content = ''
 export let title = ''
-export let hasButton = true
+export let hasButton = false
 export let open = false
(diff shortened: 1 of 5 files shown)


src/components/InfoModal.svelte
```diff
@@ -3,14 +3,15 @@ import { Dialog, IconButton } from '@silintl/ui-components'
 
 export let content = ''
 export let title = ''
-
-let open = false
+export let hasButton = false
```
Schparky, line 6: If I understand this correctly, this controls whether the component should have a button to open itself. The name `hasButton` doesn't seem to help since instances have buttons (ok, cancel). Would a name like `hasInfoButton` or  `showOpenButton` work instead of `hasButton`?

Schparky (approved):
Just a prop naming question.

PR #564, opened 2023-02-23
add renew date

hobbitronics:
- show the next renew date if there is no end date on an item


src/components/ItemDetails.svelte
```diff
@@ -30,6 +30,9 @@ $: status = (item.coverage_status || '') as ItemCoverageStatus
 $: showRevisionMessage = item.status_reason && status === ItemCoverageStatus.Revision
 $: startDate = formatDate(item.coverage_start_date)
 $: endDate = formatDate(item.coverage_end_date)
+$: year = getYear(startDate)
```
Schparky, line 33: Maybe `thisYear` (instead of `year`) would add a bit of clarity

Schparky (approved):

PR #569, opened 2023-02-24
Sort recent changes

hobbitronics:
- refactored so Claims are guaranteed at the beginning of the array in recent changes on admin/home
- For Signators Claims willl only be of status `Review3` so they will be first. Stewards don't see these
see https://trello.com/c/4dKEP19z/399-for-stewards-reorder-action-cards-on-home

image

### DIFF
--- a/src/components/RadioOptions.svelte
+++ b/src/components/RadioOptions.svelte
@@ -22,7 +22,7 @@ const optionId = (option: RadioOption) => `radio-options-${String(option.value).
   display: block;
 }
 input {
-  margin-right: 0.3rem;
+  margin-right: 0.6rem;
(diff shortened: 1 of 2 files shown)


src/components/RadioOptions.svelte
```diff
@@ -22,7 +22,7 @@ const optionId = (option: RadioOption) => `radio-options-${String(option.value).
   display: block;
 }
 input {
-  margin-right: 0.3rem;
+  margin-right: 0.6rem;
```
hobbitronics, line 25:
```suggestion
  margin-right: 0.3rem;
```

Schparky (approved):

PR #572, opened 2023-03-02
audit renewals and repairs

hobbitronics:
- add /audits view
- add buttons to run audits and repairs
- add tables to show items found in audits and repairs
After the audit result has loaded
image
After running the repair
image
Item had an active claim
image
To help admin find claims if action is needed
image

### DIFF
--- a/src/data/audits.ts
+++ b/src/data/audits.ts
@@ -24,5 +24,4 @@ export const repairAudits = async (date: string): Promise<void> => {
   const response = (await CREATE('repairs', { repair_type: 'renewal', date })) as RepairsResult
 
   repairedAudits.set(response)
-  audits.set({} as AuditResult)
 }
--- a/src/data/audits.ts
+++ b/src/data/audits.ts
@@ -8,8 +8,16 @@ export type AuditResult = {
 }
 
 export const audits = writable({} as AuditResult)
+export const repairedAudits = writable({} as AuditResult)
 
 export const runAudits = async (date: string): Promise<void> => {
   const response = (await CREATE('audits', { audit_type: 'renewal', Date: date })) as AuditResult
   audits.set(response)
+}
+
+export const repairAudits = async (date: string): Promise<void> => {
+  const response = (await CREATE('audits', { audit_type: 'repair', Date: date })) as AuditResult
+
+  repairedAudits.set(response)
+  audits.set({} as AuditResult)
 }
--- a/src/data/audits.ts
+++ b/src/data/audits.ts
@@ -11,12 +11,12 @@ export const audits = writable({} as AuditResult)
 export const repairedAudits = writable({} as AuditResult)
 
 export const runAudits = async (date: string): Promise<void> => {
-  const response = (await CREATE('audits', { audit_type: 'renewal', Date: date })) as AuditResult
+  const response = (await CREATE('audits', { audit_type: 'renewal', date })) as AuditResult
   audits.set(response)
 }
 
 export const repairAudits = async (date: string): Promise<void> => {
-  const response = (await CREATE('audits', { audit_type: 'repair', Date: date })) as AuditResult
+  const response = (await CREATE('repairs', { repair_type: 'renewal', date })) as AuditResult
 
   repairedAudits.set(response)
   audits.set({} as AuditResult)
--- a/src/data/audits.ts
+++ b/src/data/audits.ts
@@ -1,12 +1,15 @@
 import type { PolicyItem } from './items'
 import { CREATE } from 'data'
+import { writable } from 'svelte/store'
 
 export type AuditResult = {
   AuditType: string
   Items: PolicyItem[]
 }
 
-export const runAudits = async (date: string): Promise<AuditResult> => {
+export const audits = writable({} as AuditResult)
+
+export const runAudits = async (date: string): Promise<void> => {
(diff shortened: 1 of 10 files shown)


src/data/audits.ts
```diff
@@ -0,0 +1,28 @@
...
+export const runAudits = async (date: string): Promise<void> => {
+  const response = (await CREATE('audits', { audit_type: 'renewal', date })) as AuditResult
+  audits.set(response)
+}
+
+export const repairAudits = async (date: string): Promise<void> => {
+  const response = (await CREATE('repairs', { repair_type: 'renewal', date })) as RepairsResult
+
+  repairedAudits.set(response)
+  audits.set({} as AuditResult)
```
hobbitronics, line 27: I'm planning on removing the line above

src/pages/admin/audits.svelte
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
+import AuditsOrRepairTable from './_components/AuditsOrRepairTable.svelte'
+import { audits, repairAudits, repairedAudits, runAudits } from 'data/audits'
+import { formatPageTitle } from 'helpers/pageTitle'
+import { isEqual } from 'lodash-es'
```
hobbitronics, line 5: delete unused import

src/pages/admin/audits.svelte
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const repair = async () => {
+  try {
+    await repairAudits(utcDate)
+    setNotice('Succesfully repaired item records found to be at fault.')
+  } catch {
+    setNotice('There was an error repairing the item records. Please try again.')
+  }
+}
```
hobbitronics, line 30: Maybe I should run `runAudits` onMount but also leave the button there so if the admin needs to delete a claim before running it they can?

src/pages/admin/audits.svelte
```diff
@@ -0,0 +1,70 @@
...
+
+onMount(() => {
+  haveAuditResults || runAudits(utcDate)
+})
+</script>
+
+<Page>
+  <h4>Audits results (items that were incorrectly renewed and billed)</h4>
```
Schparky, line 55:
```suggestion
  <h3>Audit results</h3>
  <p>items that were incorrectly renewed and billed</p>
```

src/pages/admin/_components/AuditsOrRepairTable.svelte
```diff
@@ -0,0 +1,58 @@
...
+      <Datatable.Header.Item>Annual Premium</Datatable.Header.Item>
+      <Datatable.Header.Item>End Date</Datatable.Header.Item>
+      <Datatable.Header.Item>Coverage Status</Datatable.Header.Item>
+      <Datatable.Header.Item>Action</Datatable.Header.Item>
+    </Datatable.Header>
+    <Datatable.Data>
+      {#each items as item (item.id)}
+        <Datatable.Data.Row on:click={() => gotoItemDetails(item)} clickable>
+          <Datatable.Data.Row.Item>{item.name}</Datatable.Data.Row.Item>
+          <Datatable.Data.Row.Item>{item.policy_id}</Datatable.Data.Row.Item>
```
Schparky, line 41: I think it would be better to leave the policy ID column out. Since it means nothing to users, it provides no value.

src/pages/admin/_components/AuditsOrRepairTable.svelte
```diff
@@ -0,0 +1,58 @@
...
+
+const preventRowClick = async () => {
+  canViewItemDetails = false
+}
+</script>
+
+{#if items.length}
+  <Datatable>
+    <Datatable.Header>
+      <Datatable.Header.Item>Name</Datatable.Header.Item>
```
Schparky, line 30: I noticed that the header row looks clickable because the cursor changes to a pointing finger, but it does nothing. If it's not too much trouble, either make it sortable or don't make it look clickable.

src/data/items.ts
```diff
@@ -373,3 +374,16 @@ export const parseItemForAddItem = (item: PolicyItem): NewItemFormData => {
     uniqueIdentifier: item.serial_number,
   }
 }
+
+export const isItemWithActiveClaim = async (itemId: string, policyId: string): Promise<boolean> => {
+  await loadClaimsByPolicyId(policyId)
+  let hasActiveClaim = false
+
+  get(claims).forEach((claim) => {
+    if (claimIsOpenButNotDraft(claim)) {
```
Schparky, line 383: Paid and Denied are also safe states. This is a case where the backend should include a `isSafeToUpdate` property on the item object.

hobbitronics: This function returns true for `items` with an active `claim` to display (as long as the `item` is found).

Shouldn't it return `false` for "Paid" and "Denied" while open `claims` (except "drafts" can be updated) should return `true`?

This is al academic as far as the ui is concerned now but just thought it might help with you writing the backend logic.

Schparky: As it's named `claimIsOpenButNotDraft`, it should not include Paid and Denied. That's a problem with the helper name, not the logic. If it was named something like `cannotBeEdited` then it wouldn't be a problem. The backend logic was done a long time ago. I'm just exposing it on the API.

Schparky:
This PR is getting messy with a lot of work in progress. Would it make sense to close it and start a new one when you're done?

PR #573, opened 2023-03-06
audit renewals

hobbitronics:
- add /audits view
- add buttons to run audits and repairs
- add tables to show items found in audits and repairs
After the audit result has loaded
image
After running the repair
image
Item had an active claim
image
To help admin find claims if action is needed
image

### DIFF
--- a/src/data/items.ts
+++ b/src/data/items.ts
@@ -378,5 +378,7 @@ export const parseItemForAddItem = (item: PolicyItem): NewItemFormData => {
 }
 
 export const getUneditableItems = (items: PolicyItem[]): PolicyItem[] => {
-  return items.filter((item) => !assertItemCanBeUpdated(item))
+  return items.filter((item) => {
+    return item.can_be_updated === false
+  })
 }
--- a/src/data/items.ts
+++ b/src/data/items.ts
@@ -2,6 +2,7 @@ import { CREATE, DELETE, GET, UPDATE } from '.'
 import { throwError } from '../error'
 import { derived, get, writable } from 'svelte/store'
 import { selectedPolicyId } from './role-policy-selection'
+import { assertItemCanBeUpdated } from '../validation/assertions'
 
 export enum ItemCoverageStatus {
   Draft = 'Draft',
@@ -374,4 +375,8 @@ export const parseItemForAddItem = (item: PolicyItem): NewItemFormData => {
     riskCategoryId: item.risk_category.id,
     uniqueIdentifier: item.serial_number,
   }
+}
+
+export const getUneditableItems = (items: PolicyItem[]): PolicyItem[] => {
+  return items.filter((item) => !assertItemCanBeUpdated(item))
 }
--- a/src/data/items.ts
+++ b/src/data/items.ts
@@ -1,5 +1,5 @@
 import { CREATE, DELETE, GET, UPDATE } from '.'
-import { claimIsOpen, claims, loadClaimsByPolicyId } from 'data/claims'
+import { claimIsOpenButNotDraft, claims, loadClaimsByPolicyId } from 'data/claims'
 import { throwError } from '../error'
 import { derived, get, writable } from 'svelte/store'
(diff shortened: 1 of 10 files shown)


src/pages/admin/audits.svelte
```diff
@@ -20,20 +20,21 @@ $: haveRepairResults = repairedItems?.length
 $: repairIsDisabled = !haveAuditResults || !!haveRepairResults || !!itemsWithOpenClaim.length
 
 const onClick = () => {
+  itemsWithOpenClaim = []
   runAudits(utcDate)
 }
 
-const checkForOpenClaims = async () => {
+const checkForOpenClaims = () => {
```
Schparky, line 27: The name of this function (`checkForOpenClaims`) and the variable `itemsWithOpenClaim` assume that "open" and "can be deleted" are logically inverse of each other. While that may be true, it's not good for the UI code to assume that. You could rename this function to `findLockedItems` but it seems better to use the `assertItemsCanBeDeleted` function in assertions.ts instead.

hobbitronics: I agree on the name. I thought about moving it to assertions, but if I did that it would be throwing an error and interrupting the function and not updating `itemsWithOpenClaim`. So I think it is inherently different than our assertions. If you have any ideas how to overcome that issue I'm open to suggestions.

Schparky: It would be a compromise, but you could make `assertItemsCanBeDeleted` return a bool and let the caller decide what to do with it. Wouldn't be an "assertion" then, I suppose. Maybe it would be worth adding a helper `allItemsCanBeDeleted` that both the audits page and the assert could use.

src/pages/admin/audits.svelte
```diff
@@ -34,7 +34,7 @@ const checkForOpenClaims = () => {
 }
 
 const repair = async () => {
-  checkForOpenClaims()
+  findLockedItems()
   if (itemsWithOpenClaim.length) {
```
Schparky, line 38: @hobbitronics `itemsWithOpenClaim` has the same naming issue

src/pages/admin/audits.svelte
hobbitronics: How about this ````export const getUneditableItems = (items: PolicyItem[]): PolicyItem[] => {
  return items.filter((item) => !assertItemCanBeUpdated(item))
}

Schparky: Looks good

src/validation/assertions.ts
```diff
@@ -45,3 +46,7 @@ export function assertEmailAddress(email: string, errorMessage: string): void {
     throwError(errorMessage)
   }
 }
+
+export const assertItemCanBeUpdated = (item: PolicyItem): boolean => {
+  return item.can_be_updated === true
```
Schparky, line 51: This doesn't fit the paradigm in this file, where "assert" means "throw an error if this isn't true". For now, just remove this function and put the flag check directly in `getUneditableItems`.

src/data/items.ts
```diff
@@ -373,3 +376,7 @@ export const parseItemForAddItem = (item: PolicyItem): NewItemFormData => {
     uniqueIdentifier: item.serial_number,
   }
 }
+
+export const getUneditableItems = (items: PolicyItem[]): PolicyItem[] => {
+  return items.filter((item) => !assertItemCanBeUpdated(item))
```
Schparky, line 381:
```suggestion
  return items.filter((item) => !item.can_be_updated
```

src/data/items.ts
```diff
@@ -2,6 +2,7 @@ import { CREATE, DELETE, GET, UPDATE } from '.'
 import { throwError } from '../error'
 import { derived, get, writable } from 'svelte/store'
 import { selectedPolicyId } from './role-policy-selection'
+import { assertItemCanBeUpdated } from '../validation/assertions'
```
Schparky, line 5:
```suggestion
```

Schparky (approved):

PR #574, opened 2023-03-06
ignore Jetbrains IDE files

Schparky:


hobbitronics (approved):
Looks good to me. I don't know why this change looks like it removed the last line though. I pulled it and its there.

PR #576, opened 2023-03-08
Add hashes to build files

hobbitronics:
- add hash to build files to prevent unwanted caching
- remove index.html from being tracked as it is now generated by a template


rollup.config.js
```diff
@@ -11,13 +11,15 @@ import routify from '@roxi/routify/plugins/rollup'
...
+import html from '@rollup/plugin-html'
 
 const production = !process.env.ROLLUP_WATCH
 
 export default {
   input: 'src/main.ts',
   output: {
-    file: 'dist/bundle.js',
+    dir: 'dist',
+    entryFileNames: 'bundle.[hash].js',
```
Schparky, line 22: Curious why the css file doesn't need an entry here.

Schparky (approved):
Good work

rollup.config.js
hobbitronics: looks like it preserves the extension for other files https://rollupjs.org/configuration-options/#output-entryfilenames

PR #580, opened 2023-03-13
hide inactive entities

Schparky:
### Fixed
- Switch from using `docker-compose up` to `docker-compose run` for the `make docker` command.
- Filter out inactive and household entity codes for new policy form.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,7 @@ start: install local-dev
 
 docker:
 	docker-compose run --rm app npm install
-	docker-compose up -d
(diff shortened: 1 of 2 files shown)


hobbitronics (approved):
I'm curious what is the benefit of using docker-compose run?

Schparky:
> I'm curious what is the benefit of using docker-compose run?

It makes `make docker` more like `make`. It shows the console without having to do a `docker-compose log -f` and it terminates the container when you Ctrl-C out of it.

PR #581, opened 2023-03-14
Release 1.7.1 - Filter entity codes on new policy form

Schparky:
### Fixed
- Switch from using `docker-compose up` to `docker-compose run` for the `make docker` command.
- Filter out inactive and household entity codes for new policy form.


hobbitronics (commented):
Looks good.

hobbitronics (approved):

PR #583, opened 2023-03-14
show the annual renewal status

Schparky:


src/pages/admin/renewals.svelte
```diff
@@ -18,8 +30,13 @@ metatags.title = formatPageTitle('Admin > Renewals')
 <Page>
   <h3>Annual renewals</h3>
   <p>annual process for renewing coverage</p>
-  <p></p>
+  <p>
+    Number of items to renew: {status?.items_to_process?.toLocaleString('en-US')}
```
hobbitronics, line 34: is their a reason to not let the browser decide the locale?

hobbitronics: if the property is undefined do you wan't to show something like "loading...".

hobbitronics: "loading..." isn't right here, but maybe "not loaded yet"

src/pages/admin/renewals.svelte
```diff
@@ -18,8 +30,13 @@ metatags.title = formatPageTitle('Admin > Renewals')
...
+    {#if isProcessing && !status.is_complete}
+      <Button class="ml-1" on:click={onClickRefresh}>refresh</Button>
+    {/if}
+  </p>
```
hobbitronics, line 38: I am guessing there is a delay on the backend from `processPolicyRenewals` finishing and `getPolicyRenewalStatus` responding with the updated data and that's why you've gone with the refresh button rather than awaiting `processPolicyRenewals`?

src/pages/admin/renewals.svelte
Schparky: > is their a reason to not let the browser decide the locale?

Only because I don't know how to do that, and this will very likely only be used by staff at the central office in the U.S. I'm open to making it more flexible, though.

hobbitronics (approved):
I've got a few questions, but nothing blocking. Looks good.

src/pages/admin/renewals.svelte
Schparky: > if the property is undefined do you wan't to show something like "loading...".

Maybe, but the response is typically pretty quick.

hobbitronics: Not a big deal either way, but I think you just leave that parameter out. "In implementations without Intl.DateTimeFormat support, this parameter is ignored and the host's locale is usually used." see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString)

src/pages/admin/renewals.svelte
Schparky: Yes, awaiting `processPolicyRenewals` won't help because it returns long before the background process actually finishes. The "correct" way to do this would be with websockets, but that's a bit overkill for this.

PR #585, opened 2023-03-16
add claims/new and allow ClaimForm to choose an item

hobbitronics:
## Added
- Add a slot to ClaimForm
- Add "New Claim" button (uncomment)
- Add claims/new

## Changed
- prevent ClaimForm from submitting without and itemId

image


src/pages/policies/[policyId]/claims/new.svelte
```diff
@@ -1,2 +1,96 @@
...
+    <p>
+      It looks like there is already a claim for that item. Please
+      <a href={routes.customerClaimDetails(policyId, existingClaim.id)}>click here</a> to see its details.
+    </p>
+  {/if}
+
+  <h1>New claim</h1>
+
+  <ClaimForm itemIsSelectable {item} on:save-for-later={onSaveForLater} on:submit={onSubmit}>
+    <Select
```
Comment (Schparky) [inline src/pages/policies/[policyId]/claims/new.svelte:88]: Seems like this `Select` should be inside the component and hidden when `itemIsSelectable` is false. Is that not possible? Just doesn't seem like a normal use of a slot. At least not a *default* slot.

Comment (hobbitronics) [reply src/pages/policies/[policyId]/claims/new.svelte:88]: I didn't want it to be a default part of the ClaimForm as it's very complicated already and that would add more props and functions as well. We could just have it outside of ClaimForm and remove the prop since the data is passing in as `Item` anyways.

Comment (hobbitronics) [reply src/pages/policies/[policyId]/claims/new.svelte:88]: would this be preferrable? ```  <div class="w-50">
    <Select
      label="Item"
      {selectedID}
      {options}
      on:change={(event) => (itemId = event.detail.id)}
      on:populated={onPopulated}
    />
  </div>

  <ClaimForm itemIsSelectable {item} on:save-for-later={onSaveForLater} on:submit={onSubmit} />```

Comment (Schparky) [reply src/pages/policies/[policyId]/claims/new.svelte:88]: Yes, I think so. 

If ClaimForm is very complicated, maybe it needs some refactoring to extract subcomponents. Not a task for this PR, but maybe worth considering after this is done, or perhaps even before finishing this PR. Doing this the "wrong" way because of technical debt only adds more debt.

Schparky (approved):

PR #588, opened 2023-03-17
Add item selector

hobbitronics:
- move ItemSelector into ClaimForm

### DIFF
--- a/src/components/forms/ClaimForm.svelte
+++ b/src/components/forms/ClaimForm.svelte
@@ -23,10 +23,11 @@ import type { ClaimFormData } from './claims/types'
 import { assertHas } from '../../validation/assertions'
 import { Button, Form, IconButton, MoneyInput, TextArea } from '@silintl/ui-components'
 import { createEventDispatcher, onMount } from 'svelte'
+import ItemSelector from './claims/_components/ItemSelector.svelte'
 
 export let claim = {} as Claim
 export let item = {} as PolicyItem
-export let hideItemName = false
+export let items = [] as PolicyItem[]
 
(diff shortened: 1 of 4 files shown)


src/components/forms/ClaimForm.svelte
```diff
@@ -23,10 +23,11 @@ import type { ClaimFormData } from './claims/types'
 import { assertHas } from '../../validation/assertions'
 import { Button, Form, IconButton, MoneyInput, TextArea } from '@silintl/ui-components'
 import { createEventDispatcher, onMount } from 'svelte'
+import ItemSelector from './claims/_components/ItemSelector.svelte'
```
Schparky, line 26: sort imports

src/components/forms/ClaimForm.svelte
```diff
@@ -167,8 +168,10 @@ function onInfoClick(event: Event) {
 </script>
 
 <div class="w-50">
-  {#if !hideItemName}
+  {#if items.length < 1}
```
Schparky, line 171: Seems this is more conventional, but my personal preference would be `if items.length > 0` and reverse the conditions. Go with your own preference, though.
```suggestion
  {#if !items?.length}
```

src/components/forms/ClaimForm.svelte
Schparky, line 1: I like that this change didn't affect other instances of the ClaimForm component.

Schparky (approved):
nice

src/components/forms/ClaimForm.svelte
hobbitronics: In the end I was surprised how little complexity it added.

PR #590, opened 2023-03-17
improve claim form validation

hobbitronics:
- instead of disabling buttons use validation and required on inputs

### DIFF
--- a/src/components/DateInput.svelte
+++ b/src/components/DateInput.svelte
@@ -4,6 +4,7 @@ export let disabled = false
 export let autofocus = false
 export let padding = '12px'
 export let width = '220px'
+export let required = false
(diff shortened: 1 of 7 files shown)


Schparky (commented):
I feel for you having to wade through this form for 2+ days. It's very complex!

The change to making inputs all required is a good change, I think. Couldn't the validation functions be simpler as a result? I tried commenting those lines and get strange results. The browser usually blocks submission, but not if the missing piece is the FMV money input. Why would that one be any different?

But as it is, I got an incorrect error message at one point: I did not choose Repairable/Not Repairable but got the message "Please enter a fair market value".

Schparky (approved):

PR #592, opened 2023-03-22
fix ItemForm allowing 0, move functions

hobbitronics:
- prevent allowing `0` for `marketValue`
- move functions to a helper file
- fix a warning


Schparky (approved):
looks good

PR #594, opened 2023-03-23
minor form improvements

hobbitronics:
- Add "repairable" header to ClaimForm
- fix MoneyInput defaulting to 0
- move assertion to helper function


Schparky (approved):
good changes

PR #596, opened 2023-05-09
add hsts header

hobbitronics:
image

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -34,6 +34,7 @@ async function customFetch<T>(method: FetchMethod, uri: string, body: any = unde
   const headers: HeadersInit = {
     Authorization: `Bearer ${getToken()}`,
     'Content-Type': 'application/json',
+    'Strict-Transport-Security': 'max-age=31536000',
   }
 
   // when dealing with FormData, i.e., when uploading files, allow the browser to set the request up


forevermatt (approved):

forevermatt (changes requested):
I think this actually has to come from the server... sending it as part of the request headers is, I think, ineffective.

Baggerone:
> I think this actually has to come from the server... sending it as part of the request headers is, I think, ineffective.

@forevermatt What do you mean by `the server`, the `cover-api`?

forevermatt:
> > I think this actually has to come from the server... sending it as part of the request headers is, I think, ineffective.
> 
> @forevermatt What do you mean by `the server`, the `cover-api`?

Whichever server is responding to the HTTP / HTTPS requests that we are trying to ensure are requested using HTTPS. In this case, I think the goal is to use this HSTS header to get the UI to be served more reliably over HTTPS, but the header was added in JavaScript calls from the UI to the API. It needs to be added to the webserver response that is serving up the UI files (which might be Cloudflare pages, in this case).

hobbitronics:
> > > I think this actually has to come from the server... sending it as part of the request headers is, I think, ineffective.
> > 
> > 
> > @forevermatt What do you mean by `the server`, the `cover-api`?
> 
> Whichever server is responding to the HTTP / HTTPS requests that we are trying to ensure are requested using HTTPS. In this case, I think the goal is to use this HSTS header to get the UI to be served more reliably over HTTPS, but the header was added in JavaScript calls from the UI to the API. It needs to be added to the webserver response that is serving up the UI files (which might be Cloudflare pages, in this case).

I think you are correct about it being Cloudflare. There were instructions on how to do this in [HSTS for Cover](https://itse.youtrack.cloud/issues?q=project%3A%20%7BETH%7D&preview=CVR-9)

PR #601, opened 2023-07-13
set engines, run npm update

hobbitronics:
Most related to https://itse.youtrack.cloud/issue/ITSE-492/Update-Node-version-in-all-our-Node-applications-from-16-to-18-by-September


forevermatt (approved):

package.json
```diff
@@ -1,5 +1,9 @@
 {
   "name": "@silintl/cover-ui",
+  "engines": {
+    "node": ">=18 <19",
```
briskt, line 4: Should the version in `docker-compose.yml` match this?

hobbitronics: Yes, thanks for catching that.

briskt (approved):

PR #604, opened 2023-07-28
Feature/report csv+zip

merchako:
[CVR-613](https://itse.youtrack.cloud/issue/CVR-613) On the admin report page, add a new file download line. Describe the difference between the different downloads:

- The original download is for Sage and will give a CSV of all transactions (as before)
- The second download is for NetSuite and will give a ZIP of transactions split into different CSVs by type. I added a field to ReportLedger for this file, which will need to be populated before release (TODO @jason-jackson).

I also improved the semantics and formatting of the page.

There was already a bug where the file link will sometimes stop being a link. Even though it loads as a link, when I go away and come back, it's no longer a link—not clickable, blue, or underlined! I don't know whether it had been reported or now. I haven't been able to reliably replicate the behavior.

Devon wrote
> Based on your screenshot, where the .csv file is located, I'd like to move forward with just listing the csv and the zip file in that column.  Whether it is turned into a dropdown button for each report created or just put a <br> after the csv is up to you.  If the entire row is currently linked to download the CSV, then we'll need to adjust that so just the csv and zip files are linked.  This is a short-term temporary change that only affects Admins, so I'd prefer it being completed as simply and quickly as possible.

screenshot_2023-07-27_at_11 06 54_pm
Screenshot 2023-07-28 at 10 07 43 AM

### DIFF
--- a/src/data/ledger.ts
+++ b/src/data/ledger.ts
@@ -5,8 +5,7 @@ import { writable } from 'svelte/store'
 
 export type LedgerReport = {
   id: string
-  file: CoverFile // Mixed all transactions (Sage)(original)
-  zip: CoverFile // Split transactions into files by type (NetSuite)
+  file: CoverFile
   type: string
   date: string
   is_cleared: boolean
--- a/src/data/ledger.ts
+++ b/src/data/ledger.ts
@@ -5,7 +5,8 @@ import { writable } from 'svelte/store'
 
 export type LedgerReport = {
   id: string
-  file: CoverFile
+  file: CoverFile // Mixed all transactions (Sage)(original)
+  zip: CoverFile // Split transactions into files by type (NetSuite)
   type: string
   date: string
   is_cleared: boolean
--- a/src/pages/admin/reports/[reportId].svelte
+++ b/src/pages/admin/reports/[reportId].svelte
@@ -73,9 +73,7 @@ dl {
     <dl>
       <dt>Report Date</dt>
       <dd>{formatFriendlyDate(report.date)}</dd>
-      <dt>
-        File
-      </dt>
+      <dt>File</dt>
       <dd><FileLink on:expired={getReport} file={report.file} /></dd>
       <dt>Cleared</dt>
       <dd>{report.is_cleared ? 'Yes' : 'No'}</dd>
--- a/src/pages/admin/reports/[reportId].svelte
+++ b/src/pages/admin/reports/[reportId].svelte
@@ -76,6 +76,10 @@ th {
         <td><FileLink on:expired={getReport} file={report.file} /></td>
       </tr>
       <tr>
+        <th>File</th>
+        <td><FileLink on:expired={getReport} file={report.file} /></td>
+      </tr>
+      <tr>
         <th>Cleared</th>
         <td>{report.is_cleared ? 'Yes' : 'No'}</td>
       </tr>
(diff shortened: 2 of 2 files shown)


briskt:
> There was already a bug where the file link will sometimes stop being a link. Even though it loads as a link, when I go away and come back, it's no longer a link—not clickable, blue, or underlined! I don't know whether it had been reported or now. I haven't been able to reliably replicate the behavior.

It's not a bug, though admittedly a UX issue. It change to unlinked text when the URL expires. We use self-signed S3 URLs, which have an expiration time. That time is pretty short, so that's why you're seeing the link disappear.

hobbitronics (approved):
Looks good to me.

merchako:
I'm expecting to hear from @jason-jackson before merging this branch since I'm not in a position to populate the new link.

Unless y'all'd prefer I merge to _origin/develop_ and Jason add in the URL before merging to _origin/main_?

src/data/ledger.ts
```diff
@@ -5,7 +5,8 @@ import { writable } from 'svelte/store'
 
 export type LedgerReport = {
   id: string
-  file: CoverFile
+  file: CoverFile // Mixed all transactions (Sage)(original)
+  zip: CoverFile // Split transactions into files by type (NetSuite)
```
jason-jackson, line 9: Would it be cleaner to have all the files listed in a single column? That way, when the Sage report goes away, we don't need to remove the column. I'd need to slightly alter the backend some to allow reports to have multiple files rather than adding another database field, but I see it as preferable to this. Thoughts?

So instead of being `file: CoverFile` it would be `file: CoverFile[]` (or however syntax is), and then `File` column could be renamed to `Files` and Report page would list the files. Directly downloading report probably would need to be changed though.

src/data/ledger.ts
```diff
@@ -5,7 +5,8 @@ import { writable } from 'svelte/store'
...
   type: string
```
jason-jackson, line 10:
```suggestion
  file: CoverFile
  type: string
```

src/pages/admin/reports/[reportId].svelte
```diff
@@ -37,52 +37,56 @@ async function getReport() {
...
+        File for Sage
+        <div class="fs-12">Mixed transations</div>
+      </dt>
+      <dd><FileLink on:expired={getReport} file={report.file} /></dd>
+      <dt>
+        File for NetSuite
+        <div class="fs-12">Split transations</div>
+      </dt>
+      <dd><FileLink on:expired={getReport} file={report.zip} /></dd>
+      <dt>Cleared</dt>
```
Comment (jason-jackson) [inline src/pages/admin/reports/[reportId].svelte:86]:
```suggestion
      <dt>
        File
      </dt>
      <dd><FileLink on:expired={getReport} file={report.file} /></dd>
      <dt>Cleared</dt>
```

jason-jackson (commented):
I think these are all the changes needed based on our discussion yesterday. If you want to keep the Split/Mixed transactions, you could, but you'll need to check if filename contains `sage`.

src/pages/admin/reports/[reportId].svelte
```diff
@@ -37,52 +37,50 @@ async function getReport() {
...
-        <th>Transactions</th>
-        <td>{report.transaction_count}</td>
-      </tr>
-    </table>
+    <dl>
+      <dt>Report Date</dt>
+      <dd>{formatFriendlyDate(report.date)}</dd>
+      <dt>
+        File
+      </dt>
```
Comment (jason-jackson) [inline src/pages/admin/reports/[reportId].svelte:78]:
```suggestion
      <dt>File</dt>
```

PR #606, opened 2023-08-31
improve Inactive Items display in table CVR-654 & CVR-653

hobbitronics:
[CVR-654](https://itse.youtrack.cloud/issue/CVR-654/ItemTable-Improve-display-of-inactive-items)
[CVR-653](https://itse.youtrack.cloud/issue/CVR-653/ItemsTable.svelte-Low-contrast-text-for-no-coverage-items)

## Changed
- Inactive.title: 'Not covered'
- refactor sorting and helpers to remove complexity from Tables
- improve some types
- show 'Coverage ended' if ended already
- make table row low contrast if "not covered" `status` of "Inactive" and  "Denied"


src/components/Datatable/ItemsTable.svelte
```diff
@@ -286,30 +285,36 @@ and Model
...
+          <RowItem {item}>{item.serial_number || ''}</RowItem>
+          <RowItem {item}>{item.make || ''}</RowItem>
+          <RowItem {item}>{item.model || ''}</RowItem>
         {/if}
-        <Datatable.Data.Row.Item class={getStatusClass(item.coverage_status)}>
-          {#if item.coverage_status === ItemCoverageStatus.Approved && item.coverage_end_date}
-            <div class="red">Covered through {formatFriendlyDate(item.coverage_end_date)}</div>
+        <RowItem {item} className={getStatusClass(item.coverage_status)}>
+          {#if willEnd(item)}
+            <div class:red={!hasEnded(item)}>
```
forevermatt, line 296: It seems like "will end" (approved with a `coverage_end_date`) and "has ended" (inactive with a `coverage_end_date`) are mutually exclusive conditions.

I think `!hasEnded(item)` will always evaluate to true here.

src/components/Datatable/RowItem.svelte
forevermatt, line 1: It looks like this new RowItem component can only be used for showing a PolicyItem. Is that correct? If so, is there some way would could make that more readily apparent (such as in the component name or folder somehow)?

If not, that's okay, it was just minimally unexpected in the other diffs where `Datatable.Data.Row.Item` was being replaced by `RowItem` that it was also only available for use with PolicyItems.

src/components/Datatable/ItemsTable.svelte
hobbitronics: Good catch!

src/components/Datatable/RowItem.svelte
hobbitronics: Yeah, probably best to do that just in case sometime tries using it in ClaimsTable or elsewhere

merchako (changes requested):
Thank you! I look forward to seeing what y'all choose next for this.

src/components/Datatable/PolicyItems/RowItem.svelte
```diff
@@ -0,0 +1,18 @@
...
+import type { PolicyItem } from 'data/items'
+
+export let item: PolicyItem
+export let numeric = false
+export let className = ''
+
+const getRowClass = (item: PolicyItem) => {
+  if (noCoverage(item.coverage_status)) return `low-contrast ${className}`
+  return className
+}
```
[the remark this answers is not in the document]
hobbitronics: This syntax doesn't seem to work. Javascript isn't able to be parsed inside style tags.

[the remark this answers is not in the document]
merchako: Please treat it as pseudo code. You can put in a string value instead of the interpolated statement. Or maybe TypescriptSvelte have a way of inserting a value at the compiler level to keep to a single source of truth.

src/components/global.css
```diff
@@ -66,7 +66,12 @@ p {
...
-  footer, .drawer {
+  footer,
+  .drawer {
     display: none !important;
   }
 }
+
+td.low-contrast.mdc-data-table__cell {
+  color: var(--mdc-theme-text-secondary-on-light);
+}
```
merchako, line 77: I'm hoping we don't have to have such a specific CSS rule in _global.css_. I'm holding out for https://github.com/silinternational/cover-ui/pull/606/files#r1320888519

merchako (changes requested):

hobbitronics:
In order for css classes to be applied to svelte components they need to be
global. Our options are to make a less specific rule (which I was losing
the specificity war on), use :global() within the component file or alter
the Table.Row.Item in ui-components.

src/components/global.css
hobbitronics: see https://github.com/silinternational/ui-components/pull/221 for possible solution

src/components/Datatable/PolicyItems/RowItem.svelte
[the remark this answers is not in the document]
merchako: I don't know what the exact value for Inactive would be, but perhaps the result should be something like
```svelte
<Datatable.Data.Row data-coverage-status={item.coverage_status}>
   …
</Datatable.Data.Row>
…
<style>
[data-coverage-status=inactive] .mdc-data-table__cell {
  color: var(--mdc-theme-low-contrast);
}
</style>
```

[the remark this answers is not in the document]
hobbitronics: is data-coverage-status a prop? Also is there a way to check for 'disabled' as well?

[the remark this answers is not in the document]
merchako: I made up `data-coverage-status`. It isn't yet a prop, but I'm suggesting it could be an attribute on the HTML element (which you'd want to bind to a prop). 

We don't have a concept of being "disabled" on `DataTable` yet, but we could. In your _ui-components_ PR, I suggest using something like `emphasis: string enum` to name and style various cell display states. Disabled is one that I suggest to match the [MDC guide on emphasis](https://m2.material.io/design/color/text-legibility.html#text-backgrounds).

merchako (approved):

PR #607, opened 2023-08-31
less margin between drawer menu items CVR-655

hobbitronics:
[CVR-655](https://itse.youtrack.cloud/issue/CVR-655/Why-drawer-menu-items-have-extra-margin)

### DIFF
--- a/src/components/AppDrawer.svelte
+++ b/src/components/AppDrawer.svelte
@@ -44,9 +44,6 @@ $beforeUrlChange((event: CustomEvent, route: string, { url }: { url: string }) =
   display: block;
   margin: 0 auto;
 }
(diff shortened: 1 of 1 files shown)


src/components/AppDrawer.svelte
```diff
@@ -45,7 +45,7 @@ $beforeUrlChange((event: CustomEvent, route: string, { url }: { url: string }) =
   margin: 0 auto;
 }
 :global(.drawer .mdc-drawer__content div a.mdc-deprecated-list-item) {
-  margin: 16px 8px;
+  margin: 8px;
 }
```
merchako, line 49: I think we can remove this rule entirely. I don't think we need any margin.

merchako (changes requested):

PR #608, opened 2023-08-31
CVR-650 Set statement-name default value

forevermatt:
[CVR-650](https://itse.youtrack.cloud/issue/CVR-650) Set vehicle name default value

---

### Added
- Add `Vehicle` Risk Category constant
- Set a default value for Statement Name (regardless of category) but let the user customize (i.e. override) it


src/data/items.ts
```diff
@@ -42,6 +42,7 @@ export type RiskCategory = {
 export enum RiskCategoryNames {
   Stationary = 'Stationary',
   Mobile = 'Mobile',
+  Vehicle = 'Vehicle',
```
forevermatt, line 45: @briskt If we go with this implementation, will we need to add a migration to cover-api to add this new Risk Category to the database? Since that new database record will require a cost_center I assume not (it would be added manually), but wanted to check.

jason-jackson: I'd think it would have to unless I'm mistaken.

briskt: Yes, it will probably be added by a migration since we don't have any endpoints for managing risk categories. We should also consider adding at least a list endpoint (e.g. `GET /risk-categories`) so the risk category list doesn't have to be hardcoded here.

src/components/forms/ItemForm.svelte
```diff
@@ -63,8 +64,17 @@ $: make,
...
   marketValueUSD,
   accountablePersonId && categoryId && name && debouncedSave()
-$: selectedCategoryIsStationary =
-  $categories.find((c) => c.id === categoryId)?.risk_category?.name === RiskCategoryNames.Stationary
+$: selectedCategory = $categories.find((c) => c.id === categoryId)
+$: selectedCategoryIsStationary = selectedCategory?.risk_category?.name === RiskCategoryNames.Stationary
+$: selectedCategoryIsVehicle = selectedCategory?.risk_category?.name === RiskCategoryNames.Vehicle
+$: shortNameExample = assembleShortNameExample(selectedCategoryIsVehicle, make, model)
+
+const assembleShortNameExample = (isVehicle, make, model) => {
```
hobbitronics, line 72: Nice, want to throw this in itemFormHelpers.ts?

hobbitronics (approved):
Lgtm!

src/components/forms/ItemForm.svelte
```diff
@@ -214,7 +224,7 @@ span.label {
   <p>
     <TextFieldWithLabel
       label="Short name"
-      description="This label will appear on your statements."
+      description={'This label will appear on your statements.' + (shortNameExample ? ` Example: ${shortNameExample}` : '')}
```
merchako, line 227:
```suggestion
      description="Customize what will appear on your statements"
```

merchako (changes requested):
The user-given name wins, and the assembled name should be a _default value_, not an example in the helper text. I thought about setting the generated name in the description, but I'd prefer to give specific examples appropriate to each category (TBD). It's generally a good practice to [avoid placeholders](https://www.nngroup.com/articles/form-design-placeholders/), but specifically I want to pre-populate this field because every item *will need* a value here, and I want to set users up to have helpful values by default without having to think up their own.

This is for *all* items, not only vehicles.

Here's what should win:

1. User-given value, if any
2. \`{make} {model} {?year}\`, if unique
3. `{make} {model} {?year} ({uniqueIdentifier.slice(-6)})`

persistentHelperText = "Customize what will appear on your statements"

src/components/forms/ItemForm.svelte
forevermatt: That wording fix will be coming in a separate PR (re: [CVR-651](https://itse.youtrack.cloud/issue/CVR-651/Update-ItemForm)). Thanks, though. 👍

briskt:
> * User-given value, if any
> * `{make} {model} {?year}`, if unique
> * `{make} {model} {?year} ({uniqueIdentifier.slice(-6)})`

@merchako of course the second ("if unique") value will require an API call, and thus a significant amount of complexity. Would it be a problem to skip that and always append the unique identifier if no user-given value is provided?

merchako:
> Would it be a problem to skip that and always append the unique identifier if no user-given value is provided?

I was wondering about that. Okay, go ahead and skip the API call and always put in the unique ID slice.

src/data/items.ts
forevermatt: Okay, good to know the new risk category needs to be added by a migration.

As far as a `GET /risk-categories` endpoint, I don't have an objection per se, but it _is_ still helpful to have it hard-coded here so that we have a TypeScript constant we can use in comparisons in the code. Given that, I don't know whether the risk-categories endpoint gains us anything. Or did I miss some benefit of what you were suggesting?

briskt: Just trying to avoid duplication. I guess the only way to truly avoid duplication in some cases is to make a full-stack app.

briskt (approved):

merchako (approved):
Looks good to me. Thank you!

PR #610, opened 2023-09-01
CVR-651 Update ItemForm (partially)

forevermatt:
This partially implements "[CVR-651](https://itse.youtrack.cloud/issue/CVR-651) Update ItemForm"

---

### Added
- Add section headings to ItemForm

### Changed (non-breaking)
- Move "Short Name" field down to the "For your own use" section
- Move "Notes" text area down to the "For your own use" section
- Update labels and persistent helper text in ItemForm
- Update wording in currency-convert text (everywhere)
- Replace `TextFieldWithLabel` usages with `TextField` (but preserve desired widths) in ItemForm

### Fixed
- Get updated version of ui-components (for `TextField` improvements)
- Use more `label` attributes and fewer custom `span`s
- Use new "statement name" (not "short name") in form validation, too


src/components/forms/ItemForm.svelte
```diff
@@ -199,50 +196,65 @@ span.label {
...
   <p>
-    <TextFieldWithLabel
-      label="Short name"
-      description="This label will appear on your statements."
-      required
-      bind:value={name}
+    <TextField
+      label="Model (optional)"
+      class="mw-300"
+      description="e.g., iPhone 10 Max 64 GB, A1921, or Land Cruiser"
```
briskt, line 210: @merchako would it be worth holding back the vehicle text (e.g. "Land Cruiser", "VIN") so the other changes can be released sooner?

briskt (approved):
If these changes are important to get released right away, the vehicle-specific text should be removed (for now).

src/components/forms/ItemForm.svelte
merchako: I wouldn't worry about it, @briskt . The word "VIN" already exists in the Unique Identifier helper text, so these examples aren't making it any _worse_. Then you don't have to deal with the complexity of holding it back.

Of course, if you want to hold it back anyway, I won't get in your way 🫡

merchako:
Is there a way to view this on Cloudflare? I’m not able to see some pages on my local build because I my AppsDev account doesn't have a Household associated with it.

src/components/forms/ItemForm.svelte
merchako, line 1: None of the content on this page is a paragraph, so we should be able to get rid of all `<p>` tags.

src/components/forms/items/itemFormHelpers.ts
```diff
@@ -13,7 +13,7 @@ export const areMakeAndModelRequired = (item: PolicyItem, categoryId: string): b
 export const validateForSave = (formData: NewItemFormData | UpdateItemFormData): void => {
   assertHas(formData.accountablePersonId, 'Please Assign an Accountable Person')
```
merchako, line 14:
```suggestion
  assertHas(formData.accountablePersonId, 'Please indicate who will be responsible the item')
```

merchako (approved):
None of these are

src/components/forms/ItemForm.svelte
forevermatt: Unfortunately(-ish), we need to wrap the form elements in _something_ because the ui-component library's "Form" element's CSS sets some custom spacing around every immediate child of a `form` element, so removing the `p` tags messes up some of the spacing around the persistent helper text elements otherwise.

forevermatt:
> Is there a way to view this on Cloudflare? 

To more directly answer your question, yes and no. There are Cloudflare Pages builds for PRs (see the "Checks" tab near the top of this PR), but our CORS settings for cover-api prevent them from being functional copies of the Cover UI.

PR #611, opened 2023-09-08
CVR-644 add icon to itemsTable

hobbitronics:
[CVR-644](https://itse.youtrack.cloud/issue/CVR-644/ItemTable-Prepend-item-names-with-category-icon)

- Add icon to ItemsTable
- Add "iconify-icon"
- Add itemCategoryName

image

### DIFF
--- a/src/components/Datatable/PolicyItems/itemTableHelpers.ts
+++ b/src/components/Datatable/PolicyItems/itemTableHelpers.ts
@@ -1,3 +1,4 @@
+import type { itemCategoryName } from 'data/itemCategories'
 import { ItemCoverageStatus, type PolicyItem } from 'data/items'
 
 export function hasEnded(item: PolicyItem): boolean {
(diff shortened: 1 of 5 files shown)


src/components/Datatable/PolicyItems/itemTableHelpers.ts
```diff
@@ -11,3 +12,34 @@ export function willEnd(item: PolicyItem): boolean {
...
+    case 'Musical instruments':
+      return 'mdi:bugle'
+    case 'Photography and videography':
+      return 'mdi:camera'
+    case 'Vehicles':
+      return 'mdi:car'
+    default:
+      return 'mdi:help-circle'
+  }
+}
```
merchako, line 45: This style of using display strings will likely become brittle if we ever change the Menu Item text, which is likely. We had a similar issue with the _jobs-dashboard_ configuration data. This can merge now, but I recommend creating a similar object-based solution [like you did for the jobs board](https://github.com/silinternational/jobs-dashboard/pull/98).

merchako (approved):
Great work! Just what I wanted.

src/components/Datatable/PolicyItems/itemTableHelpers.ts
hobbitronics: This is complicated by the fact that the Item Categories come from the backend. We could use the categories `id` instead of the name for this mapping, or else we could get these icon strings in the backend, which would be best I think.

merchako: Can there be one config file shared by both repos? I miss SvelteKit.

merchako: BTW, what I intended to quote was
```
    case 'Musical instruments':
       return 'mdi:bugle'
```
The "Musical instruments" is the fragile part.

hobbitronics:
Is not even a file in the repo. It's in the database, but adding the icons
there should be OK. We should make a card for the backend so we won't have
this fragile connection.

PR #612, opened 2023-09-11
npm→bun

merchako:
Use Bun instead of NPM. It's a lot faster.
requires [installation of bun.sh](https://bun.sh/docs/installation)


package.json
```diff
@@ -44,7 +44,7 @@
     "eslint-plugin-svelte3": "^3.2.1",
     "lodash-es": "^4.17.21",
     "material-components-web": "^14.0.0",
-    "npm-run-all": "^4.1.5",
+    "bun-run-all": "^4.1.5",
```
hobbitronics, line 47: I don't think this will work

hobbitronics (changes requested):
We need to make sure all the commands work. I haven't been able to get docker to work. I think it would be nice to use bun-bundler.

package.json
merchako: oh whoops!

merchako: Resolved in [fd3bf81](https://github.com/silinternational/cover-ui/pull/612/commits/fd3bf817ce7c0470f63e7aa8ac3cb54e3b14d7c7)

package.json
```diff
@@ -5,18 +5,18 @@
...
     "clean": "run-p -l clean:build clean:deps",
     "clean:build": "rimraf dist/bundle* dist/workbox*",
-    "clean:deps": "npm prune",
+    "clean:deps": "bun prune",
     "format": "prettier --write ./src/**/*.{js,ts,svelte}",
     "lint": "eslint ./src/**/*.{js,ts,svelte}",
     "lint:fix": "eslint --fix ./'src/**/*.{js,ts,svelte}'",
-    "prelint": "npm run format",
+    "prelint": "bun run format",
     "local:dev": "run-p -l build:dev serve:http",
```
hobbitronics, line 20:
```suggestion
     "local:dev": "conc \"bun run build:dev\" \"bun run serve:http\"",
```

package.json
```diff
@@ -5,18 +5,18 @@
...
     "local:prod": "run-p -l build:prod serve:http",
```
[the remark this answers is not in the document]
hobbitronics: You'll need `npm i -D concurrently`

briskt:
I think we need to remove bun.lockb from the repo and add it to .gitignore

hobbitronics:
I've merged the other bun PR so closing this. Hopefully bun will add more compatibility so we can use it in Docker with Rollup.

PR #613, opened 2023-09-12
add make bun for developers who prefer bun

hobbitronics:
- This give developers who have experienced slow or buggy builds with node an alternative tool, bun. M1/M2 have poor performance of node and docker and sometimes the build doesn't seem to be triggered by changes or the build hangs.


briskt (approved):

Makefile
```diff
@@ -18,3 +18,6 @@ build-prod:
 
 clean:
 	npm run clean
+
+bun:
```
briskt, line 22: Might be good to start documenting some of these options in the README

PR #622, opened 2023-09-18
must have forgotten to npm i

hobbitronics:


package-lock.json
```diff
@@ -21,14 +21,14 @@
         "@types/qs": "^6.9.7",
         "@typescript-eslint/eslint-plugin": "^4.33.0",
         "@typescript-eslint/parser": "^4.33.0",
+        "concurrently": "^8.2.1",
```
forevermatt, line 24: These look like package-lock.json changes I saw [here](https://github.com/silinternational/cover-ui/pull/613/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R43), and I think that PR was already merged into `develop`. Are you still seeing the need for this PR? (I'm just trying to catch up on all of the changes since I was out.)

hobbitronics: I seems I installed concurrently but didn't uninstall npm-run-all before running npm install there @forevermatt

forevermatt: Okay. If you think this is still needed, I'm fine with that.

forevermatt (approved):

PR #634, opened 2023-09-27
use TextField CVR-647

hobbitronics:
[CVR-647](https://itse.youtrack.cloud/issue/CVR-647/Replace-all-TextFieldWithLabel-with-TextField-and-deprecate)

## Add
- add AccountablePeopleList component
## Remove
- removed TextFieldWithDescription
## Change
- use TextField with label and description props
- refactor code into component

image
image
image
image

PR #638, opened 2023-10-02
change some fields and labels

hobbitronics:
- Addresses some suggestions made in https://github.com/silinternational/cover-ui/pull/634

### DIFF
--- a/src/pages/admin/entities/[entityId].svelte
+++ b/src/pages/admin/entities/[entityId].svelte
@@ -5,7 +5,7 @@ import { formatPageTitle } from 'helpers/pageTitle'
 import { entityDetails } from 'helpers/routes'
 import { metatags } from '@roxi/routify'
 import { onMount } from 'svelte'
-import { Button, Page, Progress, TextField, setNotice } from '@silintl/ui-components'
+import { Button, Page, TextField, setNotice } from '@silintl/ui-components'
 
 export let entityId: string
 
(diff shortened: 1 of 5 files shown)


forevermatt (approved):

src/pages/admin/entities/[entityId].svelte
```diff
@@ -5,7 +5,7 @@ import { formatPageTitle } from 'helpers/pageTitle'
 import { entityDetails } from 'helpers/routes'
 import { metatags } from '@roxi/routify'
 import { onMount } from 'svelte'
-import { Button, Page, TextField, setNotice } from '@silintl/ui-components'
+import { Button, Page, Progress, TextField, setNotice } from '@silintl/ui-components'
```
Comment (hobbitronics) [inline src/pages/admin/entities/[entityId].svelte:8]:
```suggestion
import { Button, Page, TextField, setNotice } from '@silintl/ui-components'
```

Comment (hobbitronics) [inline src/pages/admin/entities/[entityId].svelte]: I didn't end up using Progress as we already have Linear in the base layout.

src/pages/admin/entities/[entityId].svelte
```diff
@@ -36,11 +36,11 @@ const onSave = async () => {
...
       <h4>Code: {entity.code}</h4>
 
-      <TextField label="Name" bind:value={entity.name} />
+      <TextField required label="Name" bind:value={entity.name} />
 
-      <TextField label="Income Account" bind:value={entity.income_account} />
+      <TextField required label="Income Account" bind:value={entity.income_account} />
 
-      <TextField label="Parent Entity" bind:value={entity.parent_entity} />
+      <TextField required label="Parent Entity" bind:value={entity.parent_entity} />
```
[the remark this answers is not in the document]
Comment (hobbitronics) [reply src/pages/admin/entities/[entityId].svelte]: Good catch. I will assume it's optional, but the api is letting us update all the fields here with empty values, so we might have to dig a little deeper to find out if its needed.

PR #643, opened 2023-10-04
tailwind is not tailwindcss

hobbitronics:
## Changed
- update a few Prettier related deps

- While starting to create an issue for tailwindcss because of multiple dependabot warnings I noticed we had tailwind installed so I installed the correct tailwindcss instead.

- I guess this means none of the utility classes were being applied yet, but I don't think there were many.
- we should verity tailwind css is working

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,8 +53,8 @@
         "svelte": "^3.55.1",
         "svelte-check": "^2.9.0",
         "svelte-preprocess": "^4.10.7",
-        "tailwind": "^4.0.0",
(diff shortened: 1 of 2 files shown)


mtompset (commented):
Given that I don't know how to test it, I'll leave this to someone else to sign off. However, it looks like a reasonable set of changes.

forevermatt (approved):
Bizarre. I'm sure at least some of the Tailwind CSS classes were working... perhaps we were "getting lucky" because of a sub-dependency or something like that. Thanks for fixing this.

PR #645, opened 2023-10-05
remove start and end date from item and fix ts errors

hobbitronics:
## Removed
- start and end date from ItemForm

## Added
- make check
- make lint

## Fixed
- broken import
- lots of ts errors/warning (down to 5 mostly from bad routify types)

### DIFF
--- a/src/components/FilePreview/FilePreview.svelte
+++ b/src/components/FilePreview/FilePreview.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import Banner from '../components/Banner.svelte'
+import Banner from '../Banner.svelte'
 import { formatDate } from '../../helpers/dates'
 import { Button } from '@silintl/ui-components'
 import { createEventDispatcher } from 'svelte'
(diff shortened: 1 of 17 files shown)


src/components/FilePreview/FilePreview.svelte
```diff
@@ -31,7 +31,7 @@ function onDelete(event: CustomEvent, id: string) {
 }
 </style>
 
-<div class="preview flex justify-between align-items-center br-8px p-10px mb-1 {$$props.class}" on:click>
+<div class="preview flex justify-between align-items-center br-8px p-10px mb-1 {$$props.class}" on:click on:keydown={(e) => {if (e.key === 'Enter') e.currentTarget.click()}}>
```
briskt, line 34: Doesn't look related to "remove start and end date from item". Can you add something to the PR description and/or move it to its own PR?

briskt (approved):
Thanks for bringing this up to date and cleaning up the types. Can you add to the PR title something about Typescript cleanup? Better to use separate PRs, but this is ok for now.

src/components/FilePreview/FilePreview.svelte
hobbitronics: Sorry, I wanted to do ts checks to make sure this change didn't cause any other issues and got carried away.

briskt: It looked to me like this added handling for Enter keystroke. Is that correct?

hobbitronics: yes, there was a ts warning about not having non interactive elements with click handler, but no key handler.

PR #646, opened 2023-10-05
CVR-705 (b): Fix incorrect reliance on coverage_start_date before approval

forevermatt:
### Fixed
- CVR-705 (b): Fix incorrect reliance on `coverage_start_date` before approval


briskt (approved):
Good catch

PR #647, opened 2023-10-05
CVR-705 (c): Update other checkout details re: monthly vs yearly

forevermatt:
### Fixed
- Fix start-date calculations used for checkout page (re: monthly vs yearly)
- Fix renewal-date calculations used for checkout page (re: monthly vs yearly)
- Fix `formatDate()` to treat Go empty-date values as not a date

### Added
- Add `getPremiumDescription()` helper
- Add `getRenewalDate()` helper
- Add `getStartDate()` helper
- Add `isBeforeMonthlyCutoff()` helper
- Add `isItemDraft()` helper
- Add `isMeaningfulDateString()` helper
- Add `isMonthly()` helper
- Add `startOfFutureMonth()` helper

### DIFF
--- a/src/components/ItemDetails.svelte
+++ b/src/components/ItemDetails.svelte
@@ -27,7 +27,7 @@ $: $policies && (policy = getPolicyById(policyId))
 $: statusText = getItemStatusText(item)
 $: status = (item.coverage_status || '') as ItemCoverageStatus
 $: showRevisionMessage = item.status_reason && status === ItemCoverageStatus.Revision
-$: startDate = formatDate(item.coverage_start_date)
+$: startDate = formatDate(item?.coverage_start_date) || '(when approved)'
 $: endDate = formatDate(item.coverage_end_date)
 $: renewDate = getRenewalDate()
 $: commonDetails = {
--- a/src/components/ItemDetails.svelte
+++ b/src/components/ItemDetails.svelte
@@ -10,7 +10,7 @@ import {
 } from 'data/items'
 import { getPolicyById, loadPolicy, policies, Policy, PolicyType } from 'data/policies'
 import { isBeforeMonthlyCutoff } from 'helpers/coverage'
-import { formatDate } from 'helpers/dates'
+import { formatDate, startOfFutureMonth } from 'helpers/dates'
 import { formatMoney } from 'helpers/money'
 import InfoBoxModal from './InfoBoxModal.svelte'
 import { formatDistanceToNow } from 'date-fns'
@@ -71,23 +71,14 @@ const getItemStatusText = (item: PolicyItem) => {
 
 const getMonthlyRenewalDate = (item: PolicyItem): string => {
   if (itemIsApproved(item)) {
-    const nextMonth = new Date()
-    nextMonth.setMonth(nextMonth.getMonth() + 1)
-    nextMonth.setDate(1)
-    return formatDate(nextMonth.toISOString())
+    return startOfFutureMonth(1)
   }
 
   if (isItemDraft(item)) {
     if (isBeforeMonthlyCutoff()) {
-      const nextMonth = new Date()
-      nextMonth.setMonth(nextMonth.getMonth() + 1)
-      nextMonth.setDate(1)
-      return formatDate(nextMonth.toISOString())
+      return startOfFutureMonth(1)
     } else {
-      const monthAfterNext = new Date()
-      monthAfterNext.setMonth(monthAfterNext.getMonth() + 2)
-      monthAfterNext.setDate(1)
-      return formatDate(monthAfterNext.toISOString())
+      return startOfFutureMonth(2)
     }
   }
 
--- a/src/components/ItemDetails.svelte
+++ b/src/components/ItemDetails.svelte
@@ -1,10 +1,9 @@
 <script lang="ts">
-import Banner from './Banner.svelte'
 import ItemBanner from './banners/ItemBanner.svelte'
 import MessageBanner from './banners/MessageBanner.svelte'
 import { PolicyItem, ItemCoverageStatus } from 'data/items'
 import { getPolicyById, loadPolicy, policies, Policy, PolicyType } from 'data/policies'
-import { formatDate } from '../helpers/dates'
+import { formatDate } from 'helpers/dates'
 import { formatMoney } from 'helpers/money'
 import InfoBoxModal from './InfoBoxModal.svelte'
 import { formatDistanceToNow } from 'date-fns'
--- a/src/components/ItemDetails.svelte
+++ b/src/components/ItemDetails.svelte
@@ -16,7 +16,6 @@ export let policyId: string
 export let isAdmin: boolean = false
 
 let policy: Policy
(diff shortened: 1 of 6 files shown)


src/helpers/coverage.ts
```diff
@@ -0,0 +1,67 @@
+import { BillingPeriod, isItemDraft, itemIsApproved, MonthlyCutoffDay, PolicyItem } from 'data/items'
+import { formatDate, isMeaningfulDateString, startOfFutureMonth } from 'helpers/dates'
+import { formatMoney } from 'helpers/money'
+
+const getMonthlyRenewalDate = (item: PolicyItem): string => {
+  if (itemIsApproved(item)) {
```
briskt, line 6: Do you need argument checking, here and in `getStartDate` and `getYearlyRenewalDate`? I see that function signatures vary slightly. Some have `(item: PolicyItem)` and some have `(item: PolicyItem | undefined)`. Will that inconsistency bite us later?
```js
  if (!item || !item.id) {
    return ''
  }
```

src/components/ItemDetails.svelte
```diff
@@ -29,9 +28,9 @@ $: $policies && (policy = getPolicyById(policyId))
 $: statusText = getItemStatusText(item)
 $: status = (item.coverage_status || '') as ItemCoverageStatus
 $: showRevisionMessage = item.status_reason && status === ItemCoverageStatus.Revision
-$: startDate = formatDate(item.coverage_start_date)
+$: startDate = getStartDate(item)
```
briskt, line 31: Do you want this same change in Checkout.svelte?

src/components/ItemDetails.svelte
```diff
@@ -29,9 +28,9 @@ $: $policies && (policy = getPolicyById(policyId))
...
 $: endDate = formatDate(item.coverage_end_date)
-$: renewDate = formatDate(`${renewYear}-01-01`)
+$: renewDate = getRenewalDate(item)
```
briskt, line 33: Similarly, should `getRenewalDate` be used in Checkout.svelte?

src/components/ItemDetails.svelte
```diff
@@ -49,7 +48,7 @@ $: minimumDeductible = item?.category?.minimum_deductible
 $: minimumDeductibleDescription = minimumDeductible ? `(subject to ${formatMoney(minimumDeductible)} minimum)` : ''
 $: moneyDetails = {
   Value: formatMoney(item.coverage_amount),
-  Premium: `${formatMoney(item.annual_premium)} / yr`,
+  Premium: getPremiumDescription(item),
```
briskt, line 51: I see `formatMoney(item.annual_premium)` in Checkout.svelte as well.

src/data/items.ts
```diff
@@ -359,6 +359,10 @@ export const itemIsApproved = (item: PolicyItem): boolean => {
   return item.coverage_status === ItemCoverageStatus.Approved
 }
 
+export const isItemDraft = (item: PolicyItem): boolean => {
```
briskt, line 362: Naming is a little different here, as compared to `itemIsApproved` and `itemIsInactive`. Perhaps this should be `itemIsDraft`.

briskt (approved):
Looks great, I like the readability. I offered a few comments that are not essential changes.

hobbitronics (approved):
Looks good, hopefully no merge conflicts.

src/components/ItemDetails.svelte
```diff
@@ -1,10 +1,10 @@
 <script lang="ts">
-import Banner from './Banner.svelte'
```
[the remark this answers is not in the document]
hobbitronics: I doesn't seem to cause a merge conflict.

src/helpers/coverage.ts
hobbitronics: it will reduce ts errors to have `(item: PolicyItem | undefined)` if you might pass in `undefined`

src/components/ItemDetails.svelte
hobbitronics: ItemDetails is used in Checkout.svelte.

hobbitronics:
@forevermatt I created a [PR](https://github.com/silinternational/cover-ui/pull/648) targeting this one that fixes a few things and addresses the remaining youtrack issues. Merge it in if you like or it can go into develop later.

src/helpers/coverage.ts
forevermatt: I actually intentionally made some of these functions just accept a PolicyItem, choosing to handle the `undefined` cases in the functions that call these. If we do later change the code to call these with something that could be undefined, the IDE will show a warning, which can help us catch those coding mistakes.

src/data/items.ts
forevermatt: Yeah, I'd like to suggest changing the names of the other functions to match this naming pattern, but didn't want to include that suggestion in this PR for the sake of limiting scope creep.

Naming values as startments (e.g. `itemIsApproved`) and functions as questions (e.g. `isItemApproved()`) seems like a useful way to avoid naming collisions between functions that answer yes/no questions and the variables we use to store the returned value of those functions.

For example, this lets us do this:
```js
$: itemIsDraft = isItemDraft(item)
```

Thoughts? Again, the renaming (for consistency) can happen in a separate PR, whether we go with this change or revert my non-matching function name back to matching the others. Way to notice the little details, though!

src/components/ItemDetails.svelte
forevermatt: I don't see any code using `formatDate()` in `components/checkout/Checkout.svelte`... can help you me find what you're referring to?

src/helpers/coverage.ts
forevermatt: For the most part, I only restricted the type to `PolicyItem` on functions that aren't exported, so effectively "private" functions. (The one exception I see is `getStartDate()`.)

src/components/ItemDetails.svelte
briskt: I must have had an old copy of that file on my machine.

src/helpers/coverage.ts
briskt: I'm fine with the way you have it if it was by design.

src/components/ItemDetails.svelte
briskt: I was looking at old code.

src/components/ItemDetails.svelte
briskt: I was looking at old code.

src/data/items.ts
briskt: As I mentioned in our Slack conversation, I recommend either (1) changing the naming of the other functions in a separate PR and merge that before this PR, or (2) use the "old" naming here and file an issue to later make the naming change.

forevermatt: Created a separate issue for making this name change:

[CVR-739](https://itse.youtrack.cloud/issue/CVR-739) Rename `itemIsApproved()` and similar functions to put the "is" first: `isItemApproved()`

PR #648, opened 2023-10-09
CVR-736 CVR-737 Add minimum premium to checkout and update ItemCategory

hobbitronics:
[CVR-736](https://itse.youtrack.cloud/issue/CVR-736/Update-UI-to-reflect-the-minimum-premium)
[CVR-737](https://itse.youtrack.cloud/issue/CVR-737/Update-UI-logic-to-reflect-that-the-minimum-premium-applies-to-both-the-annual-premium-and-the-prorated-annual-premium)

# Fixed
- prettier not formatting again
- isBeforeMonthlyCutoff value was being used
- ts errors

## Added
- ItemCategory to PolicyItem

## Changed
- Yearly checkout message to show minimum and always show payment
image

### DIFF
--- a/src/components/checkout/MonthlyCheckoutMessage.svelte
+++ b/src/components/checkout/MonthlyCheckoutMessage.svelte
@@ -12,7 +12,7 @@ $: monthlyPremium = item.monthly_premium
 </script>
 
 <span>
-  {#if isBeforeMonthlyCutoff}
+  {#if isBeforeMonthlyCutoff()}
     Pay {formatMoney(monthlyPremium)} from {org} account {accountOrHouseholdId}
     each month.
   {:else}
--- a/src/components/checkout/YearlyCheckoutMessage.svelte
+++ b/src/components/checkout/YearlyCheckoutMessage.svelte
@@ -8,17 +8,24 @@ export let item: PolicyItem | undefined = undefined
 export let org: string = ''
 
 $: annualPremium = item?.annual_premium
-$: proratedAnnualPremiumOrMin = Math.max(item?.prorated_annual_premium || 0, item?.category?.minimum_premium || 0)
+$: proratedAnnualPremium = item?.prorated_annual_premium || 0
 
 $: year = new Date().getFullYear()
 $: renewYear = year + 1
 $: renewDate = formatDate(`${renewYear}-01-01`)
-$: premiumEqualsProrated = annualPremium === proratedAnnualPremiumOrMin
+$: premiumEqualsProrated = annualPremium === proratedAnnualPremium
 </script>
 
 <span>
-  Pay {formatMoney(proratedAnnualPremiumOrMin)}
-  {premiumEqualsProrated ? 'for' : 'for the remainder of'}
-  {year} from {org} account
-  {accountOrHouseholdId}. Auto-renew and pay {formatMoney(annualPremium)} on {renewDate}.
+  {#if proratedAnnualPremium > 0}
+    {#if premiumEqualsProrated}
+      Pay {formatMoney(proratedAnnualPremium)} for {year} from {org} account {accountOrHouseholdId}.
+    {:else}
+      Pay {formatMoney(proratedAnnualPremium)} for the remainder of {year} from {org} account
+      {accountOrHouseholdId}.
+    {/if}
+    Auto-renew and pay {formatMoney(annualPremium)} on {renewDate}.
+  {:else}
+    No payment needed right now. Auto-renew for {formatMoney(annualPremium)} on {renewDate}, paid from {org} account {accountOrHouseholdId}.
+  {/if}
 </span>
--- a/src/components/checkout/YearlyCheckoutMessage.svelte
+++ b/src/components/checkout/YearlyCheckoutMessage.svelte
@@ -8,17 +8,12 @@ export let item: PolicyItem | undefined = undefined
 export let org: string = ''
 
 $: annualPremium = item?.annual_premium
-$: proratedAnnualPremiumOrMin = getGreaterOfTwoValues(
-  item?.prorated_annual_premium || 0,
-  item?.category?.minimum_premium || 0
-)
+$: proratedAnnualPremiumOrMin = Math.max(item?.prorated_annual_premium || 0, item?.category?.minimum_premium || 0)
 
 $: year = new Date().getFullYear()
 $: renewYear = year + 1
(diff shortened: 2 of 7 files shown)


src/components/checkout/YearlyCheckoutMessage.svelte
```diff
@@ -1,26 +1,29 @@
...
+import { formatDate } from 'helpers/dates'
 import { formatMoney } from 'helpers/money'
 
 export let accountOrHouseholdId: string = ''
 export let item: PolicyItem | undefined = undefined
 export let org: string = ''
 
 $: annualPremium = item?.annual_premium
-$: proratedAnnualPremium = item?.prorated_annual_premium
+$: proratedAnnualPremiumOrMin = getGreaterOfTwoValues(
```
jason-jackson, line 11: Any reason not using `Math.max`?

jason-jackson (commented):
Looks mostly good, would like to check some more things before I approve (like checkout page), but if it looks fine to others I don't have any qualms that hold this up.

src/components/checkout/YearlyCheckoutMessage.svelte
hobbitronics: Just felt like reinventing the wheel.

src/components/checkout/YearlyCheckoutMessage.svelte
```diff
@@ -8,17 +8,12 @@ export let item: PolicyItem | undefined = undefined
 export let org: string = ''
 
 $: annualPremium = item?.annual_premium
-$: proratedAnnualPremiumOrMin = getGreaterOfTwoValues(
-  item?.prorated_annual_premium || 0,
-  item?.category?.minimum_premium || 0
-)
+$: proratedAnnualPremiumOrMin = Math.max(item?.prorated_annual_premium || 0, item?.category?.minimum_premium || 0)
```
briskt, line 11: As you noticed and commented in CVR-736, you can avoid this altogether since the backend adjusts the premium by applying the minimum.

src/components/checkout/YearlyCheckoutMessage.svelte
```diff
@@ -1,26 +1,24 @@
...
 
 <span>
-  {#if proratedAnnualPremium > 100}
-    Pay {formatMoney(proratedAnnualPremium)} for the remainder of {year} from {org} account
-    {accountOrHouseholdId}. Auto-renew and pay {formatMoney(annualPremium)} on {renewDate}.
-  {:else}
-    No payment needed right now. Auto-renew for {formatMoney(annualPremium)} on {renewDate},
-    paid from {org} account {accountOrHouseholdId}.
-  {/if}
+  Pay {formatMoney(proratedAnnualPremiumOrMin)}
```
briskt, line 20: We may still want to check `proratedAnnualPremium > 0` since it's technically possible for no immediate payment to be required, especially for inexpensive items near the end of the year.

briskt (approved):
Approved, though it's worth considering whether we may still want the "no payment needed right now" message.

src/helpers/money.ts
```diff
@@ -5,7 +5,7 @@ export const convertToCents = (dollars?: number | string): number => {
   return Math.round(Number(dollars) * 100) // Round to avoid #'s like 7001.000000000001
 }
 
-export const formatMoney = (cents: number): string => {
+export const formatMoney = (cents: number | undefined): string => {
   if (!cents || !Number.isFinite(+cents)) {
```
forevermatt, line 9: Good catch on updating this to be in line with our actual usage of it (which does include passing it `undefined` sometimes).

However, if an `undefined` is passed, I think it may be more appropriate to return an empty string, rather than explicitly saying `$0.00` when (perhaps) some data simply hasn't loaded yet. Thoughts?
```suggestion
export const formatMoney = (cents: number | undefined): string => {
  if (cents === undefined) {
    return ''
  }
  if (!cents || !Number.isFinite(+cents)) {
```

src/data/items.ts
```diff
@@ -61,7 +62,7 @@ export type PolicyItem = {
   accountable_person: AccountablePerson
   annual_premium: number
   billing_period: number /* in months, e.g. 1 or 12 */
-  category: any /*ItemCategory*/
+  category: ItemCategory
```
forevermatt, line 65: This seems fine... I wonder why we didn't have this strongly typed before. Do you happen to remember?

src/helpers/money.ts
briskt: I agree, particularly if it's only `undefined` during page loading. In any case, showing "$0.00" would be misleading.

src/components/checkout/YearlyCheckoutMessage.svelte
```diff
@@ -1,26 +1,24 @@
...
+  {premiumEqualsProrated ? 'for' : 'for the remainder of'}
```
forevermatt, line 21: Would you mind splitting this out into two complete / separate blocks of content (rather than having conditionals changing the phrasing mid-sentence)?

src/components/checkout/MonthlyCheckoutMessage.svelte
```diff
@@ -12,7 +12,7 @@ $: monthlyPremium = item.monthly_premium
 </script>
 
 <span>
-  {#if isBeforeMonthlyCutoff}
+  {#if isBeforeMonthlyCutoff()}
```
forevermatt, line 15: Thanks for catching this 🙄

forevermatt (commented):
Mostly looks good, just a few comments which ask for changes

src/components/checkout/YearlyCheckoutMessage.svelte
hobbitronics: I saw where the minimum was applied to the premium (yearly/monthly), but not the prorated premium. For my understanding, can you point me to the code that does that?

src/data/items.ts
hobbitronics: As the type was out of date w/API, we may not have realized that this type was the same on the PolicyItem.

src/components/checkout/YearlyCheckoutMessage.svelte
briskt: Thanks for asking. I forgot that it isn't applied to the annual prorated premium. It doesn't affect anything at the moment since only vehicles have a minimum and they aren't prorated. But I filed an issue to fix this anyway.

PR #649, opened 2023-10-09
CVR-705 (d): Update the Items Table re: monthly vs yearly billing periods

forevermatt:
### Fixed
- When sorting by premium, use the field being displayed 
  * Previously it was sorting by `prorated_annual_premium`, but displaying `annual_premium`. If that was intentional, please let me know.

### Added
- For items billed monthly, show that in the Items Table

Screenshot 2023-10-09 at 3 07 01 PM

---

I wrestled a little bit with various ways to handle showing annual vs. monthly premiums. This approach seems to work, since it both (A) visually indicates which are paid monthly, and what that monthy amount is, and (B) still shows the annual premium so that sorting the table by the Premium column still produces results that make sense (since the primary number shown is the annual premium, even for items billed monthly).


briskt (approved):
I like that presentation. Looks good to me.

PR #650, opened 2023-10-11
move app dependencies into `dependencies` section of package.json

briskt:
This has two benefits: (1) allows scanning with tools like Deepfactor and Snyk, (2) makes it more clear which dependencies are used for development and which are necessary for the app itself.

See the third section of https://jsramblings.com/do-dependencies-devdependencies-matter-when-using-webpack/


package.json
```diff
@@ -40,24 +35,39 @@
     "@types/lodash-es": "^4.17.7",
     "@types/node": "^16.11.6",
     "@types/qs": "^6.9.8",
```
forevermatt, line 37: I would guess that all of the `@types/` would actually be devDependencies. @hobbitronics, thoughts?

forevermatt (commented):
One comment, but the rest of it looks good to me. Thanks for sifting through those.

package.json
```diff
@@ -67,14 +77,6 @@
     "rollup-plugin-postcss": "^4.0.2",
     "rollup-plugin-svelte": "^7.1.6",
     "rollup-plugin-terser": "^7.0.2",
```
hobbitronics, line 79: I would say rollup and these plugins are `dependencies` since they're involved in making the bundle so could have CVE's that affect it.

hobbitronics (changes requested):
My thought is anything that could leak vulnerabilities into our code is a `dependency` since we are doing this primarily to scan for vulnerabilities. See comments for specific cases.

hobbitronics (approved):
I may be wrong on my initial intuition since it seems the convention is that bundlers are a devDependency, but I still wonder how you catch vulnerabilities that these may introduce.

briskt:
I'll go with the latest recommendations, but we can change at any time. It might help to see what Deepfactor says about these.

PR #651, opened 2023-10-11
scan with Deepfactor

briskt:


hobbitronics (approved):
Looks like it ran, and looks good to me.

PR #654, opened 2023-10-12
CVR-743  add monthly premium to recent claim section

hobbitronics:
[CVR-743](https://itse.youtrack.cloud/issue/CVR-743/monthly-annual-premiums-in-RecentActivityTable)
- I missed adding this to the recent claims table

### DIFF
--- a/src/components/RecentActivityTable.svelte
+++ b/src/components/RecentActivityTable.svelte
@@ -59,7 +59,14 @@ const getFormattedClaimItemPremium = (claim: Claim): string => {
           <RowItem>{recentChange.Claim.status_change}</RowItem>
           <RowItem>{getClaimItemPersonName(recentChange.Claim)}</RowItem>
           <RowItem numeric>{getFormattedClaimItemValue(recentChange.Claim)}</RowItem>
-          <RowItem numeric>{getFormattedClaimItemPremium(recentChange.Claim)}</RowItem>
+          <RowItem numeric>
+            {getFormattedClaimItemPremium(recentChange.Claim)}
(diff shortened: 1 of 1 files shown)


briskt (approved):

src/components/RecentActivityTable.svelte
```diff
@@ -59,7 +59,14 @@ const getFormattedClaimItemPremium = (claim: Claim): string => {
           <RowItem>{recentChange.Claim.status_change}</RowItem>
           <RowItem>{getClaimItemPersonName(recentChange.Claim)}</RowItem>
           <RowItem numeric>{getFormattedClaimItemValue(recentChange.Claim)}</RowItem>
-          <RowItem numeric>{getFormattedClaimItemPremium(recentChange.Claim)}</RowItem>
+          <RowItem numeric>
+            {getFormattedClaimItemPremium(recentChange.Claim)}
+            {#if isMonthly(recentChange.Claim.claim_items[0]?.item)}
+              <div>
+                <small class="tw-opacity-60">({getFormattedClaimItemPremium(recentChange.Claim)}/month)</small>
```
forevermatt, line 66: It looks like this will show the annual premium here as well, rather than the monthly premium. You could just `/ 12` here, but it may be better to set up another function that uses the actual `monthly_premium` field, to ensure we show consistent values (by having the backend do all of that math, to avoid differences in rounding).

forevermatt (changes requested):

src/components/RecentActivityTable.svelte
hobbitronics: Good catch

forevermatt (approved):

PR #655, opened 2023-10-12
CVR-705 Add monthly premiums to policy page

hobbitronics:
[CVR-705](https://itse.youtrack.cloud/issue/CVR-705/Update-UI-to-handle-monthly-vs-yearly-premiums)
## Added
- Added InfoDialog
- Added Monthly Premium (sum of all) to policy details

image

### DIFF
--- a/src/components/InfoModal.svelte
+++ b/src/components/InfoModal.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { Dialog, IconButton } from '@silintl/ui-components'
+import { InfoDialog } from '.'
 
 export let content = ''
 export let title = ''
@@ -12,6 +13,6 @@ const buttons: Dialog.AlertButton[] = [{ label: 'Ok', action: 'cancel', class: '
 {#if hasInfoButton}
   <IconButton class="gray" icon="info" on:click={() => (open = true)} />
 {/if}
-<Dialog.Alert {open} {buttons} defaultAction="cancel" {title} titleIcon="info" on:closed={() => (open = false)}>
+<InfoDialog infoIsOpen={open} {title} on:closed={() => (open = false)}>
   {content}
-</Dialog.Alert>
+</InfoDialog>
--- a/src/pages/policies/[policyId].svelte
+++ b/src/pages/policies/[policyId].svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { Breadcrumb, CardsGrid, ClaimsTable, ItemsTable, Row, Strikes, CustomerReport } from 'components'
+import { Breadcrumb, CardsGrid, ClaimsTable, InfoDialog, ItemsTable, Row, Strikes, CustomerReport } from 'components'
 import { isLoadingById, loading } from 'components/progress'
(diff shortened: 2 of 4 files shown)


src/pages/policies/[policyId].svelte
```diff
@@ -185,8 +200,15 @@ section:not(:first-child) {
       <dl>
         <dt>Coverage</dt>
         <dd>{coverage}</dd>
-        <dt>Premium</dt>
+        <dt>Yearly Premium</dt>
         <dd>{premium} per year</dd>
+        <dt class="tw-flex tw-items-center">
```
Comment (briskt) [inline src/pages/policies/[policyId].svelte:205]: Would it be difficult to hide the monthly premium if a policy has no vehicles? If not, it might be better to omit this since the large majority of policies do not have vehicles.

briskt (commented):
Question and suggestion about monthly premium on policy page.

src/components/InfoModal.svelte
```diff
@@ -12,6 +13,6 @@ const buttons: Dialog.AlertButton[] = [{ label: 'Ok', action: 'cancel', class: '
 {#if hasInfoButton}
   <IconButton class="gray" icon="info" on:click={() => (open = true)} />
 {/if}
-<Dialog.Alert {open} {buttons} defaultAction="cancel" {title} titleIcon="info" on:closed={() => (open = false)}>
+<InfoDialog infoIsOpen={open} {title} on:closed={() => (open = false)}>
```
forevermatt, line 16: It looks like both this InfoModal code and the inner InfoDialog code will be setting the InfoDialog's `infoIsOpen` to false when the dialog is closed. Would it be better to only have one of them do it? (Just a question; no request for change here unless you think it's better changed.)

src/pages/policies/[policyId].svelte
```diff
@@ -63,6 +73,7 @@ $: policyName = getNameOfPolicy(policy)
 $: policyName && (metatags.title = formatPageTitle(`Policies > ${policyName}`))
 $: coverage = formatMoney(approvedItems.reduce((sum, item) => sum + item.coverage_amount, 0))
 $: premium = formatMoney(approvedItems.reduce((sum, item) => sum + item.annual_premium, 0))
+$: monthlyPremiumTotal = formatMoney(approvedItems.reduce((sum, item) => sum + item.monthly_premium, 0))
```
Comment (forevermatt) [inline src/pages/policies/[policyId].svelte:76]: Unfortunately, I think we'll need to filter the items in both of these lists first by the `billing_period`, because even monthly items have an `annual_premium` value (which equals 12 times their monthly premium), so the `premium` calculation here currently already includes all of the monthly premiums.

I think it's helpful that we have the annual_premium data even for monthly-billed items (e.g. we're using it in the ItemsTable), but that means we have a little "gotcha" to watch our for when doing sums.

src/pages/policies/[policyId].svelte
Comment (forevermatt) [reply src/pages/policies/[policyId].svelte:205]: I think it'll be pretty easy to only show this when there _are_ items that are billed monthly. 👍

forevermatt (changes requested):

forevermatt (approved):

PR #657, opened 2023-10-16
change button spacing and typography

hobbitronics:
[CVR-674](https://itse.youtrack.cloud/issue/CVR-674/Improve-Button-typography)

## Changed
- typography
- Button letter spacing

These were reverted accidentally

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -67,7 +67,7 @@ export default {
         require('@tailwindcss/nesting'),
         require('tailwindcss'),
         require('autoprefixer'),
-      ]
+      ],
     }),
     routify({
       dynamicImports: false,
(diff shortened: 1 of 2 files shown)


rollup.config.js
```diff
@@ -109,7 +109,7 @@ function getHtml(script, css) {
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link href="https://fonts.googleapis.com/css?family=Material+Icons&display=block" rel="stylesheet">
```
mtompset, line 110: What about /'ing this line too (and others in this block)? The closing slashes are only required for XML. HTML allows with or without, as far as I know. It just seems weird to have a mixture of them.

briskt (approved):

forevermatt (approved):

rollup.config.js
hobbitronics: Normally I let prettier do the formatting, but since this is a big string it can't. I think I will update these in a separate PR.

PR #660, opened 2023-10-17
CVR-710: Fix logic bug re: needs receipt vs. repair/replacement

forevermatt:
[CVR-710](https://itse.youtrack.cloud/issue/CVR-710) Approval button says "Okay To Replace" when it should say "Okay To Repair"

---

### Fixed
- Fix logic bug re: needs receipt vs. repair/replacement

### DIFF
--- a/src/pages/policies/[policyId]/claims/[claimId].svelte
+++ b/src/pages/policies/[policyId]/claims/[claimId].svelte
@@ -108,6 +108,7 @@ $: needsReceipt = claimStatus === ClaimStatus.Receipt
 $: needsFile = needsReceipt || isEvidenceNeeded(claimItem, claimStatus)
 $: allowDelete = needsRevision || [ClaimStatus.Receipt, ClaimStatus.Draft].includes(claimStatus)
 
+$: isRepair = payoutOption === PayoutOption.Repair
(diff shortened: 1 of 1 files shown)


jason-jackson (approved):

src/pages/policies/[policyId]/claims/[claimId].svelte
```diff
@@ -108,6 +108,7 @@ $: needsReceipt = claimStatus === ClaimStatus.Receipt
 $: needsFile = needsReceipt || isEvidenceNeeded(claimItem, claimStatus)
 $: allowDelete = needsRevision || [ClaimStatus.Receipt, ClaimStatus.Draft].includes(claimStatus)
 
+$: isRepair = payoutOption === PayoutOption.Repair
 $: needsRepairReceipt = needsReceipt && payoutOption === PayoutOption.Repair
```
Comment (hobbitronics) [inline src/pages/policies/[policyId]/claims/[claimId].svelte:112]: you could use isRepair in line 112 as well

Comment (forevermatt) [reply src/pages/policies/[policyId]/claims/[claimId].svelte:112]: True. I thought about that, but it seemed better to keep lines 112 and 113 similar (so it's easy to see the difference between them), and I didn't want to also add an `isReplacement` just for line 113.

PR #662, opened 2023-10-17
Fix the condition for when to show the Delete button on an Item Form

forevermatt:
### Fixed
- Fix the condition for when to show the Delete button on an Item Form

---

I once again tested a function instead of calling it. I think this was leftover from a rename I did, where I apparently missed this occurrence.


jason-jackson (approved):
Good catch changing method to actual boolean

mtompset (approved):
Nice catch.

PR #663, opened 2023-10-18
CVR-752: Add monthly renewals button

forevermatt:
### Added
- Add "Monthly Renewal" section/buttons to `/admin/renewals` page


briskt (approved):
looks good, just a couple of ideas offered

briskt (approved):

PR #665, opened 2023-10-19
fix bug preventing goto after removing item in checkout CVR-748

hobbitronics:
[CVR-748](https://itse.youtrack.cloud/issue/CVR-748/After-discarding-item-during-checkout-go-back-to...-somewhere)

### DIFF
--- a/src/pages/policies/[policyId]/items/[itemId]/checkout.svelte
+++ b/src/pages/policies/[policyId]/items/[itemId]/checkout.svelte
@@ -22,7 +22,7 @@ const onEdit = () => {
 }
 
 const onDelete = async (event: CustomEvent<string>) => {
-  await deleteItem(policyId, event.detail)
+  deleteItem(policyId, event.detail) //don't await this or item will be undefined before the page navigates
   $goto(items(policyId))
 }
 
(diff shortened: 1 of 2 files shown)


src/pages/policies/[policyId]/items/[itemId]/checkout.svelte
```diff
@@ -22,7 +22,7 @@ const onEdit = () => {
 }
 
 const onDelete = async (event: CustomEvent<string>) => {
-  await deleteItem(policyId, event.detail)
+  deleteItem(policyId, event.detail) //don't await this or item will be undefined before the page navigates
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items/[itemId]/checkout.svelte:25]: Thanks for figuring out what the problem was here.

Would it be worth adding some kind of a `catch()` or something as well, so that if the deleteItem() call does fail, we can setNotice() or tell the user in some other way?

src/pages/policies/[policyId]/items/[itemId]/checkout.svelte
```diff
@@ -4,8 +4,9 @@ import { deleteItem, loadItems, PolicyItem, selectedPolicyItems, submitItem } fr
 import { formatPageTitle } from 'helpers/pageTitle'
 import { itemDetails, items, itemEdit } from 'helpers/routes'
 import { goto, metatags, params } from '@roxi/routify'
-import { Page } from '@silintl/ui-components'
+import { Page, setNotice } from '@silintl/ui-components'
 import { onMount } from 'svelte'
+import { set } from 'date-fns'
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/items/[itemId]/checkout.svelte:9]: It looks like this import was accidentally added
```suggestion
```

forevermatt (approved):
One more small thing to fix, but otherwise this looks good. Thanks! Pre-approving, since the final correction is an easy line deletion.

PR #666, opened 2023-10-19
fix deleting a dependent while reassigning items CVR-106

hobbitronics:
[CVR-106](https://itse.youtrack.cloud/issue/CVR-106/Error-when-trying-to-delete-dependent-who-is-an-accountable-person-on-an-Item-with-no-coverage.)

## Fixed
- dependent/members delete error due to not updating the item first

### DIFF
--- a/src/components/RemoveDependentModal.svelte
+++ b/src/components/RemoveDependentModal.svelte
@@ -31,18 +31,19 @@ const allTheButtons: Dialog.AlertButton[] = [
 
 $: buttons = selectedAccountablePersonOption || radioValue === 'no-one' ? allTheButtons : cancelButton
 
-const dispatch = createEventDispatcher<{ remove: string; cancel: string; assign: string; closed: string }>()
+const dispatch = createEventDispatcher<{ remove: string; assign: string }>()
 
 const handleDialog = (e: CustomEvent) => {
-  e.detail ? dispatch(e.detail) : dispatch('closed')
-  e.detail === 'remove' && selectedAccountablePersonOption && dispatch('assign', selectedAccountablePersonOption.id)
+  if (e.detail === 'remove') {
+    dispatch('remove', selectedAccountablePersonOption.id)
+  }
 }
 const onSelect = (e: CustomEvent) => {
   selectedAccountablePersonOption = e.detail
 }
 </script>
 
-<Dialog.Alert {open} {buttons} defaultAction="cancel" {title} on:closed={handleDialog} on:chosen={handleDialog}>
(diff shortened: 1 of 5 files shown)


src/components/RemoveDependentModal.svelte
```diff
@@ -31,18 +31,19 @@ const allTheButtons: Dialog.AlertButton[] = [
 
 $: buttons = selectedAccountablePersonOption || radioValue === 'no-one' ? allTheButtons : cancelButton
 
-const dispatch = createEventDispatcher<{ remove: string; cancel: string; assign: string; closed: string }>()
+const dispatch = createEventDispatcher<{ remove: string; assign: string }>()
```
forevermatt, line 34: It doesn't look like this `dispatch` is ever used in this file (as modified) to dispatch an "assign" event. Is that correct?

src/components/RemoveDependentModal.svelte
```diff
@@ -31,18 +31,19 @@ const allTheButtons: Dialog.AlertButton[] = [
...
 
 const handleDialog = (e: CustomEvent) => {
-  e.detail ? dispatch(e.detail) : dispatch('closed')
-  e.detail === 'remove' && selectedAccountablePersonOption && dispatch('assign', selectedAccountablePersonOption.id)
+  if (e.detail === 'remove') {
+    dispatch('remove', selectedAccountablePersonOption.id)
+  }
 }
 const onSelect = (e: CustomEvent) => {
   selectedAccountablePersonOption = e.detail
```
forevermatt, line 42: It doesn't look like this `selectedAccountablePersonOption` value is ever used any more, just set. How does this selected value get passed along to something that uses it (such as for a call to the API)? (Thanks)

src/data/items.ts
```diff
@@ -368,22 +368,31 @@ export const itemIsPending = (item: PolicyItem): boolean => {
...
+        make: item.make,
+        model: item.model,
+        name: item.name,
+        riskCategoryId: item.risk_category.id,
+        uniqueIdentifier: item.serial_number,
+      }),
+    )
+  }
+
+  await Promise.all(promises)
```
forevermatt, line 395: Nice 👍

forevermatt (approved):

briskt:
Can you do this as a hotfix? It would get it out the door quicker and make the coming mega-PR a bit smaller and better scoped to just vehicle-related changes.

hobbitronics:
> Can you do this as a hotfix? It would get it out the door quicker and make the coming mega-PR a bit smaller and better scoped to just vehicle-related changes.

Yes, we can do that.

forevermatt (approved):

PR #668, opened 2023-10-24
Use browsers' built-in currency formatting, to get thousands separators

forevermatt:
### Fixed
- Use browsers' built-in currency formatting, to get thousands separators


briskt (approved):
nice

PR #672, opened 2023-10-25
Fix monthly premium total

forevermatt:
[CVR-761](https://itse.youtrack.cloud/issue/CVR-761) Fix monthly premium total

---

### Fixed
- Fix monthly premium total on Policy details page

---

When running it locally, I was seeing a monthly premium of $13.33 that should have been $12.50. On further inspection, it looks like even annually-billed items (at least sometimes) have a monthly_premium value.

Breaking it out to get the lists of (approved) annual and monthly items first, then summing them, seems to have fixed it.

### DIFF
--- a/src/helpers/coverage.ts
+++ b/src/helpers/coverage.ts
@@ -65,3 +65,4 @@ export const isBeforeMonthlyCutoff = (): boolean => {
 }
 
 export const isMonthly = (item: PolicyItem) => item.billing_period === BillingPeriod.Monthly
+export const isYearly = (item: PolicyItem) => item.billing_period === BillingPeriod.Yearly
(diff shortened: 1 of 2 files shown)


briskt:
> When running it locally, I was seeing a monthly premium of $13.33 that should have been $12.50. On further inspection, it looks like even annually-billed items (at least sometimes) have a monthly_premium value.

Was this entirely a front-end issue? Your comment sounds like it might be incorrect coming from the API.

briskt (approved):

forevermatt:
> > When running it locally, I was seeing a monthly premium of $13.33 that should have been $12.50. On further inspection, it looks like even annually-billed items (at least sometimes) have a monthly_premium value.
> 
> Was this entirely a front-end issue? Your comment sounds like it might be incorrect coming from the API.

I don't know what we want the API to do in this regard. The UI code (at least on the Policy page) now handles this correctly.

I _think_ it would make sense for monthly-billed items to have both an annual_premium and monthly_premium, and for annually-billed items to only have an annual_premium (not a monthly_premium). The UI code previously assumed that annually-billed items wouldn't have a (non-zero) monthly_premium value, which led to this bug. As the code stands now, it could go either way. It probably would be good to return a `monthly_premium` of `0` for annually-billed items, but I don't know how much time/effort it's worth to do that.

PR #674, opened 2023-11-01
Uninstall sass

hobbitronics:
## Remove
- remove sass. I realized sass code is being parsed by plugins and nothing uses sass.


mtompset (approved):
This makes sense if it isn't used.

forevermatt (approved):

PR #675, opened 2023-11-30
CVR-784  Move build warnings to a list

hobbitronics:
[CVR-784](https://itse.youtrack.cloud/issue/CVR-784/clean-up-build-warnings)

## Change
- move remaining warnings to a filter list with a Todo
- update svelte-preprocess to remove a warning

see https://github.com/rollup/rollup/issues/408 for more info on filtering warnings

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
         "rollup-plugin-svelte": "^7.1.6",
         "rollup-plugin-terser": "^7.0.2",
         "svelte": "^3.55.1",
-        "svelte-preprocess": "^4.10.7",
+        "svelte-preprocess": "^5",
         "tailwind-utopia": "github:domchristie/tailwind-utopia",
         "tailwindcss": "^3.3.3",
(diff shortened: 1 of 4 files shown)


mtompset (approved):
My review doesn't count for much given a lack of familiarity with the code, but I think it is okay.

briskt:
> My inclination is to avoid filtering out the warnings, letting them keep showing until we fix them, but that's just a hunch/preference.

If I was pressed, I'd say the same. But not knowing the implications of these warnings, I can't really say how important it is to address them.

forevermatt (approved):
@hobbitronics As long as we will still have a YouTrack issue open about coming back and fixing these warnings that will now be hidden, I'm fine if you proceed with merging this.

hobbitronics:
Down to one [CVR-788](https://itse.youtrack.cloud/issue/CVR-788/fix-A11y-noninteractive-element-cannot-have-nonnegative-tabIndex-value)

PR #678, opened 2024-04-03
update ui-components and fix ts errors

hobbitronics:


src/components/AppDrawer.svelte
```diff
@@ -30,7 +30,7 @@ onMount(() => {
 $: drawerWidth = `${drawerEl?.offsetWidth || 0}px`
 
 //@ts-expect-error
-$beforeUrlChange((event?: PopStateEvent, route?: ClientNodeApi, { url }) => {
+$beforeUrlChange((_?: PopStateEvent, __?: ClientNodeApi, { url }) => {
```
briskt, line 33: What is this `_?` magic?

hobbitronics: Just TypeScript Optional Parameters, but types are no longer required with `_` so I can remove them

briskt (approved):

PR #686, opened 2024-04-09
prevent claim against non-active item CVR-808

hobbitronics:
[CVR-808](https://itse.youtrack.cloud/issue/CVR-808)

### DIFF
--- a/src/pages/policies/[policyId]/claims/new.svelte
+++ b/src/pages/policies/[policyId]/claims/new.svelte
@@ -9,8 +9,9 @@ import {
   loadClaimsByPolicyId,
   submitClaim,
 } from 'data/claims'
-import { loadItems, PolicyItem, selectedPolicyItems } from 'data/items'
+import { itemIsApproved, loadItems, PolicyItem, selectedPolicyItems } from 'data/items'
 import { selectedPolicyId } from 'data/role-policy-selection'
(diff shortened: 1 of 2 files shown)


src/pages/policies/[policyId]/claims/new.svelte
```diff
@@ -22,23 +23,23 @@ export let policyId = $selectedPolicyId
 
 let itemId: string
 
+$: items = $selectedPolicyItems.filter(itemCanClaimFilter) || []
```
Comment (forevermatt) [inline src/pages/policies/[policyId]/claims/new.svelte:26]: Based on the old code, it looks like `$selectedPolicyItems` isn't always defined (e.g. on a direct page load to this view, before the API calls have all resolved?). You may want to protect against that possibility, too:
```suggestion
$: items = $selectedPolicyItems?.filter(itemCanClaimFilter) || []
```

forevermatt (approved):
Approved, with one suggestion

src/pages/policies/[policyId]/claims/new.svelte
Comment (hobbitronics) [reply src/pages/policies/[policyId]/claims/new.svelte:26]: It seems to handle this ok because of the await. I suppose it's possible there could be a race condition from reactive statement, so I'll check `selectedPolicyItems` directly. I don't think the ? is needed as it defaults to an empty array.

PR #687, opened 2024-04-12
Fix ItemForm CVR-93

hobbitronics:
[CVR-93](https://itse.youtrack.cloud/issue/CVR-93)

image

image

image

TODO: update ui-components to released version when available

### DIFF
--- a/src/components/YearInput.svelte
+++ b/src/components/YearInput.svelte
@@ -51,13 +51,18 @@ afterUpdate(() => (width = `${element?.offsetWidth}px`))
 const focus = (node: any) => autofocus && node.focus()
 </script>
 
+<style>
+.error {
+  color: var(--mdc-theme-error);
+}
+</style>
+
 <label
   class="mdc-text-field mdc-text-field--outlined mdc-text-field--with-leading-icon {$$props.class ||
     ''} textfield-radius"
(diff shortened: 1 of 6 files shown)


src/components/YearInput.svelte
```diff
@@ -26,6 +28,7 @@ $: isValidYear = isFourDigitYear(value)
 $: showValidationMessages = !!(hasFocused && hasBlurred && !isFocused && value)
 $: error = showValidationMessages && !isValidYear
 $: mdcTextField && (mdcTextField.valid = !error)
+$: mdcTextField && (mdcTextField.valid = !showError)
```
briskt, line 31: can `mdcTextField.valid` be assigned to two different values?

briskt (approved):
looks good!

src/components/YearInput.svelte
hobbitronics: no, but `showError`'s value will only be set manually (a reactive statement might cause a back and forth scenerio) so it should be overriden if there is a change in the error status. I implemented this differently in `TextField` so if this is an issue we can do it like: `class:mdc-invalid-class={showError}` but I didn't notice a problem yet. I'll check it again before deployment.

PR #692, opened 2024-04-22
Release 1.12.2

hobbitronics:
### Fixed
- https://github.com/silinternational/cover-ui/pull/686
- https://github.com/silinternational/cover-ui/pull/687
- https://github.com/silinternational/cover-ui/pull/688
- https://github.com/silinternational/cover-ui/pull/689
- https://github.com/silinternational/cover-ui/pull/690
- https://github.com/silinternational/cover-ui/pull/691
- broken item search
### Feature branch checklist

- [x] Documentation (README, etc.)


src/components/forms/DependentForm.svelte
```diff
@@ -164,46 +169,80 @@ const onChosen = (event: CustomEvent) => (formData.country = event.detail)
     {/if}
     <p>
       <span class="header">Primary Location<span class="required-input">*</span></span>
-      <CountrySelector country={formData.country} on:chosen={onChosen} />
+      <CountrySelector
+        required
+        showError={showCountryError}
+        country={formData.country}
+        on:chosen={onChosen}
+        on:focus={() => (showCountryError = false)}
```
jason-jackson, line 177: It seems strange to use `focus` here, is that correct? Other ones are using `blur` or `change`. Mainly `blur`.

jason-jackson (commented):
I had some questions, which I suspect are fine.

src/components/forms/DependentForm.svelte
```diff
@@ -134,27 +140,26 @@ const onSubmit = () => {
...
 }
 
-.float-right {
-  float: right;
-}
-
-.form-button {
-  margin: 0.5rem;
+const onKeydown = (event: KeyboardEvent) => {
+  if (event.key === 'Enter') {
```
forevermatt, line 152: I'm curious: why the decision to go with an on onkeydown event on the button, rather than on onsubmit event on the form? It seems like this misses various ways the form could be submitted.

src/components/forms/ItemForm.svelte
```diff
@@ -52,6 +58,16 @@ let name = ''
...
 let year: number | undefined
 
+let hasModelWarn = false
+let hasMakeWarn = false
+let hasSerialWarn = false
+let hasNameError = false
+let hasAccountableError = false
+let hasCategoryError = false
+let hasValueError = false
+let hasYearError = false
```
forevermatt, line 68: Is the use of `Warn` vs `Error` meaningful in these variable names?

src/components/forms/DependentForm.svelte
hobbitronics: This is happening before the `submit` event since the validation from using `required` prevents it.

src/components/forms/ItemForm.svelte
hobbitronics: yes, Warn is orange, Error is red (applied to the field).

src/components/forms/DependentForm.svelte
hobbitronics: the was the only event exposed, so while I could make a change in the library, this was "good enough"

src/components/forms/DependentForm.svelte
hobbitronics: A little more info: The submit will still happen when all required fields are filled, but this is a way to color all the not filled out fields when submit is prevented by html validation.

briskt (approved):

jason-jackson (approved):

PR #696, opened 2024-05-14
use cookies instead of session storage CVR-610

hobbitronics:
[CVR-610](https://itse.youtrack.cloud/issue/CVR-610)

### Added

- Cookies library 🍪

### Changed

- token is now stored in a cookie 🍪

### Feature branch checklist


briskt (approved):

forevermatt (commented):
Moving this over from sessionStorage to a cookie means it will automatically be included in all web requests to the host. Is that desired? Are we then including it twice?

Just a couple of questions to sort through before proceeding all the way to production with this.

hobbitronics:
> Moving this over from sessionStorage to a cookie means it will automatically be included in all web requests to the host. Is that desired? Are we then including it twice?
> 
> Just a couple of questions to sort through before proceeding all the way to production with this.

Good call @forevermatt. I See where we can make those changes on the UI, but I think this will have to happen with some changes to the api, to be able to use the cookie in the request headers.

briskt (changes requested):
After further thought, I'd like to know if this can be done with most of the work happening on the backend.

hobbitronics:
I agree with @briskt on this being done on the backend as httpOnly cookies won't be accessible via Javascript.

PR #699, opened 2024-05-28
add return to param CVR-774

hobbitronics:
### Added

- redirect to return-to param

[CVR-774](https://itse.youtrack.cloud/issue/CVR-774)

--
### Feature branch checklist

### DIFF
--- a/src/authn/index.ts
+++ b/src/authn/index.ts
@@ -16,8 +16,8 @@ export const login = async (inviteCode = ''): Promise<void> => {
   if (inviteCode) {
     url += `?invite=${inviteCode}`
   }
-  if (returnUrl !== ROOT && returnUrl !== HOME) {
-    url += returnUrl.includes('?') || inviteCode ? `&return-to=${returnUrl}` : `?return-to=${returnUrl}`
+  if (returnUrl !== ROOT && returnUrl !== HOME && !inviteCode) {
+    url += returnUrl.includes('?') ? `&return-to=${returnUrl}` : `?return-to=${returnUrl}`
   }
   const responseData = await POST<AuthLoginResponse>(url)
   if (responseData.RedirectURL) {
--- a/src/authn/index.ts
+++ b/src/authn/index.ts
@@ -1,6 +1,7 @@
 import { CREATE as POST } from 'data'
 import { clearApp } from 'data/storage'
 import { throwError } from '../error'
+import { HOME, ROOT } from 'helpers/routes'
 import { writable } from 'svelte/store'
 
 export type AuthLoginResponse = {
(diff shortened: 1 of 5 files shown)


src/authn/index.ts
```diff
@@ -10,10 +11,14 @@ export type AuthLoginResponse = {
 export const showApp = writable(false)
 
 export const login = async (inviteCode = ''): Promise<void> => {
+  const returnUrl = location.pathname
   let url = 'auth/login'
   if (inviteCode) {
     url += `&invite=${inviteCode}`
```
briskt, line 17: Does this work? Seems like it should be `?invite=`

src/authn/index.ts
```diff
@@ -10,10 +11,14 @@ export type AuthLoginResponse = {
...
   }
+  if (returnUrl !== ROOT && returnUrl !== HOME) {
+    url += `?return-to=${returnUrl}`
```
briskt, line 20: This may need to be conditional, based on whether or not there's already a `?` in the URL.

briskt (changes requested):

src/authn/index.ts
hobbitronics: I haven't tested it, but wondered about that.

hobbitronics: I'm not sure how to get an invitation code to test

briskt (approved):

src/authn/index.ts
```diff
@@ -10,9 +11,13 @@ export type AuthLoginResponse = {
 export const showApp = writable(false)
 
 export const login = async (inviteCode = ''): Promise<void> => {
+  const returnUrl = location.pathname
   let url = 'auth/login'
   if (inviteCode) {
-    url += `&invite=${inviteCode}`
+    url += `?invite=${inviteCode}`
+  }
+  if (returnUrl !== ROOT && returnUrl !== HOME) {
```
hobbitronics, line 19:
```suggestion
  if (returnUrl !== ROOT && returnUrl !== HOME && !inviteCode) {
```

src/authn/index.ts
```diff
@@ -10,9 +11,13 @@ export type AuthLoginResponse = {
...
+    url += returnUrl.includes('?') || inviteCode ? `&return-to=${returnUrl}` : `?return-to=${returnUrl}`
```
hobbitronics, line 20:
```suggestion
    url += returnUrl.includes('?') ? `&return-to=${returnUrl}` : `?return-to=${returnUrl}`

```

hobbitronics (commented):
Maybe we should just not include the return-to when logging in from an invite code url.

src/authn/index.ts
briskt: Yeah, probably not.

PR #708, opened 2024-06-13
Release - use item flags for delete, disable button for Denied

hobbitronics:
### Fixed

- use item flags for assertions when deleting
- disable the batch item delete button for Denied items as the request will fail

### Feature branch checklist


briskt (approved):
approved, though you may want a second reviewer since I was the one that approved the first PR

PR #719, opened 2024-07-23
silence sentry on local dev

hobbitronics:
### Fixed

- silence sentry on local dev

[ITSE-1396](https://itse.youtrack.cloud/issue/ITSE-1396)
### Feature branch checklist


src/main.ts
```diff
@@ -6,26 +6,27 @@ import * as Sentry from '@sentry/svelte'
 const environment = process.env.CF_PAGES_BRANCH
 const dsn = process.env.SENTRY_DSN
 
-console.debug(`Sentry.init ${dsn} ${environment}`)
-Sentry.init({
-  dsn,
-  integrations: [Sentry.browserTracingIntegration(), Sentry.replayIntegration()],
-  environment,
+if (environment !== 'feature/branch-name' && environment !== 'local') {
```
mtompset, line 9: This condition is strange. What is the point of the first part of the condition?

hobbitronics: As mentioned in the other PR, its the default value when copying `example.env`, but this was the easiest way to identify local environments. I could check the hostname is `cover.local` instead, but we already had this env defined.

briskt (approved):

PR #720, opened 2024-07-23
list deps

hobbitronics:
### Added

- make list-deps

### Feature branch checklist


src/main.ts
```diff
@@ -6,26 +6,27 @@ import * as Sentry from '@sentry/svelte'
 const environment = process.env.CF_PAGES_BRANCH
 const dsn = process.env.SENTRY_DSN
 
-console.debug(`Sentry.init ${dsn} ${environment}`)
-Sentry.init({
-  dsn,
-  integrations: [Sentry.browserTracingIntegration(), Sentry.replayIntegration()],
-  environment,
+if (environment !== 'feature/branch-name' && environment !== 'local') {
```
mtompset, line 9: won't that match expressly to "feature/branch-name"? Isn't the intent if you are running a git feature branch version check?

hobbitronics: yes, that is the value in `example.env` and I assume nobody will name a feature branch that

briskt (approved):

PR #721, opened 2024-07-29
add Admin user role

briskt:
### Added
- Added Admin user role

https://itse.youtrack.cloud/issue/CVR-738


hobbitronics (approved):
UI code looks good, but when I went to home as an `Admin` there was a GET to `http://cover.local:3000/steward/recent` that failed with `ErrorNotAuthorized` and debug_msg :  "actor not allowed to perform that action on this resource"

hobbitronics:
> UI code looks good, but when I went to home as an `Admin` there was a GET to `http://cover.local:3000/steward/recent` that failed with `ErrorNotAuthorized` and debug_msg : "actor not allowed to perform that action on this resource"

Pulled api changes and works

